PR #12, opened 2018-05-12
Add basic ARIA mark-up

Menelion:
I've added basic ARIA mark-up for links and buttons.

The only thing I'm doubtful about is how to organize threads and replies. I experimented a bit, but this functionality needs to be tested and might be reverted in future.

I also didn't mark up user name links that toggle user IDs: I might implement it in future, but it needs also a description, otherwise a blind user won't understand what the click on a user name does.

### DIFF
--- a/web/app/components/auth-panel/auth-panel.jsx
+++ b/web/app/components/auth-panel/auth-panel.jsx
@@ -46,7 +46,7 @@ export default class AuthPanel extends Component {
               <strong className="auth-panel__username" onClick={this.toggleUserId}>{user.name}</strong>
               {isUserIdVisible && <span className="auth-panel__user-id"> ({user.id})</span>}.
               {' '}
-              <span className="auth-panel__pseudo-link" role="link" onClick={props.onSignOut}>Sign out?</span>
+              <span className="auth-panel__pseudo-link" role="link" tabIndex="0" onClick={props.onSignOut}>Sign out?</span>
             </div>
           )
         }
@@ -66,6 +66,7 @@ export default class AuthPanel extends Component {
                       <span
                         className="auth-panel__pseudo-link"
                         role="link"
+                        tabIndex="0"
                         onClick={() => props.onSignIn(provider)}
                       >{PROVIDER_NAMES[provider]}</span>
                     </span>
@@ -83,6 +84,7 @@ export default class AuthPanel extends Component {
               <span
                 className="auth-panel__pseudo-link auth-panel__admin-action"
                 role="link"
+                tabIndex="0"
                 onClick={this.toggleBlockedVisibility}
               >{isBlockedVisible ? 'Hide' : 'Show'} blocked</span>
             )
--- a/web/app/components/auth-panel/auth-panel.jsx
+++ b/web/app/components/auth-panel/auth-panel.jsx
@@ -46,7 +46,7 @@ export default class AuthPanel extends Component {
               <strong className="auth-panel__username" onClick={this.toggleUserId}>{user.name}</strong>
               {isUserIdVisible && <span className="auth-panel__user-id"> ({user.id})</span>}.
               {' '}
-              <span className="auth-panel__pseudo-link" onClick={props.onSignOut}>Sign out?</span>
+              <span className="auth-panel__pseudo-link" role="link" onClick={props.onSignOut}>Sign out?</span>
             </div>
           )
         }
@@ -65,6 +65,7 @@ export default class AuthPanel extends Component {
                       {comma}
                       <span
                         className="auth-panel__pseudo-link"
+                        role="link"
                         onClick={() => props.onSignIn(provider)}
                       >{PROVIDER_NAMES[provider]}</span>
                     </span>
@@ -81,6 +82,7 @@ export default class AuthPanel extends Component {
             user.admin && (
               <span
                 className="auth-panel__pseudo-link auth-panel__admin-action"
+                role="link"
                 onClick={this.toggleBlockedVisibility}
               >{isBlockedVisible ? 'Hide' : 'Show'} blocked</span>
             )
--- a/web/app/components/comment/comment.jsx
+++ b/web/app/components/comment/comment.jsx
@@ -244,7 +244,7 @@ export default class Comment extends Component {
 
     if (mods.view === 'preview') {
       return (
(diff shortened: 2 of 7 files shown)


web/app/components/comment/comment.jsx
```diff
@@ -306,6 +307,8 @@ export default class Comment extends Component {
             <span className={b('comment__score', {}, { view: o.score.view })}>
               <span
                 className={b('comment__vote', {}, { type: 'up', selected: scoreIncreased, disabled: isGuest || isCurrentUser })}
+                role="button"
+                aria-disabled={isGuest || isCurrentUser ? 'true' : 'false'}
```
igoradamenko: You can use just `isGuest || isCurrentUser` as a value here, can't you?

web/app/components/comment/comment.jsx
```diff
@@ -286,6 +286,7 @@ export default class Comment extends Component {
                 <a
                   className="comment__link-to-parent"
                   href={`${o.locator.url}#${COMMENT_NODE_CLASSNAME_PREFIX}${o.pid}`}
+                  aria-label="Parent comment"
```
igoradamenko: `Go to parent comment`? And maybe it's better to add `title` here too.

web/app/components/comment/comment.jsx
```diff
@@ -244,7 +244,7 @@ export default class Comment extends Component {
 
     if (mods.view === 'preview') {
       return (
-        <div className={b('comment', props, defaultMods)}>
+        <div className={b('comment', props, defaultMods)} role="listitem article" aria-level={mods.level}>
```
igoradamenko: Instead of `role="article"` I think you should change `div` with `article`.

But don't forget to add `display: block` to `.comment` in this case.

web/app/components/auth-panel/auth-panel.jsx
```diff
@@ -46,7 +46,7 @@ export default class AuthPanel extends Component {
               <strong className="auth-panel__username" onClick={this.toggleUserId}>{user.name}</strong>
               {isUserIdVisible && <span className="auth-panel__user-id"> ({user.id})</span>}.
               {' '}
-              <span className="auth-panel__pseudo-link" onClick={props.onSignOut}>Sign out?</span>
+              <span className="auth-panel__pseudo-link" role="link" onClick={props.onSignOut}>Sign out?</span>
```
igoradamenko: If you add `role="link"` you also should add `tabIndex="0"` because if you don't it's senselessly, isn't it?

web/app/components/comment/comment.jsx
igoradamenko: BTW, why do you think that `listitem` must be used here? Are you sure that it's OK to mix roles here?

web/app/components/root/root.jsx
```diff
@@ -223,7 +223,7 @@ export default class Root extends Component {
 
                 {
                   !!pinnedComments.length && (
-                    <div className="root__pinned-comments">
+                    <div className="root__pinned-comments" role="region" aria-label="Pinned Comments">
```
igoradamenko: `Pinned comments`, please (second word is lowercased).

web/app/components/root/root.jsx
```diff
@@ -268,13 +268,13 @@ export default class Root extends Component {
 
           {
             isBlockedVisible && (
-              <div className="root__main">
+              <div className="root__main" role="region" aria-label="Blocked Users">
```
igoradamenko: `Blocked users`.

And may be it's better to move all this things to BlockedUsers component root node.

web/app/components/comment/comment.jsx
igoradamenko: Oh, I've reread your first comment. As far as I understand, it's the experimental part what you were talking about.

What if we use `article` only for Comment component, `list` for `.root__threads` in Root? And for Thread we can always use `listitem` and `list` only when it has some replies inside. So in this case we will have structure like this one:

```
list (.root__threads)
— listitem list (Thread with Threads inside)
— — article (Comment)
— — listitem (Thread without Threads inside)
— — — article (Comment)
— — listitem (Thread without Threads inside)
— — — article (Comment)
— listitem (Thread without Threads inside)
— — article (Comment)
```

I'm not an expert in a11y, but it sounds OK for me. What do you think?

web/app/components/comment/comment.jsx
Menelion: I thought about it. However, in all specifications I always saw `aria-something="true"` or `"false"` (note the quotes). So would it be OK to leave it as you suggest? If so, I'll be glad to remove those unnecessary boolean values :)

web/app/components/comment/comment.jsx
Menelion: Agreed. Did that just for conciseness, our screen readers tend to say quite a lot of things at once :).

web/app/components/comment/comment.jsx
igoradamenko: Ah, it's pretty simple.

Here we have Threads inside Thread: https://github.com/umputun/remark/blob/master/web/app/components/thread/thread.jsx#L33-L41

So, you can see that nested Threads generated from `replies` array. You can use `replies.length` as a condition for adding `list` role. For example:

```jsx
const roles = ['listitem'].concat(replies.length ? 'list' : []);
return (
  <div className={b('thread', props)} role={roles}>
```

or even simpler:

```jsx
return (
  <div
    className={b('thread', props)}
    role={['listitem'].concat(replies.length ? 'list' : [])}
  >
```

But I'm not sure should we add `list` if child Threads are collapsed? If we shouldn't then conditional must be like this: `(!collapsed && replies.length) ? 'list' : []`. I don't know what the usual a11y pattern for situation like this is.

Menelion: Thanks! Usually if something is expanded or collapsed, we add `aria-expanded` which is either `true` or `false`. BTW, where are the Expand/Collapse (pseudo-)links located?

igoradamenko:
LGTM. @Menelion, have you finished with it? May I merge it?

Menelion:
@igoradamenko, Yes, please. However, I still don't get why the CI is failing :).

igoradamenko:
That's because we don't have CI for pull requests now. Somewhen we will — #14.

But now — thank you for PR, it's awesome 🐨

PR #33, opened 2018-05-15
Improved first load time and UX on first load

paralainer:
Two changes:
* doing requests in parallel on first load
* showing bouncer before js bundle is loaded

First load on slow internet is pretty slow, and also blank screen while it loading js bundle doesn't help, so decided to improve that.

Also have an idea to load comments before we actually loaded iframe and pass a message inside iframe on load, but this is more complicated change, its better to split them, so probably next PR (if tests will show that it makes sense)

### DIFF
--- a/web/app/components/root/root.jsx
+++ b/web/app/components/root/root.jsx
@@ -51,28 +51,25 @@ export default class Root extends Component {
 
   componentDidMount() {
     const { sort } = this.state;
-
-    api.getUser()
-      .then(data => store.set('user', data))
-      .catch(() => store.set('user', {}))
-      .finally(() => {
-        api.find({ sort, url })
-          .then(({ comments = [] } = {}) => store.set('comments', comments))
-          .catch(() => store.set('comments', []))
-          .finally(() => {
-            this.setState({
-              isLoaded: true,
-              user: store.get('user'),
-            });
-
-            setTimeout(this.checkUrlHash);
-          });
-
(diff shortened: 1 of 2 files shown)


web/app/components/root/root.jsx
```diff
@@ -51,28 +51,25 @@ export default class Root extends Component {
...
-          });
-
-        api.getConfig().then(config => {
+    api.getConfig().then(config => {
           store.set('config', config);
           this.setState({ config });
-        })
-      });
+    });
+    Promise.all([
```
paralainer: It seems that we can do that these requests in parallel, there is no need to wait result of first to make a second

web/iframe.html
```diff
@@ -13,10 +13,42 @@
...
+        0% {transform: scale(0);}
+        40% {transform: scale(1);}
+        80%, 100% {transform: scale(0);}
+    }
   </style>
 </head>
 <body>
-  <div id="remark42"></div>
+  <div id="remark42">
+      <div class="iframe_preloader">
```
paralainer: Basically I am duplicating code here and in css. As code is pretty simple IMO its not a big deal, but if you have an idea of how to deduplicate it I am happy to implement.

igoradamenko: It's fine. I will try to optimise in the nearest future.

igoradamenko:
LGTM.

PR #40, opened 2018-05-19
norm admins list

kazhuravlev:
without this change - in config response we will see:

```json
{
...
"Admins": null,
...
}
```


umputun:
IMHO the better place to address this REST inconsistency will be `configCtrl`

Issue #14, opened 2018-05-12
Migrate from private drone to public travis


umputun:
I'd like to make build logs available for PR authors but don't want to keep it in my private drone. Travis seems to be the popular choice for such kind of things.

comjf:
I've used travis-ci for a number of years (though on closed source projects). I'd be happy to help look into this task if you don't already have an owner.

Any funky notes about the way your testing? I plan to dig further into the source this week.

umputun:
it will be very helpful, thx.

Nothing tricky in this build. Essentially it just runs docker build and pushes the image to docker hub. All tests as well as test coverage are the part of Dockerfile

denji:
Build artifacts on `hub.docker.com` and integration public `travis-ci` will be very useful.

AlekSi:
May you please enable repository here: https://travis-ci.org/umputun/remark? This will simplify life for people working on this issue. You may also enable "Build only if .travis.yml is present" so you don't receive bogus notifications.

PR #41, opened 2018-05-19
Some DB errors is not errors

kazhuravlev:
get comment by id on non exists id or non exists bucket:
`HTTP 400 + err` => `HTTP 404`

get comments list on non exists page bucket, userID bucket:
`HTTP 400 + err` => `HTTP 200 + blank comment list "[]"`

### DIFF
--- a/app/migrator/remark_test.go
+++ b/app/migrator/remark_test.go
@@ -28,7 +28,7 @@ func TestRemark_Export(t *testing.T) {
 	c1, err := buf.ReadString('\n')
 	assert.Nil(t, err)
 	log.Print(c1)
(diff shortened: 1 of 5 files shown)


kazhuravlev:
one of tests is failed. but this test also failed on master branch (`TestRemark_Export`). 
I fix this err in this PR

kazhuravlev:
@umputun please show me error cause. 
on my side all tests, linter checks and docker-compose build is worked fine.

umputun:
until we switch to public CI (see #14) the only way to see what failed the build the same way as CI is to run docker build.

```
app/store/engine/bolt.go:217::warning: declaration of "err" shadows declaration at app/store/engine/bolt.go:210 (vetshadow)
28s
516
app/store/engine/bolt.go:488::warning: declaration of "err" shadows declaration at app/store/engine/bolt.go:477 (vetshadow)
28s
517
app/rest/api/rest.go:354:10:warning: if block ends with a return statement, so drop this else and outdent its block (golint)
28s
518

```

umputun:
I'm not sure what the purpose of this PR. What exactly the problem it's trying to solve by adding all this code? I mean, I get what it's trying to achieve is to differentiate store errors with more granularity, but don't get why all of this even needed in the first place.

Do you have any particular use-case in mind? To me, requests for non-exiting id, bucket, and the user feels like a good fit for "bad request".

kazhuravlev:
I do not speak English well. I hope you understand me:
I think that client must be simple as possible. Client can not process bad request errors. But client must work with comments API as with faceted search.

close until #14 ))

PR #45, opened 2018-05-19
Add basic Travis CI configuration.

AlekSi:
Refs #14.


umputun:
thx for the PR. two questions:

1. why `sudo: required` ?
1. `dist: trusty` - probably doesn't matter as we use docker build, but still

PR #46, opened 2018-05-20
Remove extra lines from Travis CI config.

AlekSi:
See #45 for discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,3 @@
-dist: trusty
-sudo: required
-language: go
-
-services:
-  - docker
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,3 @@
-dist: trusty
-sudo: required
-language: go
-
-services:
```
umputun: is it a good idea? does it mean we expect docker to be in default build os/vm/container?

AlekSi: See my last comment at #45.

PR #58, opened 2018-05-29
Remove comments on production build in a web part

mxck:
Remove some comments from used libraries and own code in minimized production build.


igoradamenko:
What kind of comments are you talking about? By default Uglify must remove comments in code, except license comments.

Please, add some examples of removed comments.

mxck:
Oh, ok. My bad.
If you need to keep them, then close the pr :)

Issue #34, opened 2018-05-15
please add vk & yandex oath


bzdel:
если возможно, добавьте, пожалуйста

umputun:
Sorry, but I will not support VK oauth. For Yandex - tried to do it but run into multiple issues mostly due to lacking and incomplete documentation. I have nothing against Yandex oath, but this won't be a priority. If someone wants to add Yandex to the list of providers (see [auth/providers.go](https://github.com/umputun/remark/blob/master/app/rest/auth/providers.go)) I'll integrate it.

anatolym:
@bzdel, yandex oauth was added. I think we can close this issue.

sergio-fry:
@umputun Could you explain why don't you want to support vk ? Looks like go oauth2 supports vk provider as well.

sergio-fry:
Huh! Looks like it has been implemented after this issue was closed: golang.org/x/oauth2/vk

niocncn:
Add vk.com support please

umputun:
why? We have many popular and more "generic" providers already in place. What is the use case requiring to have vk in addition?

niocncn:
Because Vk is very popular social network in east europe. Your question is the same as "why do we need twitter provider if we have facebook".

PR #64, opened 2018-06-03
add Yandex OAuth provider #34

anatolym:
I added the Yandex OAuth provider, altered the UI and documentation, added unit-tests and check the Yandex login flow manually.
  The list of permissions has *Access to user avatar* and *Access to username, first name and surname, gender*. The latter is needed to get a friendly username. In fact a yandex user can have 3 different names:
  * `login` – same as first part of email address, this one is always required,
  * `real_name` – the combination of First and Last names, optional,
  * `display_name` – the name is used for displaying user name in yandex services, this name is visible to other yandex users.
  So more preferable name for us is `display_name`. If it's empty we use `real name`, otherwise `login`.
  The doc is here: https://tech.yandex.com/passport/doc/dg/reference/response-docpage/#norights_1

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -180,7 +180,8 @@
     "google",
     "internal",
     "jws",
-    "jwt"
+    "jwt",
+    "yandex"
   ]
   revision = "543e37812f10c46c622c9575afd7ad22f22a12ba"
 
@@ -222,6 +223,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "dd700af9d574d75a9afc92a7dbf61485a1785c38c7b93667b1be423c4fe814d7"
+  inputs-digest = "e849348430ea10a86642c328c52b516e237207389cf765ff1849c8bded391d4a"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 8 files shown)


umputun:
Thx for the PR. I'm not sure yet if we really need Yandex as another provider, but I'll think about it. Github related change will be nice to have as a separate PR and I could merge it right away.

anatolym:
@umputun, got it. I will create another PR with github refactoring.

umputun:
Regarding Yandex scope - is it possible to minimize it? We need just 3 things:

- some ID
- display name
- avatar

Things like first name and surname, gender, real name I'm trying to avoid, but not sure if it even possible with Yandex.

umputun:
@igoradamenko can you take a look at js related part of this PR?

umputun:
@anatolym could you resubmit with the recent master? We have a conflict on Gopkg.lock

anatolym:
> Regarding Yandex scope - is it possible to minimize it? We need just 3 things:

@umputun, unfortunately nope. If we don't request the scope "Access to username, first name and surname, gender", we can use user `login` only. Display name comes together with gender and other stuff of the scope.
I'm not sure that's correct behavior since yandex users can hide `login` name on public services and use his Display name.

> could you resubmit with the recent master? We have a conflict on Gopkg.lock

Sure, gimme a minute.

umputun:
thx, merged. Btw, if you could take care on twitter auth this will be really great.

anatolym:
@umputun sure, I will take a look at twitter api.

PR #65, opened 2018-06-03
refactor Github OAuth provider

anatolym:
I checked out the github API response for user profile. If user left his profile name empty, `"user": null` returns in the response. When go's `json.Unmarshal` decodes json `null` value it converts it to `nil`. So I refactored method `value()` of `userData` struct in `remark/app/rest/auth/provider.go` to avoid checking string `"<nil>"` in github provider.

(from PR #64)


umputun:
thx, LGTM

Issue #69, opened 2018-06-04
ESC is way too destructive


umputun:
I just run into this issue - tried to write a comment and hit ESC by accident. This honest mistake cleared all the text I've entered and collapsed the comment. I don't think we should even react on ESC, and if we do - reaction shouldn't be soo destructive.

igoradamenko:
Closed by #71.

Issue #62, opened 2018-05-31
annoying "back" button behaviour


cdracm:
1. go to https://radio-t.com/p/2018/05/26/podcast-599/#remark42__comment-1160825e-ee76-46b4-8a3f-5821d621a86a
2. marvel at the umputun wisdom
3. press "back" browser button. Nothing happens. Press back again and curse. Now you are back. Finally!

(there must be some redirections going on behind the scenes which prevent "back" button from going back)

umputun:
not sure if this is really what annoying, but unusual for sure. @igoradamenko probably due to anchor navigation?

igoradamenko:
Closed by #71.

PR #71, opened 2018-06-05
Prevent closing comment input field on ESC

life777:


life777:
Fixed [#69](https://github.com/umputun/remark/issues/69) and [#62](https://github.com/umputun/remark/issues/62) issues. @umputun

web/iframe.html
```diff
@@ -83,9 +83,10 @@
     function receiveMessages(event) {
       try {
         const data = typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
+        const isItHash = data.hash && data.hash.indexOf("#") === 0;
```
igoradamenko: Why do you need this check?

life777: postMessage can be called by everyone on the parent window. If someone decides to send some data with post message and instead of hash it puts some random string, it will cause redirection to the random URL with the string (because of the replace method). But If this string starts with "#", only hash part will be change, url will stay the same and we protect our self from dangerous redirections.  @igoradamenko

igoradamenko: I'm not sure that it's a big deal, but thanks for the detailed explanation. I didn't think about it :-)

Issue #78, opened 2018-06-11
Add avatar resizing


anatolym:
@umputun:

> There is a thing in the same area of auth - avatar proxy. Right now it's keeping the original image as-is and then serves it to users from AvatarStrore. It works, but the solution is not optimized. You can try to make it better by resizing images to the desired size and keep those tiny pics only instead of full originals. I think this resizing can happen synchronously at the time of login. In case this procedure is heavy for large images (I hope it is not that heavy) you may need some workes to do it in separate goroutines.

anatolym:
Implemented in #83.

Supported incoming avatar formats: PNG, JPEG, GIF.
Resized format: PNG.
If resizing not available (unsupported format) or not required (resize limit = 0px or image is smaller than the limit) – original avatar is kept.

Quick research shows avatar format usage by provider:

Provider | Format
----|----
Github | PNG
Google | JPEG
Facebook | JPEG
Yandex | JPEG

PR #83, opened 2018-06-16
add avatar resizing #78

anatolym:


anatolym:
@umputun, could you please share details of drone CI errors?https://drone.umputun.com/umputun/remark/884 (I don't have access :/ )
Thanks.

umputun:
this was unrelated to your PR, just not updated image after [this change](https://github.com/umputun/baseimage/commit/1ea43872688bd3a4032c208529dd6abc8591236a) in the build image.  I have pulled it and restarted the build, all good.

Generally, you can ignore drone's error as it runs the same thing as travis with extra steps for image publishing and deployment.

anatolym:
@umputun, got it, thanks.
I actually experienced the same problem on my laptop, but `--pull` option of docker-compose helped.

app/rest/proxy/avatar_store.go
```diff
@@ -56,7 +65,22 @@ func (fs *FSAvatarStore) Put(userID string, reader io.Reader) (avatar string, er
 		}
 	}()
 
-	if _, err = io.Copy(fh, reader); err != nil {
+	img, err := ioutil.ReadAll(reader)
```
umputun: I think this can be done directly on reader, without this ReadAll putting potentially enormous []byte to memory

app/rest/proxy/avatar_store.go
```diff
@@ -85,3 +109,45 @@ func (fs *FSAvatarStore) location(id string) string {
 	partition := checksum64 % 100
 	return path.Join(fs.storePath, fmt.Sprintf("%02d", partition))
 }
+
+// Resizes an image of supported format (PNG, JPG, GIF) to the size of "limit" px of the biggest side
+// (width or height) preserving aspect ratio.
+// Reruns nil if resizing is not needed.
+func resize(img []byte, limit int) ([]byte, error) {
```
umputun: this one may get Reader as input and return another Reader as result

app/rest/proxy/avatar_store.go
```diff
@@ -85,3 +109,45 @@ func (fs *FSAvatarStore) location(id string) string {
...
+	src, _, err := image.Decode(bytes.NewReader(img))
+	if err != nil {
+		return nil, errors.Wrap(err, "can't decode avatar image")
+	}
+
+	bounds := src.Bounds()
+	w, h := bounds.Dx(), bounds.Dy()
+	if w <= limit && h <= limit || w <= 0 || h <= 0 {
+		// If resizing image is smaller that the limit or has 0 size, return reader unchenged.
+		return nil, nil
```
umputun: any good reason why we need a special case here? I mean it can just return input as-is

app/rest/proxy/avatar_store.go
```diff
@@ -85,3 +109,45 @@ func (fs *FSAvatarStore) location(id string) string {
...
+	m := image.NewRGBA(image.Rect(0, 0, newW, newH))
+	// draw.ApproxBiLinear.Scale(m, m.Bounds(), src, src.Bounds(), draw.Src, nil)
+	draw.BiLinear.Scale(m, m.Bounds(), src, src.Bounds(), draw.Src, nil) // Slower but better quality.
+
+	var out bytes.Buffer
+	if err = png.Encode(&out, m); err != nil {
+		return nil, errors.Wrapf(err, "can't encode resized avatar buffer %s, id")
+	}
+
+	// write new image to file
```
umputun: not really to file, but to []byte

app/rest/proxy/avatar_store.go
```diff
@@ -85,3 +101,41 @@ func (fs *FSAvatarStore) location(id string) string {
 	partition := checksum64 % 100
 	return path.Join(fs.storePath, fmt.Sprintf("%02d", partition))
 }
+
+// Resizes an image of supported format (PNG, JPG, GIF) to the size of "limit" px of the biggest side
+// (width or height) preserving aspect ratio.
+// Returns original reader if resizing is not needed or failed.
+func resize(reader io.Reader, limit int) (io.Reader, error) {
```
umputun: returned error feels wrong to me as this is not really an error and caller forced to do an unusual thing - to use return value even if error != nil. I think it will be better to return just the reader and log directly inside resize fn.

app/rest/proxy/avatar_store.go
```diff
@@ -85,3 +101,41 @@ func (fs *FSAvatarStore) location(id string) string {
...
+	if err != nil {
+		return &teeBuf, errors.Wrap(err, "can't decode avatar image")
+	}
+
+	bounds := src.Bounds()
+	w, h := bounds.Dx(), bounds.Dy()
+	if w <= limit && h <= limit || w <= 0 || h <= 0 {
+		return &teeBuf, errors.New("resizing image is smaller that the limit or has 0 size")
+	}
+	var newW, newH int
```
umputun: I don't like else construct, and this part will be simpler without. To me, this little unnecessary calculation justified by better readability. 

```
newW, newH := w*limit/h, limit
if w > h {
	newW, newH = limit, h*limit/w
}
```

PR #84, opened 2018-06-17
Feature/store

umputun:
Adds structured options for two reasons:

1. Simplify params/env management
2. Preparation for additional implementation of stores and cache


app/main_test.go
```diff
@@ -49,25 +49,32 @@ func TestApplicationFailed(t *testing.T) {
 	p := flags.NewParser(&opts, flags.Default)
 
 	// RO bolt location
-	p.ParseArgs([]string{"--secret=123456", "--url=https://demo.remark42.com", "--bolt=/dev/null"})
+	p.ParseArgs([]string{"--secret=123456", "--url=https://demo.remark42.com", "--store.bolt.path=/dev/null"})
```
golangci: Error return value of `p.ParseArgs` is not checked

app/main_test.go
```diff
@@ -49,25 +49,32 @@ func TestApplicationFailed(t *testing.T) {
...
 	_, err := New(opts)
 	assert.EqualError(t, err, "can't initialize data store: failed to make boltdb for /dev/null/remark.db: "+
 		"open /dev/null/remark.db: not a directory")
 	t.Log(err)
 
 	// RO backup location
 	opts = Opts{}
-	p.ParseArgs([]string{"--secret=123456", "--url=https://demo.remark42.com", "--bolt=/tmp", "--backup=/dev/null/not-writable"})
+	p.ParseArgs([]string{"--secret=123456", "--url=https://demo.remark42.com", "--store.bolt.path=/tmp",
```
golangci: Error return value of `p.ParseArgs` is not checked

app/main_test.go
```diff
@@ -49,25 +49,32 @@ func TestApplicationFailed(t *testing.T) {
...
+		"--backup=/dev/null/not-writable"})
 	_, err = New(opts)
 	assert.EqualError(t, err, "can't check directory status for /dev/null/not-writable: stat /dev/null/not-writable: not a directory")
 	t.Log(err)
 
 	// invalid url
 	opts = Opts{}
-	p.ParseArgs([]string{"--secret=123456", "--url=demo.remark42.com", "--bolt=/tmp"})
+	p.ParseArgs([]string{"--secret=123456", "--url=demo.remark42.com", "----store.bolt.path=/tmp"})
```
golangci: Error return value of `p.ParseArgs` is not checked

Issue #96, opened 2018-06-23
Improve frontend components structure


igoradamenko:
Right now we have ten components, and most of them a really huge. There must be a way to simplify most of them by means of splitting into smaller components.

For example, comment component right now implements logic for regular comment, for pinned comment, for comment in last comments widget (!), etc. It contains lots of logic about choosing the right way to render. It definitely must be simplified.

PR #100, opened 2018-06-23
Setup testing environment

Mavrin:
I created setup for tests and write few tests on component and model.

### DIFF
--- a/web/app/common/store.test.js
+++ b/web/app/common/store.test.js
@@ -3,9 +3,11 @@ import store from './store';
 describe('Store', () => {
   it('should trigger listener on add comment', () => {
     const listener = jest.fn();
+
     store.set('comments', []);
     store.onUpdate('comments', listener);
     store.addComment({id:`new`});
+
     expect(listener).toBeCalledWith([{"comment": {"id": "new"}}]);
   });
 });
--- a/web/app/common/store.test.js
+++ b/web/app/common/store.test.js
@@ -1,0 +1,11 @@
+import store from './store';
+
+describe('Store', () => {
+  it('should trigger listener on add comment', () => {
+    const listener = jest.fn();
+    store.set('comments', []);
+    store.onUpdate('comments', listener);
+    store.addComment({id:`new`});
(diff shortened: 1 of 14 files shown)


web/app/common/store.test.js
```diff
@@ -0,0 +1,11 @@
+import store from './store';
+
+describe('Store', () => {
+  it('should trigger listener on add comment', () => {
+    const listener = jest.fn();
```
igoradamenko: Can you add some space between lines, please? Smth like this:

```
const listener = jest.fn();

store.set('comments', []);
store.onUpdate('comments', listener);
store.addComment({id:`new`});

expect(listener).toBeCalledWith([{"comment": {"id": "new"}}]);
```

And so on in the others files.

web/app/components/auth-panel/auth-panel.jsx
```diff
@@ -50,7 +51,7 @@ export default class AuthPanel extends Component {
         }
 
         {
-          !user.id && (
+          loggedIn === false && (
```
igoradamenko: `!loggedIn`?

Mavrin: Fixed. But  I don't like `expression && doSomething()` pattern. It is not easy read.

igoradamenko: I understand, but it's the only way to use conditions inside JSX ¯\_(ツ)_/¯.

Mavrin: I prefer to use  isLogged?<UserInfo>: <AuthForm>

web/app/common/store.test.js
Mavrin: Fixed

web/app/components/auth-panel/auth-panel.jsx
igoradamenko: Oh, I got it.

We have #96 for it. So if you want you can create these components and use a ternary operator here, in your next PR ^__^

PR #104, opened 2018-06-23
Refactor API definition (fixed #94)

### LINKED ISSUE #94: Refactor API definition on frontend part
We have api.js which defines all endpoints which is used on frontend: https://github.com/umputun/remark/blob/master/web/app/common/api.js.

But for some historical reasons most of them have unclear names. It would be better rename them using prefixes like `get`, `update`, etc.

I've added label `good-first-issue` here, because I think you can do it without any experience in frontend dev. Just try to fix it.

alexelev:


igoradamenko:
Awesome, thanks! 💪

PR #105, opened 2018-06-24
Polyfils

### LINKED ISSUE #102: Remove useless polyfills
When we started to develop frontend part, we added [babel-polyfill](https://www.npmjs.com/package/babel-polyfill), because it was easy. Now I'm sure that we don't need all stuff that it contains.

Remark42 should support IE11+, Android 4.4.4+ and iOS 9+, so we need to check what polyfills are actually in use, and maybe if we don't use things like `Array.from`, `Object.assign`, etc, we can remove babel-polyfill at all. Or at least we can require just some parts of the package.

babel-polyfill is the heaviest package in our set now, and we need to fix it:

screen shot 2018-06-23 at 21 28 41

Guria:
closes #102 

Polyfills (except promise from core-js) splitted into separate bundle which isn't loaded in modern browsers.
webpack-bundle-analyzer added to explore bundle sizes and its content.

Possible improvements: 
split shared dependencies like axios and preact into separate common bundle.


web/jsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "compilerOptions": {
+        "baseUrl": "./app",
```
Guria: This file is for vscode to support absolute paths.
Not sure we should it keep in repo.
Tell me what you think about it.

web/package.json
```diff
@@ -2,42 +2,44 @@
   "name": "remark-ui",
   "version": "0.1.0",
   "scripts": {
-    "build": "NODE_ENV=production webpack --config ./webpack.config.js",
+    "build": "cross-env NODE_ENV=production webpack --config ./webpack.config.js",
     "start": "webpack-dev-server --progress --hot --inline --config ./webpack.config.js",
     "reinstall": "rm -rf ./node_modules/ && npm install",
```
Guria: There is `npm ci` command in npm@6 that does the same and works quicker.

web/webpack.config.js
```diff
@@ -62,11 +65,17 @@ module.exports = {
           options: {
             presets: [
               ['env', {
-                targets: ['> 1%', 'android >= 4.4.4', 'ios >= 9'],
+                targets: {
```
Guria: Config was ignored because it was malformed

web/.eslintignore
```diff
@@ -0,0 +1 @@
+public
```
igoradamenko: Is it possible to add exclude by arguments, instead of a file? It's just a one line, so I think it's better to keep it in package.json :-)

web/jsconfig.json
igoradamenko: It looks like a regular editor's env file, which need to be added to gitignore.

Guria:
@igoradamenko I have made requested changes

Guria:
Wait a bit. I'll do so measurements

Guria:
A bit of stats with minified uncompressed sizes:

current:

Length | Name
------ | ----
 20310 | counter.js
  2673 | embed.js
 57366 | last-comments.js
144121 | remark.js
224470 | Total

with this PR

Length | Name
------ | ----
 20298 | counter.js
  2673 | embed.js
 57272 | last-comments.js
 70086 | polyfills.js
 91998 | remark.js
242327 | Total (old browsers)
172241 | Total except polyfills

We can explore if it's possible to shrink polyfills bundle, but not sure if it worth time.

igoradamenko:
Thanks! 💪

PR #109, opened 2018-06-25
ISSUE-73: UI support for DeleteMe request

### LINKED ISSUE #73: UI support for DeleteMe request
_see https://github.com/umputun/remark/issues/47#issuecomment-394103630_

UI should show a link (authed users only) to delete all info. The request is `POST /deleteme?site_id=site` and response `{"site": siteID, "user_id": user.ID, "token": tokenStr, "link": link}`

UI should invoke email client and fill "to" with admin email (part of `GET /config` response), subj with "Request to delete user's information" and add text like this:

```
Request to delete all information about <user-id> from remark42 on <site-id>

[you can provide the reason for removal request, optional]

=== DO NOT REMOVE TEXT BELOW THIS LINE ===

site: <site-id>
user: <user-id>
link: <link>
```

steppefox:
Closes #73

* Add the API post request deleteMe
* Add a trigger button
* Add an utils/email module with the template and window opener
* Add a dropdown component
* Add a button component
* Hide the user-id text and a logout bout in the dropdown
* Add a noop empty function to pass a default handler callback to components

## Issue
https://github.com/umputun/remark/issues/73

## Description
UI should show a link (authed users only) to delete all info. The request is POST /deleteme?site_id=site and response {"site": siteID, "user_id": user.ID, "token": tokenStr, "link": link}

UI should invoke email client and fill "to" with admin email (part of GET /config response), subj with "Request to delete user's information" and email text

## Screenshot and Gif
image

---

delete-me

### DIFF
--- a/web/app/components/button/button.jsx
+++ b/web/app/components/button/button.jsx
@@ -1,0 +1,69 @@
+/** @jsx h */
+import { Component, h } from 'preact';
+import noop from '../../utils/noop';
+
+export default class Button extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isClicked: false,
+      isFocused: false,
+    };
+
+    this.onMouseDown = this.onMouseDown.bind(this);
+    this.onFocus = this.onFocus.bind(this);
+    this.onBlur = this.onBlur.bind(this);
+  }
+
+  onMouseDown() {
+    this.setState({
+      isClicked: true,
+    });
+  }
+
+  onClick(e) {
+    this.props.onClick(e);
+  }
+
+  onBlur(e) {
+    this.setState({
+      isClicked: false,
+      isFocused: false,
+    });
+
+    this.props.onBlur(e);
+  }
+
+  onFocus(e) {
+    this.setState({
+      isFocused: true,
+    });
+
+    this.props.onFocus(e);
+  }
+
+  render(props, state) {
+    const { children, mix, mods, ...rest } = props;
+    const { isClicked, isFocused } = state;
+
+    return (
+      <button
+        {...rest}
+        className={b('button', { mix, mods }, { clicked: isClicked, focused: isFocused })}
+        onMouseDown={this.onMouseDown}
+        onBlur={this.onBlur}
+        onFocus={this.onFocus}
+      >
+        {children}
+      </button>
+    );
+  }
+}
+
+Button.defaultProps = {
+  type: 'button',
+  onClick: noop,
+  onBlur: noop,
+  onFocus: noop,
+};
--- a/web/app/components/dropdown/__item/dropdown__item.scss
+++ b/web/app/components/dropdown/__item/dropdown__item.scss
@@ -1,0 +1,14 @@
+.dropdown__item {
+  a, button {
+    display: block;
+    width: 100%;
+    text-align: left;
+    padding: 5px 15px;
+  }
+
+  &_separator {
+    border-bottom: 1px solid #259c9a;
+    margin-bottom: 5px;
+    padding: 5px 15px;
+  }
+}
--- a/web/app/components/dropdown/dropdown.jsx
+++ b/web/app/components/dropdown/dropdown.jsx
@@ -1,0 +1,76 @@
+/** @jsx h */
+import { Component, h } from 'preact';
+
+import Button from 'components/button';
+
+export default class Dropdown extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isActive: props.isActive || false,
+    };
+
+    this.onTitleClick = this.onTitleClick.bind(this);
+    this.onOutsideClick = this.onOutsideClick.bind(this);
+  }
+
+  onTitleClick() {
+    this.setState({
+      isActive: !this.state.isActive,
+    });
+
+    if (this.props.onTitleClick) {
+      this.props.onTitleClick();
+    }
+  }
+
+  onOutsideClick(e) {
+    if (!this.rootNode.contains(e.target)) {
+      if (this.state.isActive) {
+        this.setState({
+          isActive: false,
+        });
+      }
+    }
+  }
+
+  componentDidMount() {
+    document.addEventListener('click', this.onOutsideClick);
+
+    if (parent) {
+      parent.document.addEventListener('click', this.onOutsideClick);
+    }
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('click', this.onOutsideClick);
+
+    if (parent) {
+      parent.document.removeEventListener('click', this.onOutsideClick);
+    }
+  }
+
+  render(props, { isActive }) {
+    const { title, heading, children, mix, mods } = props;
+
+    return (
+      <div className={b('dropdown', { mix, mods }, { active: isActive })} ref={r => (this.rootNode = r)}>
+        <Button
+          aria-haspopup="listbox"
+          aria-expanded={isActive && 'true'}
+          mix="dropdown__title"
+          type="button"
+          onClick={this.onTitleClick}
+        >
+          {title}
+        </Button>
+
+        <div className="dropdown__content" tabindex="-1" role="listbox">
+          {heading && <div className="dropdown__heading">{heading}</div>}
+          <div className="dropdown__items">{children}</div>
+        </div>
+      </div>
+    );
+  }
+}
--- a/web/app/components/dropdown/dropdown.scss
+++ b/web/app/components/dropdown/dropdown.scss
@@ -1,0 +1,4 @@
+.dropdown {
+  display: inline-block;
+  position: relative;
+}
--- a/web/app/testUtils/mockStyles.js
+++ b/web/app/testUtils/mockStyles.js
@@ -1,0 +1,1 @@
+module.exports = {};
--- a/web/app/utils/email.js
+++ b/web/app/utils/email.js
@@ -27,7 +27,7 @@ link: ${link}
 }
 
 export function requestDeletion() {
-  deleteMe().then(data => {
+  return deleteMe().then(data => {
     const email = store.get('config').admin_email;
     const { subject, message } = getDeleteInformationMessage(data.user_id, siteId, data.link);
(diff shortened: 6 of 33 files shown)


umputun:
Thx. 

I don't think move "sign out" to this new menu/popup is a good idea. This menu for rare and exotic actions, like "delete my data" (probably sounds too scary, "request my data removal" could be better), "export all my data" and probably other actions like this. Logout is a very different beast, and it will be better to keep it visible as today. In addition, maintaining login/logout action one the same visual level makes more sense.

steppefox:
Ok, i'll return Sign out on its previous place. But I decide to do that because of same popup in Disqus. Logout is a rare case too.

Also going to fix my codestyle in PR to pass CI steps.

steppefox:
@umputun Why do we need to send any POST request to the API? If we don't use `token` and on the frontend-side I already have necessary data (user_id, link, site_id). For logging/statistics purpose?

umputun:
@steppefox I don't see much of technical difference between POST and GET for `/deleteme`, but `POST` method felt more appropriate for such action especially because it will "post" email message as a result. Maybe `DELETE` will make some sense here, however, because this is not immediate delete but rather "submit request" I still think `POST` will be the right thing to do.

I guess by "any POST request to the API" you meant just `/deleteme`? In case if you meant other posts - many of them send body.

Regarding "we don't use token" and "already have ... user_id" - this is not exactly accurate. UI doesn't send a token, but the browser does (as jwt cookie). Accepting user_id from the client is not a good idea from security point of view.

Guria:
@umputun I guess @steppefox asked not about GET vs POST. The question was why we need to make request at all if all necessary data for email template is already exists on client side.

steppefox:
@umputun I meant that we don't need to send `/deleteme` request at all. We have all the information to generate email on the frontend-side without any API request. We already store the user_id in memory and use it to understand that user is logged in.
What kind of security reasons do you mean? Like fake `delete me` email with someones ID?

umputun:
Ah, I see what you mean. The goal of this request is to retrieve user's token to allow the actual delete (`/admin/deleteme`) to safely get user_id and site_id from that token. Pls, note - the frontend has no access to the token as it stored as httpOnly cookie. The backend needs this token to validate the actual request (`link` field you get from `POST /deleteme`) and to extract user_id and site_id from that token. Btw, user_id not even used for the actual removal, the only goal of having it in email is the audit trail. 

All of this needed to prevent fake `deleteme` requests.

steppefox:
Screenshot with the fix:

image

web/app/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+  componentWillUnmount() {
+    document.removeEventListener('click', this.onOutsideClick);
+    parent && parent.document.removeEventListener('click', this.onOutsideClick);
+  }
+
+  render(props, { isActive }) {
+    const { title, heading, children } = props;
+
+    return (
+      <div className={b('dropdown', { mods: { isActive }})} ref={this.bindRef('_dropdown')}>
```
igoradamenko: Let's use clearer code for refs? Like `ref={r => (this.rootNode = r)}`. 

I mean, right now you can't understand this code without reading the `bindRef` definition. But it's the strange wrapper and you can omit it, can't you?

web/app/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+export default class Dropdown extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isActive: props.isActive || false,
+    };
+
+    this.onClickTitle = this.onClickTitle.bind(this);
```
igoradamenko: `onTitleClick`?

web/app/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+      isActive: !this.state.isActive
+    });
+
+    if (this.props.onClickTitle) {
+      this.props.onClickTitle();
+    }
+  }
+
+  onOutsideClick(e) {
+    if (!this._dropdown.contains(e.target) && e.target !== this._dropdown) {
```
igoradamenko: Are you sure that you need the second condition here? `contains` checks the node itself, doesn't it?

https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

web/app/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+        this.setState({
+          isActive: false,
+        });
+      }
+    }
+  }
+
+  componentDidMount() {
+    document.addEventListener('click', this.onOutsideClick);
+    parent && parent.document.addEventListener('click', this.onOutsideClick);
```
igoradamenko: Can you use full conditions with `if` etc, please?

web/app/components/dropdown/dropdown.jsx
igoradamenko: The same story about BEM-helper, as for button.

web/app/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+    parent && parent.document.removeEventListener('click', this.onOutsideClick);
+  }
+
+  render(props, { isActive }) {
+    const { title, heading, children } = props;
+
+    return (
+      <div className={b('dropdown', { mods: { isActive }})} ref={this.bindRef('_dropdown')}>
+        <Button
+
```
igoradamenko: Let's remove empty line here and below. I don't sure that they add readability, but they look strange.

web/app/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+          aria-haspopup="listbox"
+          aria-expanded={isActive && 'true'}
+
+          className={b('dropdown__title')}
```
igoradamenko: You don't need BEM-helper here. I think you want to use `mix="dropdown__title"`.

web/app/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,81 @@
...
+          type="button"
+          onClick={this.onClickTitle}
+        >
+          {title}
+        </Button>
+
+        <div className={b('dropdown__content')} tabindex="-1" role="listbox">
```
igoradamenko: The same story. You don't need to use `b` if you don't pass any other arguments but first one. Here you can write just `className="dropdown__content"`.

web/app/components/dropdown/dropdown.scss
```diff
@@ -0,0 +1,32 @@
+.dropdown {
+  display: inline-block;
+  position: relative;
+
+  &__title {
```
igoradamenko: Also, let's move elements' styles to separated files.

web/app/components/dropdown/dropdown.scss
```diff
@@ -0,0 +1,32 @@
...
+    left: 50%;
+    transform: translate(-50%, 5px);
+    width: 170px;
+    background-color: #fff;
+    border: 2px solid #259c9a;
+    border-radius: 3px;
+    padding: 0 0 5px;
+  }
+
+  &_is-active {
```
igoradamenko: `active`.

web/app/utils/email.js
```diff
@@ -0,0 +1,31 @@
...
+  const message = encodeURIComponent(`Request to delete all information about ${userId} from remark42 on ${siteId}
+
+[you can provide the reason for removal request, optional]
+
+=== DO NOT REMOVE THE TEXT BELOW THIS LINE ===
+
+site: ${siteId}
+user: ${userId}
+link: ${link}
+`).replace('%0A', '%0D%0A');
```
igoradamenko: Can you explain this replacement, please? I think you need to add comment about it.

web/app/testUtils/mockStyles.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
igoradamenko: What does this file do?

web/app/components/dropdown/dropdown.jsx
steppefox: Nope, `e.target` could be any children html element too

web/app/components/button/button.jsx
```diff
@@ -0,0 +1,71 @@
...
+export default class Button extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isClicked: false,
+      isFocused: false,
+    };
+
+    this.onMouseDown = this.onMouseDown.bind(this);
```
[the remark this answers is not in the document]
steppefox: I've removed callback prop, but kept the method, because it's necessary to handle state

web/app/components/dropdown/__item/dropdown__item.scss
```diff
@@ -0,0 +1,8 @@
+.dropdown__item {
+  a, button {
```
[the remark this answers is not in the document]
steppefox: It's necessary to make a clickable area with the whole width of the dropdown. I guess it's fine here, but i can change the rule to: `> a, > button`.

web/app/testUtils/mockStyles.js
steppefox: Mocks imports of scss files during the tests

Guria: it is noop for importing styles in jest environment

umputun:
I have tried this PR and found two issues:

1. Opening email client (click on  "Request my data removal") does nothing on safari and complains about blocked popup on chrome. As soon as I enabled popups on chrome it worked. Probably regular user won't figure this out. Can we eliminate popup?
1. Some auth mismatch as admin trying to click on the link. Most likely something on the backend side, will take a look

steppefox:
I'll check the conflicting files and try to avoid popup issue

Guria:
@umputun rebased and introduced deleteme.html page for admin. Popup issue should be also fixed.
Test it again, please.

umputun:
it works. two minor additions:

- redirect via deleteme.html adds &site - no need, this info already part of the generated token
- we want to enforce logout at the same time to clean cookies

Issue #98, opened 2018-06-23
Move user info to separated iframe


igoradamenko:
If someone clicks on user's name, they will see user's info:

<img width="760" alt="screen shot 2018-06-23 at 14 12 39" src="https://user-images.githubusercontent.com/6537798/41809033-9db82070-76ef-11e8-96ad-81429456709a.png">

But it looks awful. We did it because it was a quick way to implement this feature, but it must be better. 

So I think that we need to open user's info in separated block, like Disqus does:

<img width="1280" alt="screen shot 2018-06-23 at 14 15 11" src="https://user-images.githubusercontent.com/6537798/41809062-e1307460-76ef-11e8-8238-3d8448e4611f.png">

But it's impossible to do inside the main iframe. So we need to create a separated iframe with user's info dynamically, and paste it to original page. And this iframe must have possibility to handle some actions like voting for comments, etc. 

I'm pretty sure that it isn't an easy task, so you are welcome to create draft PR which will be improved later.

Menelion:
@igoradamenko, what I'm going to ask is a bit off topic, but still related: why did you decide to go for iframes for the comment widget itself? Why not a `div` with AJAX-powered content, for example?

igoradamenko:
@Menelion, because of CORS. When we use iframe we can do requests to the server without any problems (because our iframe is served by our server). We can't do PUT, POST, etc with CORS using XHR (AJAX) from the domain which is not allowed by policy.

Yeah, actually I think it's possible to change policy, but there are a lot of other problems, like HTTP / HTTPS restrictions, subdomains policy, etc. So if we want to build a widget-like product, we need to use iframe.

If we talk exactly about this task, I think it's possible to use div-ish system, but we need to transfer all actions from user-info block to our iframe, and then send requests to server from iframe.

So, I think that it's easier to create another one iframe-ish widget for user-info. But I could be wrong in some details, so if anyone thinks that it could be better not to use iframes at all, and they can create PoC PR — they are welcome :-)

iegik:
Because of ifreme, `user-info` window can be positioned relatively to the ifreme, noir whole page window. As alternative, we can provide some of "iframe event bridge API", where host page can receive information about users from remark and use for ~their own risk~ displaying own popup or information window.
https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

umputun:
this implementation has some issues

[...]

PR #110, opened 2018-06-26
Add: user-info in iframe Issue #98

lexich:
Issue #98 

Hi @igoradamenko Pls check my PR. I'm only moved user-info component to iframe with minimal UI changes.

screen shot 2018-06-26 at 11 54 25
video


Guria:
Had to revert this PR due to styling issues:
image

Also I think we can avoid manual creating of DOM node for user iframe.
Just create it dynamically in embed.js

Issue #107, opened 2018-06-24
Complete transition to JWT and get rid of direct checks in controllers


umputun:
After switching from sessions to JWT, the code has some leftovers from the session's days. For example, it checks directly (in controllers) for blocked users, does extra checks for auth permission and verifies flag. All of this should be switched to jwt, propagated in login controller and validated statelessly (against claims) in auth middleware.

umputun:
The change should include:

1. short (5-10 minutes) JWT expiration. 
1. automatic refresh allowing even expired JWT to be used as a base for refreshed tokens
1. refresh should update not just `exp` field but also user's flags, like `admin`, `verified`, `banned` and so on
1. all controllers should count on the token (from request context) and don't do any extra checks directly
1. blocked users should fail auth on middleware level
1. cookie TTL defines max inactivity interval and should be decreased from 1y to 1w

Issue #106, opened 2018-06-24
Implement and document fully functional, local development environment


umputun:
To test (end-to-end) functionality locally, contributors need a simple way to deploy both backend and frontend. Currently, we have some description of how to bring local backend and emulate logged in users; however, this is used mostly by contributors of the backend side.

To make such local server frontend–friendly I'm going to add a new auth provider with two dev users:
- `devreg` (regular user permission) and `devauth` (admin permission). 

Then will need to describe how to setup such environment locally.

umputun:
Added minimal but functional stand-alone setup. To bring it up run

```
docker-compose -f compose-dev.yml build
docker-compose -f compose-dev.yml up
```

It starts remark42 on http://127.0.0.1:8080 and adds local oauth2 provider "Dev". To access UI demo page http://127.0.0.1:8080/web

That `compose-dev.yml`also defines if logged user admin or not. By default, it will be the admin, and to switch it to regular user comment/remove `-ADMIN=dev_user`.

umputun:
added an ugly form to select dev user in order to switch between admin and regular users.

umputun:
@igoradamenko - you planned to add this info to README. Feel free to close this one as you done.

igoradamenko:
@umputun, thank you for the reminder. I've added info about this testing env to the readme.

Guria:
I tried to build on docker windows. It seems went ok. But I can't run it now:
```
PS C:\ws\misc\remark> docker-compose -f .\compose-dev.yml up
Starting remark42-dev ... done
Attaching to remark42-dev
remark42-dev | init
remark42-dev | execute /srv/start.sh
remark42-dev | prepare environment
remark42-dev | start remark42 server
remark42-dev | Usage: /sbin/su-exec user-spec command [args]
remark42-dev exited with code 0
```

umputun:
most likely this is because you didn't define `USER` uid in your env. I have no clue about uid on win but the simplest way will be just to set it directly in `compose-dev.yml`, i.e. replace `- USER` by `- USER=1001`

umputun:
for https://github.com/umputun/remark/issues/98#issuecomment-403754560 need to add the ability to show/set/select avatar

umputun:
Select/upload avatar is to much trouble for dev user. Instead I have added GitHub-style automatic avatars. like this:

[...]

PR #113, opened 2018-06-27
feature/jwtcleanup

umputun:
#106 and #107

### DIFF
--- a/backend/app/main_test.go
+++ b/backend/app/main_test.go
@@ -18,7 +18,11 @@ import (
 )
 
 func TestApplication(t *testing.T) {
-	app, ctx := prepApp(t, 18080, 500*time.Millisecond)
+	app, ctx := prepApp(t, 500*time.Millisecond, func(o Opts) Opts {
+		o.Port = 18080
+		return o
+	})
+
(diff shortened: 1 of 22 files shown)


backend/app/main_test.go
```diff
@@ -44,12 +48,37 @@ func TestApplication(t *testing.T) {
 	app.Wait()
 }
 
+func TestApplicationDevMode(t *testing.T) {
```
golangci: `TestApplicationDevMode` is unused

backend/app/rest/auth/dev_provider_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// auth routes for all providers
+	router := chi.NewRouter()
+	router.Route("/auth", func(r chi.Router) {
+		r.Mount("/dev", srv.Provider.Routes()) // mount auth providers as /auth/{name}
+	})
+
+	ts := &http.Server{Addr: fmt.Sprintf("127.0.0.1:%d", 8080), Handler: router}
+	go srv.Run()
+	go ts.ListenAndServe()
```
golangci: Error return value of `ts.ListenAndServe` is not checked

PR #116, opened 2018-06-29
feature/subtree

### LINKED ISSUE #114: Hide fully deleted subtree
In this case there is no need to show deleted comments
grpub-201806-28095323-xiaf5

Currently, we do such filtering on the server side for the top-level trees only, and it will be nice to have the same processing for any given subtree.

umputun:
fix #114


backend/app/rest/tree_test.go
```diff
@@ -125,272 +158,13 @@ func BenchmarkTree(b *testing.B) {
...
-
-	actual = strings.Replace(actual, "\n", " ", -1)
-	actual = reSpaces.ReplaceAllString(actual, " ")
-	return expected, actual
+// loadJsonFile read fixtrue file and clear any custom json formatting
+func mustLoadJSONFile(t *testing.T, file string) []byte {
+	expJSON, err := ioutil.ReadFile(file)
+	require.Nil(t, err)
+	expTree := Tree{}
+	err = json.Unmarshal(expJSON, &expTree)
```
golangci: ineffectual assignment to `err`

Issue #92, opened 2018-06-23
Add linter for SCSS


igoradamenko:
I think it would be nice to use Stylelint:

https://github.com/stylelint/stylelint

matrixik:
Maybe also using https://prettier.io/ would be nice.

Guria:
Also want to suggest prettier for js codebase as well.
There is great setup integrated with eslint.
I have experince in such setup in many projects

PR #117, opened 2018-06-30
Add Prettier to js & scss files Issue #92

lexich:
There are 2 commits.
1. First commit is configuration for prettier and eslint config fix.
2. Second reformat code
Maybe the best way to drop second commit it there will be conflicts


Guria:
Few suggestions here:
* collect all config/setup changes in one commit without reformat applied
* apply reformat in separate commit with faked author like prettier@remark42.com
* add husky with precommit script to eslint -fix. Though not sure @umputun would be happy to have nodejs in his git hooks :)

lexich:
@Guria
If 'husky' will be ok (but I agree with possible @umputun reaction) good way to add 'lint-staged'

All configs is in separate config at this moment

umputun:
Not sure I got it, but If this husky thing enforces precommit checks and will make poor backend developers to install nodejs – sure I don't want it. Currently, all build-related dependencies isolated in Dockerfile and we want to keep it this way.

lexich:
@umputun I understand you worries and It’s the main reason why there isn’t `husky` in this PR

Guria:
Though, it wouldn't affect lucky backend developers since build-related things is isolated.
`husky` will setup nodejs scripts into hooks only if developer ran `npm install` in local environment.
But that means that this developer has nodejs and wants to edit frontend files for some reason.

igoradamenko:
I agree with @Guria and think that we need to setup lint-staged, because it looks safe for backend-guys. If some of them install frontend deps, it looks that they want to develop frontend part. Also, lint-staged allows to configure linters using globs, so we can configure it only for frontend's files, such as .js, .jsx, .scss, etc.

lexich:
@igoradamenko add lint-staged and husky

Guria:
The only problem is other pending PRs. Either we should merge this first and then "rebase" all pending PRs or drop reformatting commit, set prettier/prettier rule to warn in eslint and let husky+lint-staged do the job in iterations.

lexich:
@Guria I agree with your point. Unfortunately PRs appling is very slowly. I have 2 ready PRs and nothing happens.

Reformatting commit we'll drop, but after that it needs to fix code for eslint approving.

igoradamenko:
OK, @Guria, now you can merge this PR whenever you prefer.

lexich:
Closed by #132

Issue #119, opened 2018-06-30
Vendor build dependencies


umputun:
Sometimes we have failed builds due to third-party modules/services. For example

```
Step 27/43 : RUN     cd /srv/web &&     npm i && npm run lint && npm run test && npm run build &&     rm -rf ./node_modules
 ---> Running in bd12192eb9f5
npm ERR! code EADDRNOTAVAIL
npm ERR! errno EADDRNOTAVAIL
npm ERR! request to https://registry.npmjs.org/chalk failed, reason: connect EADDRNOTAVAIL 2400:cb00:2048:1::6812:5f60:443 - Local (:::0)
```

This happens mostly on frontend build, but not always. Sometimes backend fails as it can't reach coveralls.

I would like to have our build dependencies fully vendored, i.e. it should be able to build everything without the internet connection. All optional steps, like coveralls, should not fail the build.

umputun:
Coveralls was the only external dependency needed to run `build-backend` stage, made optional.

umputun:
After discussion with @igoradamenko it sounds like such vendoring in js/npm world is not something usual unless proxy involved.

Guria:
See #123

PR #123, opened 2018-07-01
reveal node_modules in separate container

Guria:
This allows to reuse docker fs layer with node_modules
in case if there were no changes in package*.json
As a result it decreases dependency on network and build times
in rebuilds at least in local environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,18 +49,26 @@ RUN \
     go build -o remark -ldflags "-X main.revision=${version} -s -w" ./app
 
 
-FROM node:9.4-alpine as build-frontend
+FROM node:10.6-alpine as build-frontend-deps
+
+ARG CI
+
+RUN apk add --no-cache --update git
+ADD web/package.json /srv/web/package.json 
+ADD web/package-lock.json /srv/web/package-lock.json
+RUN cd /srv/web && npm ci
+
+FROM node:10.6-alpine as build-frontend
 
 ARG CI
 ARG SKIP_FRONTEND_TEST
 
 ADD web /srv/web
-RUN apk add --no-cache --update git
-RUN \
-    cd /srv/web && npm i && \
-    if [ -z "$SKIP_FRONTEND_TEST" ] ; then npm run lint && npm run test ; \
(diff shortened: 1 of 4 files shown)


Guria:
related to https://github.com/umputun/remark/issues/119#issuecomment-401595070

Guria:
To get fs layer reuse in CI we might tweak our setup like described at https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-248915326

umputun:
This is an interesting idea, but I'm not sure about that travis part. You can add travis's change to this PR, and we could see how it works. The second part - we will need smth similar for drone. And the last part I don't know if such complexity even justified. Initially, in #119 I hoped to make frontend build dependency vendored directly, which seems to be non-trivial. However, if the change also decreases build time (significantly), it could be another good reason,

Guria:
@umputun Is there a way to see build log from drone.io? I tried to open link, but it says just Loading. Login attempt is also failed. Btw, what it used for?

It seems like drone.io have better docker integration and maybe it would be simpler setup.

Guria:
It seems travis config didn't helped.

umputun:
@Guria I think travis won't allow PR change for the `.travis.yml` due to security concerns. The only way to test it is your own travis against forked repo.

Guria:
@umputun after series of experiments I decided to stop experimenting with enabling docker fs layer caching on CI. It seems like overhead on cache packing/unpacking is more then potential win in build time.
I also introduced custom image with git and cached binary for node-sass. It used only in step that collects node_modules. Build container is still based on official image.

umputun:
> I also introduced custom image with git and cached binary for node-sass

I don't like the new dependency on `guria/build-node:latest` as a part of our build process. If this image base-enough to be added to  https://github.com/umputun/baseimage - pls PR to that repo.

Guria:
@umputun  removed custom dependency since we don't rely on node-sass binary anymore

This doesn't improve anything on CI. But it's very useful for local development since it speeds up build times and saves network bandwidth, especially on limited mobile connections.

umputun:
LGTM

Issue #111, opened 2018-06-26
As a user, I want to be notified about actions on my comments


misterioss:
I really miss features from Disqus where I can see:
- someone replied to my comment
- someone upvoted my comments (upvotes summary, not every click on sign "+")

Guria:
Hmm, remark doesn't have any ways to sent you notifications as far as I can see.

umputun:
@Guria the only notification you can get now are RSS based, i.e. pull, for site updates and post updates. You can feed them into IFTTT and make tweets, telegram, slack messages or anything else you like.

@misterioss probably I can add "user RSS" to report replies, but votes summary is more complicated, and I don't think anyone actually doing it. I don't recall any notification on votes from disqus, reddit or hackernews.

also, see https://github.com/umputun/remark/issues/15

akosourov:
@umputun I have a few questions, it relates what we talked at #125 
1. For example a user posts a comment. Then somebody replies him with comment `A`. And user wants to be notified, that is ok. But if another one replies to comment `A` (not to initial comment) what should happen? Should remark notify user or not? This is reply to user's reply
2. I think remark should notify user from replies to all his comments, not for last N. I mean it will better :) what do you think?

umputun:
> For example a user posts a comment. Then somebody replies him with comment A. And user wants to be notified, that is ok. But if another one replies to comment A (not to initial comment) what should happen? Should remark notify user or not? This is reply to user's reply

We can simplify this thing logically and technically if we consider direct replies only. In this case, we could match by parent-id (pid) which should be relatively straightforward. 

> I think remark should notify user from replies to all his comments, not for last N. I mean it will better :) what do you think?

RSS will be limited anyway. If the user got 1000 replies during his rss-update window we won't be able to serve such RSS with 1000 items. I don't think this will be any real issue if we set max. items for such rss to 100  and hopefully client-side (rss reader, ifttt, etc..) will pull it every 5 minutes.

akosourov:
@umputun First one is ok, I got it.

> I think remark should notify user from replies to all his comments, not for last N. I mean it will better :) what do you think?

I mean about search implementation details. I can get last `K` user's comments from `users` bucket and then find replies for them from `posts` bucket. Then sort and limit up to `maxRssItems`. But if a last reply was added to `K+1` user's comment this query doesn't search it. And a user won't be notified. 
[...]

PR #125, opened 2018-07-01
add rss feed to replies #111

akosourov:
This feature adds /rss/reply endpoint that generates RSS feed with replies for the comment. 
Replies are sorted by time from last to first. Amount of replies is limited up to maxRssItems (20).
For sorting purpose sortComments function from engine package had to do public, so I'm not sure that it is a good solution
@umputun I'm will be glad any feedback

### DIFF
--- a/backend/app/rest/api/rss.go
+++ b/backend/app/rest/api/rss.go
@@ -127,7 +127,7 @@ func (s *Rest) rssRepliesCtrl(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/xml; charset=utf-8")
 	w.WriteHeader(http.StatusOK)
-	if _, err := w.Write([]byte(data)); err != nil {
+	if _, err := w.Write(data); err != nil {
 		log.Printf("[WARN] failed to send reponse to %s, %s", r.RemoteAddr, err)
 	}
 }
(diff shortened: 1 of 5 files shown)


backend/app/rest/api/rss.go
```diff
@@ -88,6 +90,48 @@ func (s *Rest) rssSiteCommentsCtrl(w http.ResponseWriter, r *http.Request) {
...
+	})
+
+	if err != nil {
+		rest.SendErrorJSON(w, r, http.StatusBadRequest, err, "can't find replies")
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/xml; charset=utf-8")
+	w.WriteHeader(http.StatusOK)
+	if _, err := w.Write([]byte(data)); err != nil {
```
golangci: unnecessary conversion

akosourov:
This relates to #111 in more general way. If you want to be notified when someone replied to your comment or replied to thread (comment) that you are interested in - you can subscribe to it through RSS reader.
This was the goal

umputun:
I read #111 differently. From what I see user would like to have a single "replies to all of my comments" RSS instead or resubscribing every time he commented on a different (sub)tree. I see how `/reply?user_id&site_id` could be useful here, but don't have in mind an easy way to implement it. Maybe you can try to address it in three steps:
1. get last user's comments
1. apply `findReplies` to each one. 
1. sort and remove duplicates

umputun:
pls, note - I'm not saying your version of `/rss/reply` is useless, just can't imagine the practical use case. Another thing is how to integrate it into UI. In order to expose this "per-comment" RSS each and every comment should have some "subscribe" link, which feels excessive to me especially for such exotic technology as RSS ;)

akosourov:
Ok, I agree. If you don't mind I'll try to implement `/reply?user_id&site_id` solution and come back later :)

umputun:
sure, thx

PR #132, opened 2018-07-06
Config for prettier + husky + lint-staged #92

lexich:
Separate prettier-config for PR #117


Guria:
Prettier rules in eslint should be set to warn for now.

Guria:
Change it to warn for now. It will allow us to not broke others PRs.
Then we will get incremental reformating in new PRs by husky.
Once I'll manage most of PRs we will switch it back to error and reformat rest of code

Issue #88, opened 2018-06-20
Add ability to ban user for limited time


umputun:
From the experience of running remark42 in the real life, I have seen a need to ban a user temporarily. We don't have any automatic way to do it, but if we extend `store.BlockedUser` with TTL and add support of such TTL to `IsBlocked` this will do it.   

In addition, we will need UI to support such selection of ban-type as well as an indication of ban type and TTL in the list of blocked users

umputun:
The backend change completed. For frontend we need a couple of minor changes:

1. Ability to select ban duration. Could be just a list of "permanent, 1 day, 7 days, 30 days."
1. In the list of banned users show this info as "banned until." 

@igoradamenko - feel free to add necessary info for contributors if my description incomplete or confusing. Probably you can set it to "first good issue" if this is really as simple as I think.

jackburg:
Господа хорошие, позвольте мне закрыть данный ишью.

Guria:
@jackburg Use English please.

If I correctly understand, you want to make a UI part of this, right?
Feel free to make a PR.

jackburg:
Ok got it.

umputun:
Looks good to me

PR #140, opened 2018-07-08
Addressed to issue #88.

jackburg:
Add ability to ban users for limited time and to view in the 'Show blocked'
list time util any specific used was blocked.


web/app/components/comment/__control/_block/comment__control_block.scss
```diff
@@ -0,0 +1,19 @@
...
+  position: relative;
+  white-space: nowrap;
+  font-weight: 700;
+  &::after {
+    content: '▾';
+    margin-left: 2px;
+  }
+}
+
+.comment__control_block__select {
```
Guria: It isn't a BEM. Elements of elements are prohibited.

web/app/components/comment/comment.jsx
```diff
@@ -164,17 +166,38 @@ export default class Comment extends Component {
...
-  toggleBlock(isBlocked) {
+  onBlockUserClick(e) {
     const {
       id,
       user: { id: userId },
     } = this.props.data;
-    const promptMessage = `Do you want to ${isBlocked ? 'unblock' : 'block'} this user?`;
 
+    const ttl = e.target.value;
+    const duration = getBlockArray().filter(el => el.value === ttl)[0].label;
```
Guria: `Array` has convenient `find` method.

web/app/components/comment/comment.jsx
```diff
@@ -560,3 +603,26 @@ function formatTime(time) {
 
   return `${date} at ${hours}:${mins}`;
 }
+
+function getBlockArray() {
```
Guria: No need in a function here. It could be declared just as constant in a corresponding file.

web/app/components/comment/comment.jsx
```diff
@@ -560,3 +603,26 @@ function formatTime(time) {
...
+  const blockArray = [
```
Guria: Use something more descriptive as a variable name like `blockingDurations`.

web/app/components/comment/__control/_block/comment__control_block.scss
jackburg: How it should be named?

Guria: Usually, such names indicate that it is time to extract this part as an independent block.
I have seen the same pattern in auth-panel__sort, which seems you used as a reference.
So it is another reason to create a new component.

jackburg: Thanks, I'll take a look

web/app/components/comment/__control/_select-label/comment__control_select-label.scss
```diff
@@ -0,0 +1,9 @@
+.comment__control_select {
+  position: absolute;
+  left: 0;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  opacity: 0;
+  cursor: pointer;
```
Guria: Please add width: 100% here. For some reason, select styled like this still has a width more than its container in Chrome. See https://github.com/umputun/remark/issues/80#issuecomment-403213881

web/app/components/comment/comment.jsx
```diff
@@ -299,6 +325,8 @@ export default class Comment extends Component {
     const config = store.get('config') || {};
     const lowCommentScore = config.low_score;
 
+    const blockingOptions = blockingDurations;
```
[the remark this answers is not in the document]
jackburg: blockingDurations var in not accessible in the template renderer, so I've create a local var (blockingOptions) here to get access to it values, could you propose any other solution?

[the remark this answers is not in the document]
Guria: Move it to constants file as I already suggested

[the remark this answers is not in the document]
jackburg: It's done

web/app/components/comment/comment.jsx
```diff
@@ -501,9 +527,25 @@ export default class Comment extends Component {
...
+                  {userBlocked && (
+                    <span {...getHandleClickProps(() => this.onUnblockUserClick())} className="comment__control">
+                      Unblock
+                    </span>
+                  )}
+
+                  {!userBlocked && (
+                    <span className="comment__control comment__control_select-label">
+                      Block
+                      <select className="comment__control_select" onChange={this.onBlockUserClick}>
```
Guria: Linter asks to use onBlur here instead of onChange.
See https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md

Guria: ok, add `/* eslint-disable jsx-a11y/no-onchange */` and create a separate issue to further investigate

jackburg: It's done. I've opened #148 issue in regard of this situation.

Issue #95, opened 2018-06-23
Refactor data store on frontend


igoradamenko:
On frontend we use Store class which is defined here:

https://github.com/umputun/remark/blob/master/web/app/common/store.js

It's a regular singleton, which implements some logic for getting / setting data and adds availability to listen changes. I didn't want to use something like Redux, because we used Preact and it would be strange to bloat our frontend with a huge library because of a simple store abstraction.

But I'm pretty sure that it's possible to rewrite this implementation to make it clearer. I think we should try to do it.

Guria:
Redux is [not so bloated in fact](https://bundlephobia.com/result?p=redux@4.0.0). But it isn't limited as just state store. It could bring much more. Devtools support will help better create mental image of the app. It opens a lot of render optimizations for free. It also helps to better decouple components #96 .
It will also make contributions for newcomers easier.

igoradamenko:
> Redux is not so bloated in fact.

It's true for SPA, but we create a widget and we need to try keep them as small as possible.

I understand cons and pros of Redux, but IDK why we need it here. Do you have any ideas? Why we need redux-like store?

igoradamenko:
Also don't forget about bindings for redux — preact-redux:

![image](https://user-images.githubusercontent.com/6537798/41812641-75b56f02-772f-11e8-93f4-72b7cc4589c9.png)

So, these two have size which equals to whole remark42 package (except polyfills: #102).

Guria:
It still just 1/4 of babel-polyfill. :) I will return to this issue right after #102

Guria:
> Why we need redux-like store?

It just too hard to pass-through state managed in few root components to multiple nested childs by props. It's hard to manage and hard to build a mental image of the app in the head.

Redux enables us the way to pass data directly to components and track what is happening in the app.
And I am sure it is crucial for 2 important things:
* easier achieve #96 
* get more contributors

What about bundle sizes. #102 already shrinks a lot of space in main remark.js bundle. We can go further and split the vendor bundle to contain preact, axios, redux and preact-redux. It will then remove duplication in existing bundles. So in sum we will be still in the win.

Guria:
A bit of stats with minified uncompressed sizes:

current:

Length | Name
------ | ----
 20310 | counter.js
  2673 | embed.js
 57366 | last-comments.js
144121 | remark.js
224470 | Total

with #102

Length | Name
------ | ----
 20298 | counter.js
  2673 | embed.js
 57272 | last-comments.js
[...]

PR #141, opened 2018-07-08
Basic redux integration for thread component

DmitryTsepelev:
Here is a very basic integration with redux. I've moved all the complex logic from the Thread component to the reducer, which is also responsible for restoring the initial state from the local storage. Reducer is already covered with a couple of specs, would be also useful to have a spec for getters. Related to #95 and #96

### DIFF
--- a/web/app/components/thread/thread.jsx
+++ b/web/app/components/thread/thread.jsx
@@ -1,63 +1,24 @@
 /** @jsx h */
 import { h, Component } from 'preact';
-
-import { LS_COLLAPSE_KEY } from 'common/constants';
-import { siteId, url } from 'common/settings';
-import store from 'common/store';
+import { connect } from 'preact-redux';
 
 import Comment from 'components/comment';
+import { setCollapse } from './thread.actions';
+import { getThreadIsCollapsed } from './thread.getters';
 
-export default class Thread extends Component {
+class Thread extends Component {
   constructor(props) {
     super(props);
-
-    if (this.props.data && this.props.data.comment) {
-      this.updateCollapsedState(this.props.data.comment);
-    }
-
     this.onCollapseToggle = this.onCollapseToggle.bind(this);
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.data && nextProps.data.comment) {
-      this.updateCollapsedState(nextProps.data.comment);
-    }
-  }
-
-  updateCollapsedState(comment) {
-    const config = store.get('config') || {};
-    const score = comment.score || 0;
-
-    this.lsCollapsedID = `${siteId}_${url}_${comment.id}`;
-
-    this.state = {
-      collapsed:
-        (!this.state.isCollapsedChanged && score <= config.critical_score) ||
-        getCollapsedComments().includes(this.lsCollapsedID),
-      isCollapsedChanged: true,
-    };
-  }
-
   onCollapseToggle() {
-    const collapsed = !this.state.collapsed;
-
-    this.setState({ collapsed: !this.state.collapsed });
-
-    let collapsedComments = getCollapsedComments();
-
-    if (collapsed) {
-      if (!collapsedComments.includes(this.lsCollapsedID)) {
-        collapsedComments = collapsedComments.concat(this.lsCollapsedID);
-      }
-    } else {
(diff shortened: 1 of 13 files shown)


web/app/components/thread/thread.jsx
```diff
@@ -1,67 +1,41 @@
...
+  render(props) {
     const {
+      collapsed,
       data: { comment, replies = [] },
       mods = {},
     } = props;
 
+    let threadComments = null;
+    if (!collapsed && !!replies.length) {
+      threadComments = replies.map(thread => (
```
Guria: I prefer inline variant. Please put it back to render

web/app/components/thread/thread.reducers.js
```diff
@@ -0,0 +1,43 @@
...
+  saveCollapsedComments(collapsedComments);
+};
+
+export const collapsedThreads = (state = initialState, action) => {
+  switch (action.type) {
+    case THREAD_SET_COLLAPSE: {
+      const currentCollapsed = state[action.comment.id];
+
+      if (action.collapsed !== currentCollapsed) {
+        persistCollapsedComments(action.comment, action.collapsed);
```
Guria: Reducers must be pure functions. No side effects allowed.

web/app/components/thread/thread.jsx
```diff
@@ -76,26 +50,21 @@ export default class Thread extends Component {
...
-  localStorage.setItem(LS_COLLAPSE_KEY, JSON.stringify(comments));
-}
+const mapDispatchToProps = { setCollapse };
+
+const enhance = connect(
+  mapStateToProps,
+  mapDispatchToProps
+);
+
+export default enhance(Thread);
```
Guria: I also prefer inline declarations here as well:
```js
export default connect(
  (state, props) => ({
    collapsed: getThreadIsCollapsed(state, props.data.comment)
  }),
  { setCollapse }
)(Thread);
```

web/app/components/thread/thread.reducers.js
Guria: We can make custom middleware to persist to localStorage. Using localStorage value for initial value is ok.

web/app/components/thread/thread.jsx
DmitryTsepelev: It will become unreadable as soon as we wrap it up with one more HOC :)

DmitryTsepelev: ✅

web/app/components/thread/thread.jsx
DmitryTsepelev: ✅

web/app/components/thread/thread.reducers.js
DmitryTsepelev: Agreed, done. If there will be more cases like this - we could try handling all the side effects with sagas

DmitryTsepelev:
@Guria Thanks for the review, I think I'm done. Please let me know if we need to change something else

web/app/store.js
```diff
@@ -0,0 +1,35 @@
...
+import { combineReducers } from 'redux';
+
+import { siteId, url } from 'common/settings';
+import { threadReducers, THREAD_SET_COLLAPSE, getCollapsedComments, saveCollapsedComments } from './components/thread';
+
+const rootReducer = combineReducers({
+  ...threadReducers,
+});
+
+const collapsedCommentsMiddleware = ({ getState }) => next => action => {
```
Guria: maybe export middleware from thread as well?

DmitryTsepelev:
@Guria I'm stuck with the devtools :( I've configured everything according to the docs, aliased all the react-related things, but redux dev tool does not see the store, while react dev tool is "waiting for react" or "waiting for the roots to load", what am I missing here?

Guria:
Ok, nevermind. We can do it in a separate PR. But we need to resolve new conflicts here.

DmitryTsepelev:
Done

DmitryTsepelev:
@Guria Are we good to go? :) I have one more PR to submit

Guria:
@DmitryTsepelev please don't use yarn. only npm 6.1.0+

DmitryTsepelev:
ah ok, got it

Issue #135, opened 2018-07-06
Migrate from node-sass to postcss with nesting plugin.


Guria:
SASS features is not heavy used as I can observe.
Only nesting syntax is used and it could be easily achieved with postcss plugin.
Also I believe it is downloading of node-sass binary causes eventual build slowdowns mentioned in #133

thepocp:
I think nesting plugin wrong solution because:
```
NOTICE: In v7.0.0 nesting at-rules like @media will no longer work, as they are not part of the nesting proposal.
```

Guria:
@thepocp I already tested with https://github.com/postcss/postcss-nested 
Will bring it soon

lexich:
Maybe the better way will be to use http://cssnext.io ?

jackburg:
@lexich Looks promising.

Guria:
@lexich @jackburg cssnext is deprecated and considered a bad practice. for now we just need to get rid of sass binary with minimal efforts and @thepocp already provided a pr #142 for that

jackburg:
@Guria thanks for the clarification.
BTW: there had been not enough next in the cssnext and it was deprecated. :)

PR #142, opened 2018-07-08
Migrate from sass to PostCSS

thepocp:
#135


web/package.json
```diff
@@ -45,8 +45,12 @@
     "jest": "^23.1.0",
     "lint-staged": "^7.2.0",
     "node-sass": "sass/node-sass#v5",
```
Guria: This should be removed.

web/app/components/input/__field/input__field.scss
```diff
@@ -3,7 +3,7 @@
   $fontSize: 16px;
   $paddingVrt: 10px;
   $lines: 3;
-  $height: $fontSize * $lineHeight * $lines + $paddingVrt * 2;
+  $height: calc($fontSize * $lineHeight * $lines+$paddingVrt * 2);
```
Guria: Isn't it just works with simple-vars plugin?

thepocp: No. We get this:
```
height:16px*1.4*3 + 10px*2;
```

PR #153, opened 2018-07-10
UserInfo uses redux store

DmitryTsepelev:
Related to #95 and #96. Changes:

1. UserInfo uses redux store (test coverage included)
2. Better test coverage for threads
3. Ability to check coverage stats using `npm run test:coverage` (let me know if we want `--coverage` flag to be turned on by default)


Guria:
Hmm. It seems like we get 2 separate redux stores :(

DmitryTsepelev:
I'm going to replace the current store soon, just want to do it in a smaller steps to avoid conflicts, do you think it makes sense to try make a bigger PR?

Guria:
My message was not about the current store, but about having 2 instances of redux store in a different iframes. One for comments list and another for UserInfo.

DmitryTsepelev:
I see your point, we're using the same store for two iframes and in the each of them we need only part of it. Maybe it makes sense to make two functions preparing store for each iframe?

DmitryTsepelev:
@Guria do you have any suggestions on how I should make it work? Thanks in advance

Guria:
I need some spare time to think about it

DmitryTsepelev:
👍

igoradamenko:
Let's just merge it and then decide what to do with all that stores :-)

DmitryTsepelev:
😀

PR #171, opened 2018-07-18
Input fix for Edge issue 169

### LINKED ISSUE #169: Comment's input field doesn't work in Edge.
I'm unable to type a comment in the input field using the Microsoft Edge browser, it's as if the input field is disabled.

Clearic:
Reactor Input component into a controlled component, this fixes #169 

And moved auto-size behavior into a separate component to reduce the complexity of Input component.


Guria:
Awesome, Thanks

Issue #74, opened 2018-06-06
Edit timeout countdown bug on frontend


Modi34:
![Alt Text](https://i.imgur.com/ZF72PKW.png)
it happens when system time is wrong.

on my windows 10 install I have a windows bug  - time is always a few hours behind for some reason.
If user time is different from server time this bug occure.
to fix it all you need to do is check that timeout is less than 301

umputun:
I don't see how this <301 check will solve anything. There is no way frontend can accurately adjust local time without syncing with the server in some way.

@igoradamenko this is not a bug but sort of unexpected thingy due to time mismatch. I'm not sure we should do anything about it but if we do - the most straightforward way is to use `Date` header and get that timestamp instead of the local time for the countdown on your side.

igoradamenko:
@umputun, I think you're right about the solution. We will try it.

jackburg:
I'll try to resolve this issue.

jackburg:
To resolve this bug I've calculated time span between server and a client and save it in every comment for future use — in the 'remaining' time for comment editing calculation. This bug appeared in both conditions, when user time is in the past and when it al little bit ahead of the server time. Now everything is working fine. @Modi34 could you please confirm?

Modi34:
@jackburg it works - great job

PR #174, opened 2018-07-21
Fix incorrect countdown timer calculation #74

jackburg:
This PR intended to fix (#74) issue.

Now when we download comments from the server, we calculating time span between server and a client and use it later when calculating time remaining for comment editing.


Guria:
@jackburg I think there is no need to store `timeDiff` in every comment.
Let's just put it in redux store.

jackburg:
@Guria ok, I've placed time diff into the store. Check it out.

jackburg:
@umputun looks like some mongo tests are not passed as expected, could you please take a look?

umputun:
This is odd. According to travis logs, it can't see `MONGO_TEST` and this is a relatively fresh change of master. However, it looks like you have merged the current master to your PR, so I don't really see what exactly confusing travis here. Can be something security related as they try to prevent malicious changes in CI file, but I don't see anything suspicious in your PR.

umputun:
I can suggest to re-do your PR on top of the current master. hopefully, it will make travis happy

jackburg:
Thanks, I'll try it.

jackburg:
Out of luck. Maybe I should undo the merge as well? @Guria any advice?

Guria:
Try to squash all commits into one

jackburg:
@Guria Looks like I'm unable to squash them, because they've already been pushed.

Guria:
You can use --force flag with push command

jackburg:
@guria how many commits should I squash? Only mine, or including commit with the merge (with umputun changes)? What do you think?

Guria:
Squash everything you see on Commits tab.
Next time to update your feature branch prefer rebase over merge from master.
Don't afraid to force push into feature branches.

jackburg:
Got it, thank you.

jackburg:
@Guria I've squashed all commits (btw I've many merge conflicts) and pushed the result with force flag, but still has this error, do you have any more advices?

Issue #156, opened 2018-07-11
Add wordpress comment import.


cooperaj:
I'm looking to move my existing comment from wordpress to discuss. I *could* export from wordpress to discuss and then from discuss to remark but it'd be nice to cut out the middle man :)

umputun:
Do you have WP spec for their export format? Also, some real example of exported comments will help as well

akosourov:
It looks like Wordpress can export comments in different formats as XML or CSV. Suppose the db structure is [here](https://codex.wordpress.org/Database_Description#Table:_wp_comments). 
If we find real examples of exported comments I can try to take this one, looks like similar to `Disqus.Import`

umputun:
yeah, should be doable and not too hard. In addition to the new Importer implementation, you will need some minor change in `rest.Migrator.importCtrl`

akosourov:
@cooperaj Could you give an example of WP exported comments?

cooperaj:
Oh sure. Do you have somewhere I can put them? Or, at least somewhere I can
send a link to them? A comment export is essentially a full export of your
wordpress instance and I don't really want that up on github :)

akosourov:
@cooperaj I think you can send email with comments to @umputun and to me (because I'm going to solve it). You can find email addresses on github user's page

cooperaj:
@akosourov I've emailed an export of my blog to both your addresses :)

akosourov:
@cooperaj yeah, see thanks
@umputun May I implement this ticket?

akosourov:
@cooperaj due to #182 merged remark has WP importer. Try to import all comments to remark and  leave feed back, please. Also check the [instruction](https://github.com/umputun/remark#initial-import-from-wordpress) about importing from WP, and if it is not so obvious please open PR to update it or simple say what is wrong

cooperaj:
👍 LGTM

With the caveat that I used the same export that you developed against so may not get to see any hidden issues.

[Some imported comments visible here](https://realmenweardress.es/2016/07/deploying-rancheros-on-vultr-instances/)

cooperaj:
Though as you can see on the last comment on that page the markdown processing hasn't quite worked.

umputun:
> The export file contains newlines just fine so it looks like the import is what removed them.

@akosourov you do it here https://github.com/umputun/remark/blob/master/backend/app/migrator/wordpress.go#L151

[...]

PR #182, opened 2018-08-02
Add wordpress importer

akosourov:
Added WP importer for #156.

### DIFF
--- a/backend/app/migrator/disqus.go
+++ b/backend/app/migrator/disqus.go
@@ -73,7 +73,7 @@ func (d *Disqus) Import(r io.Reader, siteID string) (size int, err error) {
 
 	log.Printf("[DEBUG] imported %d comments to site %s", passed, siteID)
 
-	if failed > 0 && passed == 0 {
+	if passed == 0 {
 		err = errors.New("import failed")
 	}
 	return passed, err
@@ -133,7 +133,7 @@ func (d *Disqus) convert(r io.Reader, siteID string) (ch chan store.Comment) {
 							Name: comment.AuthorName,
 							IP:   comment.IP,
 						},
-						Text:      d.cleanText(comment.Message),
+						Text:      cleanText(comment.Message),
 						Timestamp: comment.CreatedAt,
 						ParentID:  comment.Pid.Val,
 					}
@@ -158,7 +158,7 @@ func (d *Disqus) convert(r io.Reader, siteID string) (ch chan store.Comment) {
 	return commentsCh
 }
 
-func (d *Disqus) cleanText(text string) string {
+func cleanText(text string) string {
 	text = strings.Replace(text, "\n", "", -1)
 	text = strings.Replace(text, "\t", "", -1)
 	return text
--- a/backend/app/migrator/disqus.go
+++ b/backend/app/migrator/disqus.go
@@ -68,14 +68,14 @@ func (d *Disqus) Import(r io.Reader, siteID string) (size int, err error) {
 	}
 
 	if failed > 0 {
-		return passed, errors.Errorf("failed to save %d comments", failed)
+		err = errors.Errorf("failed to save %d comments", failed)
+		if passed == 0 {
+			err = errors.New("import failed")
+		}
 	}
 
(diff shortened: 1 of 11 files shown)


backend/app/migrator/migrator_test.go
```diff
@@ -19,7 +19,7 @@ func TestMigrator_ImportDisqus(t *testing.T) {
 		os.Remove("/tmp/disqus-test.xml")
 	}()
 
-	err := ioutil.WriteFile("/tmp/disqus-test.xml", []byte(xmlTest), 0600)
+	err := ioutil.WriteFile("/tmp/disqus-test.xml", []byte(xmlTestDisqus), 0600)
```
golangci: G303: File creation in shared tmp directory without using ioutil.Tempfile

backend/app/migrator/migrator_test.go
```diff
@@ -39,6 +39,32 @@ func TestMigrator_ImportDisqus(t *testing.T) {
 	assert.Equal(t, 3, len(last), "3 comments imported")
 }
 
+func TestMigrator_ImportWordPress(t *testing.T) {
+	defer func() {
+		os.Remove("/tmp/remark-test.db")
+		os.Remove("/tmp/wordpress-test.xml")
+	}()
+
+	err := ioutil.WriteFile("/tmp/wordpress-test.xml", []byte(xmlTestWP), 0600)
```
golangci: G303: File creation in shared tmp directory without using ioutil.Tempfile

backend/app/migrator/disqus.go
```diff
@@ -73,7 +73,7 @@ func (d *Disqus) Import(r io.Reader, siteID string) (size int, err error) {
 
 	log.Printf("[DEBUG] imported %d comments to site %s", passed, siteID)
 
-	if failed > 0 && passed == 0 {
+	if passed == 0 {
```
umputun: any particular reason why 0 imported should be an error? In addition, this code was forgiving a little bit by design to report as failed, only if nothing were imported.

backend/app/migrator/wordpress.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	if failed > 0 {
+		return passed, errors.Errorf("failed to save %d comments", failed)
+	}
+
+	log.Printf("[DEBUG] imported %d comments to site %s", passed, siteID)
+
+	if passed == 0 {
+		err = errors.New("import failed")
```
umputun: same as above. I don't see why nothing imported should be an error.

umputun:
Looks like some of the tests hanging. Not sure if related to your changes or something on travis side

umputun:
One more thing - you missed readme update. We need a new section similar to "Initial import from Disqus"

umputun:
well, looks like something in your PR causing this hanging test. I have tried to restart your build twice and it is hanging. At the same time, master build passed just fine.

Tried to run your tests locally and it worked fine, but something here driving travis crazy.

umputun:
another possibility - some test running before yours kept things in bad shape. I have found `migrator.remark_test.go" doesn't remove bolt files, fixed in master. You may try to merge to yours and see if helped

backend/app/migrator/disqus.go
akosourov: It looks like `if failed > 0 && passed == 0` never be executed until `if failed > 0 return ...` exists upper. Maybe swap this conditions? I mean first put `if failed > 0 && passed == 0` and then put `if failed > 0 return ...`
So when `failed` exists it should return error but differentiate cases when nothing were converted and parts were converted.

My verstion was if nothing were converted it should return error regardless of `failed` exists or nor. Yeah, I agree it is incorrect

akosourov:
> One more thing - you missed readme update. We need a new section similar to "Initial import from Disqus"

Of course, I'll add it at this evenening/night

> You may try to merge to yours and see if helped

Ok, I'll try. thanks

akosourov:
@umputun
1. I've added WP section in readme file. But I haven't got a WP site, so this instruction can be not so strict. So I'll ask the author of ticket to correct instruction and leave feed back about WP importer.
2. I've changed error return at disqus.go and wordpress.go.
Please review

PR #188, opened 2018-08-12
Use comment formater for wp importer

akosourov:
for #156 
WP importer starts to use CommentFormatter to convert content in common way as remark does when saves comments.

### DIFF
--- a/backend/app/migrator/disqus.go
+++ b/backend/app/migrator/disqus.go
@@ -133,7 +133,7 @@ func (d *Disqus) convert(r io.Reader, siteID string) (ch chan store.Comment) {
 							Name: comment.AuthorName,
 							IP:   comment.IP,
 						},
-						Text:      cleanText(comment.Message),
+						Text:      d.cleanText(comment.Message),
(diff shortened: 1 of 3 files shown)


akosourov:
Maybe because 666 travis build number :)

umputun:
`migrator/wordpress.go:103:21: undefined: store.NewCommentFormater`

akosourov:
I'm not sure about my go code, is it go-way or not:
1. In order to satisfy store.CommentConverter interface I've created an empty struct with necessary method
2. Next one is about order in entity definition. I mean a define type after function definition. So i'm not sure is it a normal style or I should put type and after all type put func.
```
type ...
type...
func ...
type ...
func
```
If anyone can say is it correct or not I will be glad. Thanks a lot!

umputun:
@akosourov fyi - I have simplified your code a little bit. No real need to make an empty struct just to satisfy `CommentConverter` interface. I have added `Convert` func to `WordPress` itself.

akosourov:
@umputun Thanks for showing me a better solution!

PR #192, opened 2018-08-14
Use comment formatter for wp importer

akosourov:
for #156 
WP importer starts to use CommentFormatter to convert content in common way as remark does when saves comments.
Fix previous PR #188 (I did git rebase master && git push --force but nothing happen in PR)


akosourov:
I've returned back `cleanText` to Disqus struct. I suppose disqus Importer should use CommentFormatter too

See little mistake @umputun. You've created fom**R**ater.go

Issue #187, opened 2018-08-11
Focusable elements outline looks odd


Guria:
We need a consistent and well-looking focus indicator for controls.
Default browser outline is too odd for most mouse / touch users.

Clearic:
I agree, focus-ring on control elements looks weird. In my projects to solve this problem, I use [focus-visible polyfill](https://github.com/WICG/focus-visible). What it does is it hides outline when you interact with controls using a mouse, but outline still appears when using a keyboard. Thus we get best of both worlds, mouse users don't see the focus-ring and the site remains accessible by keyboard.

What do you think about this solution?

Guria:
Yeah, we can try it. Does it play nice with React / Preact?  Anyway, PR is welcome

Clearic:
It works absolutely fine with React. Tomorrow I'll prepare PR.

umputun:
@Guria - is this one done? LGTM

umputun:
I'm not sure what are "elements" here. The only one I see is "X" on last comments, and this one looks good. One minor inconsistency - it can't be closed by Esc, but if this is a big deal we can postpone this part

Clearic:
> One minor inconsistency - it can't be closed by Esc

The UserInfo can't be closed by Esc right now, because there is a bug. I'll fix it.

igoradamenko:
It works right now, thanks @Clearic.

Issue #16, opened 2018-05-13
CSRF in API method "comment"


dsdolzhenko:
Exploit example: https://jsfiddle.net/j37bvLzt/

```
<body onload='document.forms[0].submit()'>
  <form method='POST' enctype='text/plain' action="https://demo.remark42.com/api/v1/comment">
    <input name='{"text":"CSRF!","locator":{"site":"remark","url":"https://remark42.com/demo/"}}' value=''>
  </form>
</body>
```

umputun:
Thx a lot. Added protection for all calls with auth as well as for any "semi-open" calls like `/find`

PR #193, opened 2018-08-14
Add focus-visible polyfill, issue #187

Clearic:
1. Added focus-visible polyfill
2. Removed unnecessary `outline: none` from styles
3. Made buttons Pin, Block, Delete visible under focus

### DIFF
--- a/web/app/common/polyfills.js
+++ b/web/app/common/polyfills.js
@@ -1,4 +1,5 @@
 import 'core-js/es7/promise';
+import 'focus-visible';
 
 export default function loadPolyfills() {
   const fillCoreJs = () => {
--- a/web/app/components/comment/__action/comment__action.scss
+++ b/web/app/components/comment/__action/comment__action.scss
@@ -9,10 +9,6 @@
     color: #31c7c5;
   }
 
-  &:focus {
-    outline: none;
-  }
-
   + .comment__action {
     margin-left: 8px;
   }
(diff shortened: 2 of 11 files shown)


Clearic:
@Guria I've fixed the shape of the outline.

This close button appears with an outline because it is outside of the iframe. I think we should not apply focus-visible polyfill outside of the iframe. I'd left it as it is right now.

matrixik:
Sorry for possibly stupid question (I don't code in JS or CSS):
Why adding another JS script dependency for this functionality instead of some small CSS fixes?
Like in this user style: https://userstyles.org/styles/12981/highlight-every-focused-element-highlightfocus

Clearic:
@matrixik we are not simply changing the style of focused elements here, we hide the outline of focused elements for mouse and touch users, but outline still should appear for keyboard users. Watch this video [Focus Ring! -- A11ycasts #16](https://www.youtube.com/watch?v=ilj2P5-5CjI&t=392s), it very well describes the problem we are solving here.

Guria:
@umputun suggested to completely remove close button since you can click anywhere to hide overlay and listen to ESC-keypress event to close by keyboard. But close button must stay when overlay iframe completely covers the screen. @Clearic, what do you think about it?

Clearic:
Making the close button appears only when popup completely covers the screen seems a bit complicated.

Let's just not focus the close button by default and instead focus the iframe itself, this way the focus ring just won't appear around the close button every time a user opens the user info.

Clearic:
1) I've added the capability to close the User Info panel by hitting ESC. 
2) And the close button is not focused by default now.

PR #203, opened 2018-09-14
backend/add/store: remove commented-out code

Quasilyte:


umputun:
Such automated audit without human involvment is not a thing I'd like to see here. Pls remove our system from your tests or whatever you do.

Quasilyte:
It's not automated audit but rather a linter validation combined with an attempt of fixing something it can find. The only place I can remove remark is a list of projects that were checked during that session.

umputun:
Well, I can be wrong but it was my conclusion from two PR your submitted based on this observation:

- in both cases, "issues" won't be an issue if a human would be behind those PRs. At least a human who read the codebase and understood it a little bit.
- "commented-out code" suggested for removal in 2 of 3 cases wasn't what this linter thinks, but deliberate explanation indicating a part of the testing logic, i.e. how those "magic numbers" created. Probably PR replacing those hashes by `EncodeID` calls will be something to consider, but this is not what your PR suggested to do.
- Another PR with inversed conditions and early exists wasn't something to write home about either. I guess it was caused by a linter detecting cyclomatic complexity with some low threshold (maybe 10?) but proposed fix actually reduced readability.

_pls note: I don't mind style-related PRs and, in fact, CI runs a mandatory [set of linters](https://github.com/umputun/remark/blob/master/Dockerfile#L39) plus basic checks by [goreportcard](https://goreportcard.com/report/github.com/umputun/remark) to enforce it. However, I don't like fixes out of the context based solely on linter's report._

Quasilyte:
@umputun to keep things short: you're right and thanks for the response.

On the bright side, linter found no real issues (ones that are not style-related).

PR #204, opened 2018-09-14
backend/app/migrator: simplify single case type switch

Quasilyte:
Rewrite single case type switch as "v, ok := t.(type)"
statement. Invert body execution condition with early
check with continue, reducing the code nesting by 1 level.


umputun:
I don't think this change makes the code any simpler or more readable.

Issue #170, opened 2018-07-18
Add auth handling to deleteme.html page


Guria:
#109 introduced new page deleteme.html that finishes approval of `deleteme` request by admin.
The problem is that we assume that admin is already logged in when clicking the link from an email.
We need to show current user login there, controls to signin / signout and a button to actually send deleteme request.

umputun:
I don't think we even need to show the current user login. The only function we have to enforce - if the user is not admin (or not logged in) ask him to log in as admin.

Guria:
resolved by #208

PR #208, opened 2018-10-19
Auth handling added to deleteme.html #170

ur300:
getUser api call added to check whether user is admin or not


umputun:
due to some unusual slowness on Travis end, one of the tests failed. I have increased timeout and it should be fine now. pls remerge

PR #209, opened 2018-10-20
Add https support, static mode for rest server

### LINKED ISSUE #191: Integrate LE certificate support
The goal is to allow HTTPS with automatic Let's Encrypt certificate without a need of an external proxy. In addition, we should accept existing certificates.

akosourov:
It partially closes #191 issue (I suppose may be a candidate)
I've added support for `static` ssl mode. With `--ssl.mode=static` remark runs https server (and http server) and makes http to https redirect with 307 status code.
Also I've added my self-signed certificates for testing.
And I'm not sure whether it should  touch compose file or not for certs and https port.
I will be glad any feed back :)


akosourov:
@umputun Hi, could you give a feed back about this pr? I want to show you does I correctly understand what to do. Currently I'm preparing server to start testing autocert ssl (`auto`) solution. If this approach is ok, I'll add autocert solution (probably next weekend) with this changes.

umputun:
hi. I had no opportunity to review this one in details, just took a quick look. Planned to deal with this today, however, I'm not sure what the status of this PR anymore as you just closed it.

akosourov:
Thank you if you take a look :) I just closed it because it has a little conflict and it closes only part of this issue - adds a static https support without autocert mode. 
But this pr shows a main idea of changes, and I'm not sure, is it ok or not. For example is it ok to terminate servers consequently or should do it in parallel. Next one is it ok or not to expose  config struct from rest package to configure rest struct. 
The main reason was to show you and get a feed back :)

akosourov:
I can't resubmit because I've done rebase. So I've opened new pr #213

PR #213, opened 2018-10-28
https support static mode

akosourov:
add https support #191
previous pr #209


backend/app/cmd/server.go
```diff
@@ -129,6 +130,14 @@ type NotifyGroup struct {
 	} `group:"telegram" namespace:"telegram" env-namespace:"TELEGRAM"`
 }
 
+// SSLGroup defines options group for server ssl params
+type SSLGroup struct {
+	Mode string `long:"mode" env:"MODE" description:"ssl (auto)support" choice:"none" choice:"static" choice:"auto" default:"none"`
```
umputun: probably better to call this `Type` as in other groups

backend/app/cmd/server.go
```diff
@@ -143,7 +152,7 @@ type serverApp struct {
 
 // Execute is the entry point for "server" command, called by flag parser
 func (s *ServerCommand) Execute(args []string) error {
-	log.Printf("[INFO] start server on port %d", s.Port)
+	log.Printf("[INFO] start rest server in '%s' mode", s.SSL.Mode)
```
umputun: not sure how this will work with "none" mode. In this case the port should has nothing to do with SSL

backend/app/rest/api/rest.go
```diff
@@ -31,6 +31,28 @@ import (
 	"github.com/umputun/remark/backend/app/store/service"
 )
 
+// sslMode defines rest server mode (http or https)
```
umputun: probably moving all this chunk to a separate file (ssl.go, for example) will make some sense. At the same time functions related to ssl, like make HttpsServer amd httpToHTTPSRouter can go to the same file.

umputun:
Generally I don't see anything wrong with this way, just a few minor suggestions.

backend/app/cmd/server.go
akosourov: I've reverted it. I don't know what to log here in static or auto mode. In rest.Run also logs on which port server is running

backend/app/rest/api/rest.go
akosourov: Ok, done )

akosourov:
@umputun thank you for review. I've fixed it

akosourov:
I've added changes in readme.md and drone has failed. Don't understand what's wrong

umputun:
```
+ git init
Initialized empty Git repository in /go/src/github.com/umputun/remark/.git/
+ git remote add origin https://github.com/umputun/remark.git
+ git fetch --no-tags origin +refs/pull/213/merge:
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
exit status 128
```

My guess  - some networking glitch between DO and GitHub. Restarted the build and it passed just fine this time

Issue #200, opened 2018-09-06
As a user I would like to remove my own comments


m1ndful:
I haven't found a way to remove my own comments as a non-admin user.
Is this something you are planning to add?

umputun:
Currently, the only way to "remove" comment is to edit with empty (or "del") text. Probably I can add a new call for this with the same time restrictions as edit functionality. 

However, as far as I know, such removal functionality is not that usual for commenting systems, and I'm not sure if such functionality really needed.

umputun:
I have added `delete` flag to the current request (json) used by `PUT /api/v1/comment/{id}`

Now we need frontend support for this

m1ndful:
> Currently, the only way to "remove" comment is to edit with empty (or "del") text.
It doesn't really work, because once you hit on your username, it will show you history of changes.

umputun:
not sure what you mean, this is part of the current edit functionality, nothing special about this particular edit, just different text. In 5 minutes window user can edit a comment by clicking on "Edit" link, not the username.

umputun:
in case you meant a way to remove all comments, this is in place already, i.e. user can request complete removal of all comments by clicking on "Request my data removal" 

<img width="399" alt="fvd70_20180907_144240 yh5m8" src="https://user-images.githubusercontent.com/535880/45239862-58478000-b2ac-11e8-94a6-ab67508724cb.png">

m1ndful:
@umputun , looks like latest docker image is not in sync with master branch.

umputun:
if by latest you mean label `:latest` - sure, it is not from master but from tagged release. The one from master is `:master`, see https://hub.docker.com/r/umputun/remark42/tags/

PR #214, opened 2018-10-30
frontend: delete own comment button added

Reeywhaar:
As a continuation of #200 
This pull request adds delete user's own comment button while in edit time range.

screen shot 2018-10-30 at 23 16 51

Behaviour similar to admin's delete comment button.


igoradamenko:
@umputun, it's OK for me.

PR #217, opened 2018-11-05
Feature/add autocert support

akosourov:
I've added LE support with acme library. #191 issue

To test it, I've got a simple domain name and ran remark container with `auto` ssl mode. 
It is here https://akosourov.pw/ping. That works :)

But don't know how to test it locally.


backend/app/cmd/server.go
```diff
@@ -133,10 +133,11 @@ type NotifyGroup struct {
...
 type SSLGroup struct {
-	Type string `long:"type" env:"TYPE" description:"ssl (auto)support" choice:"none" choice:"static" choice:"auto" default:"none"`
-	Port int    `long:"port" env:"PORT" description:"port number for https server" default:"8443"`
-	Cert string `long:"cert" env:"CERT" description:"path to cert.pem file"`
-	Key  string `long:"key" env:"KEY" description:"path to key.pem file"`
+	Type  string `long:"type" env:"TYPE" description:"ssl (auto)support" choice:"none" choice:"static" choice:"auto" default:"none"`
+	Port  int    `long:"port" env:"PORT" description:"port number for https server" default:"8443"`
+	Cert  string `long:"cert" env:"CERT" description:"path to cert.pem file"`
+	Key   string `long:"key" env:"KEY" description:"path to key.pem file"`
+	Cache string `long:"cache" env:"CACHE" description:"dir where certificates will be stored by autocert manager" default:"./var/certs"`
```
umputun: parameter named "cache" can be confusing, as it conflict with the current cache (means smth else). It makes sense to name it "location", but even better - get rid of it completely and just use existing Cert and Key

umputun:
It is not clear how to make a test for such thing. Generally speaking we don't want to thest third-party, but our part of the logic so probably the only thing you can do here is to test httpChallengeRouter

backend/app/cmd/server.go
akosourov: I agree, but acme library expects path to directory where cert and key are stored, not paths to cert and key files. So `static` mode expects Cert and Key params, but `auto` mode expect only one directory param. So don't understand how to union them in common maybe rename it to "location"?

umputun: two possible solutions:
- get the directory from cert or key
- just name this param "acme_location"

Probably the second less confusing

akosourov:
Ok, but it not clear for me what do we want to test. I mean this router can perform two things: acme http challenge or redirect. Do we need tests for acme part? Or only redirect part

umputun:
I'd say - both.

akosourov:
Ok, I'll add

backend/app/rest/api/ssl.go
```diff
@@ -25,25 +28,71 @@ const (
...
 	})
-	return router
+}
+
+func (s *Rest) makeAutocertManager() *autocert.Manager {
+	return &autocert.Manager{
+		Prompt:     autocert.AcceptTOS,
+		Cache:      autocert.DirCache(s.SSLConfig.ACMELocation),
+		HostPolicy: autocert.HostWhitelist(s.getRemarkHost()),
+		Email:      s.DataService.AdminStore.Email(""),
```
umputun: this can be a problem for any other store. I.e. for the default (static store) empty siteID will work, but for mongo store it won't

umputun: I don't think you can get a proper admin's email in this case, because it runs on the server level and in fact, this may not be the same admin email as fo sites. Pls note - the goal of admin email(s) is to send "delete me" requests, and those can be routed differently by different sites as it happens on rest level. For acme I would suggest a minor change:

1. Add a new SSL param, smth like `acme-email` to options
1. Make it empty by default
1. If defined in CLI/env just use it
1. if not set to:
   - admin email if `AdminGroup.Type == "shared"`
   - to admin@+`RemarkURL` domain

umputun:
one more thing - I'm not sure if the default tls.Config sane, i.e. ready to be exposed. Well, maybe it is, but you can test with  https://www.ssllabs.com/ssltest/ to make sure.

Probably something like this can be helpful if defaults not setting it:

```go
    tls.Config{
		PreferServerCipherSuites: true,
		CurvePreferences: []tls.CurveID{
			tls.CurveP256,
			tls.X25519,
			tls.CurveP384,
		},
		MinVersion: tls.VersionTLS12,
		CipherSuites: []uint16{
			tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
			tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
			tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
			tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
		},
		Certificates: certs,
    }
```

akosourov:
Ok, I'll check and restart server after changes and run this analyze tool to see the difference. Thanks!

backend/app/rest/api/ssl.go
akosourov: fixed [here](https://github.com/umputun/remark/pull/217/commits/2c5a2921eefe5ac86be5ee52a8ab960de8272b37)

akosourov:
I've fixed getting admin email and changed tls.Config.
After setting custom tls.Config the ssl report stay more secure (I suppose). It is [here](https://www.ssllabs.com/ssltest/analyze.html?d=akosourov.pw). But I've added 
one extra cipher suite TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA to support Safari 8. Because without it the report `Handshake Simulation` was red (see pic). If it is not necessary I delete it. 

Travis-CI failed, but I suppose not due to my changes

image

umputun:
LGTM, thx

Issue #218, opened 2018-11-12
The operation is insecure


IRus:
remark installed on domain comments.foo.com
and I try to use it on domain blog.bar.com

```
Request to access cookie or storage on “https://comments.foo.com/web/iframe.html?site_id=some_id&url=https%3A%2F%2Fblog.bar.com%2F2017%2F06%2F08%2Fhow-to-use-apache-asynchttpclient-with-kotlin-coroutines%2F” was blocked because we are blocking all third-party storage access requests and content blocking is enabled. embed.js:1:1672
Request to access cookie or storage on “https://comments.foo.com/web/iframe.html?site_id=some_id&url=https%3A%2F%2Fblog.bar.com%2F2017%2F06%2F08%2Fhow-to-use-apache-asynchttpclient-with-kotlin-coroutines%2F” was blocked because we are blocking all third-party storage access requests and content blocking is enabled. remark.js:formatted:4259
SecurityError: The operation is insecure. remark.js:formatted:1285
```

umputun:
Is it latest safari with "prevent cross-site tracking"? If so, this is the known issue due to the iframe usage. I think all other systems with similar logic have the same problem, see for example [this discussion](https://discussions.apple.com/thread/8544741). Btw, all of this seems to work fine if domains related, i.e. www.example.com and remark42.example.com.

@igoradamenko is trying to address it. If anyone has bright ideas on how to handle it - pls share.

IRus:
Firefox with blocking third-party cookies:

![image](https://user-images.githubusercontent.com/493656/48363807-d9a9fd00-e6b7-11e8-9b25-99d8b86f737a.png)

Disqus works fine in same setup.

IRus:
> Btw, all of this seems to work fine if domains related, i.e. www.example.com and remark42.example.com

yes, for same domain it works, i.e. `comments.bar.com` can be used in `bar.com`, but not in `foo.com`.

IRus:
@igoradamenko why do you need to use iframe?

igoradamenko:
@IRus, because of security reasons; lots of them.

Could you imagine third-party widget on any site without an iframe?

IRus:
Why you treat self-hosted solution as third-party? You already load some JS, so what the problem to do all job within this file? What vectors of attack you considered?

igoradamenko:
@IRus, that's the open question, actually. 

[...]

PR #219, opened 2018-11-14
third party cookies blocked fallback

Reeywhaar:
Related to #218 

When using Firefox with "block third-party cookies: all" setting, the comments iframe fails with "the operation is insecure" error message.
Crash caused by call of any method of localStorage.
Current pr adds check if storage available and if not shows placeholder below instead of "Sign in with..."

screen shot 2018-11-14 at 06 51 07
By clicking on a "new page" user relocated to new page with the same origin as an iframe (template is  `/web/comments.ejs`)
If user blocked cookies completely placeholder will be the following

screen shot 2018-11-14 at 07 04 08

Current solution was tested mainly with Firefox, Safari seems to allow cookies even with "Prevent cross-site tracking" turned on.

### DIFF
--- a/web/app/common/constants.js
+++ b/web/app/common/constants.js
@@ -59,7 +59,7 @@ export const STORAGE_AVAILABLE = (() => {
  */
 export const IS_THIRD_PARTY = (() => {
   try {
-    return window.parent.document.location.host !== window.location.host;
+    return window.parent.location.host !== window.location.host;
   } catch (e) {
     return true;
   }
--- a/web/app/common/constants.js
+++ b/web/app/common/constants.js
@@ -38,3 +38,29 @@ export const BLOCKING_DURATIONS = [
     value: `${60 * 24}m`,
   },
 ];
+
+/**
+ * Defines if browser storage features (cookies, localsrotage)
+ * are available or blocked via browser preferences
+ */
+export const STORAGE_AVAILABLE = (() => {
+  try {
+    localStorage.setItem('localstorage_availability_test', null);
+    localStorage.removeItem('localstorage_availability_test');
+  } catch (e) {
+    return false;
+  }
+  return true;
+})();
+
+/**
+ * Defines whether iframe loaded in cross origin environment
+ * Usefull for checking if some privacy restriction may be applied
+ */
+export const IS_THIRD_PARTY = (() => {
+  try {
+    return window.parent.document.location.host !== window.location.host;
+  } catch (e) {
+    return true;
+  }
+})();
--- a/web/app/common/constants.js
+++ b/web/app/common/constants.js
@@ -43,7 +43,7 @@ export const BLOCKING_DURATIONS = [
  * Defines if browser storage features (cookies, localsrotage)
  * are available or blocked via browser preferences
  */
-export const STORAGE_AVAILABLE = (() => {
+export const IS_STORAGE_AVAILABLE = (() => {
   try {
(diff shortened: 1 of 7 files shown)


web/app/common/constants.js
```diff
@@ -38,3 +38,29 @@ export const BLOCKING_DURATIONS = [
     value: `${60 * 24}m`,
   },
 ];
+
+/**
+ * Defines if browser storage features (cookies, localsrotage)
+ * are available or blocked via browser preferences
+ */
+export const STORAGE_AVAILABLE = (() => {
```
igoradamenko: `IS_STORAGE_AVAILABLE`, m?

web/app/common/localStorage.js
```diff
@@ -0,0 +1,30 @@
+export const isAvailable = (() => {
```
igoradamenko: Why do you need check here, if you have it in constants.js?

web/app/components/thread/getCollapsedComments.js
```diff
@@ -1,5 +1,6 @@
 import { LS_COLLAPSE_KEY } from 'common/constants';
+import { getItem } from 'common/localStorage';
 
-const getCollapsedComments = () => JSON.parse(localStorage.getItem(LS_COLLAPSE_KEY) || '[]');
+const getCollapsedComments = () => JSON.parse(getItem(LS_COLLAPSE_KEY) || '[]');
```
igoradamenko: It would be better if you imported 'common/localStorage' as `LS`, because in this case it would be `LS.getItem`, and it's more readable than just odd `getItem`.

web/comments.ejs
```diff
@@ -0,0 +1,58 @@
...
+      }, {});
+    })()
+
+    if (query.site_id && query.url) {
+	  var titleElement = document.getElementById("title");
+	  titleElement.innerHTML = '<h1>Comments for <a href="'+query.url+'">'+query.url+'</a></h1>'
+
+      var remark_config = {
+        site_id: 'remark',
+        url: 'https://remark42.com/demo/',
```
igoradamenko: Are you sure about this url? Don't we need to change it?

Reeywhaar: oh my

Reeywhaar:
@igoradamenko upon changes I noticed that even though jest tests pass, `app/components/thread/thread.reducers.test.js` spawns a console.error("localStorage access denied, check browser preferences") which defined in `common/localStorage.js`, and I wonder is it fine from UX perspective, is this error message legit?

web/app/components/thread/getCollapsedComments.js
Reeywhaar: I decided to rather import as `import { getItem as localStorageGetItem } from 'common/localStorage';` as importing * is considered bad practice.

web/app/common/localStorage.js
Reeywhaar: I thought of constant in "common/constants" as it defines global state of storage including cookies, localStorage, indexeddb and so on, it just happens that it the same as a constant in `common/localStorage`, but could be changed to more sophisticated version in future. I've changed for now, because maybe I'm over-complicating it.

igoradamenko:
@igoradamenko, I think it's fine.

igoradamenko:
@Reeywhaar, thanks :-)

Reeywhaar:
@igoradamenko always welcome!

Issue #220, opened 2018-11-18
embed.js doesn't watch for DOM changes


Reeywhaar:
Problem:
Current way of embedding remark requires that element with `#remark42` id should be present on page before loading embed.js. This doesn't work well will SPA (Single page apps) which often use no-page-reload navigation via history api. Therefore remark element could appear on page at any time, not only on initial load.

Solution:
Use `mutationObserver` to track DOM changes or fallback to initial method.

----
One another aspect that remark requires id, which limits remark's user (the one who deploy it on its server) to have at maximum one element with remark's comments, or it will be against html spec to have multiple elements with same ids. I don't know if this intentional, but thing exists.

PR #221, opened 2018-11-21
Dynamic comment's node watching

Reeywhaar:
Related to #220 

This PR adds ability for Remark to watch for comment's node adding/removing dynamically via MutationObserver. This solves the the problem I've stumbled while working on https://github.com/radio-t/rtnews-ui . It uses History API navigation, so remark's node can be added to the DOM not only at initial load, but at any time. To solve this, I had to copy/paste [`embed.js` and tweak it a bit](https://github.com/radio-t/rtnews-ui/blob/0ff0ea3547cc90b4e12aa2c5f34c6527664243d9/src/remark.jsx#L11-L264). This is not only looking ugly as hell, but also very inconvenient cause if Remark's node handling will change, I should recopy/repaste and apply tweaks again and maintain initialization/deinitialization myself.

Also, PR adds ability to define custom selector or straight HTMLElement, not only `#remark42`. This is useful, if you for example want set id to "#comments", or track by class.

**Downside:** Solution uses MutationObserver which watch for any hierarchy change in document.body. This could be useless if website is not dynamic. So, I added boolean `dynamic` flag set to `false` by default.
Also, if browser doesn't support Mutation observer, the user will be left with it as is.

**Backwards compatibility:** compatible

------
I [used plain array](https://github.com/Reeywhaar/remark/blob/4d0bef74df1d79a06cc93613cd9d80f2a905af09/web/app/embed.js#L233) to store initialized nodes info. But `WeakMap` is more suitable for such case, but it's support though is very good, but is still questionable, depending on which platforms do you plan to support, like ie10, 11?

------
I noticed that same logic may be applied to `counter` and `last comments` pages. So, I wonder, do I have a point with this pr, guys? Or maybe another approach could take place. Like Initializing global `Remark` object with `initNode`, `deinitNode` methods?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ Remark42 is a self-hosted, lightweight, and simple (yet functional) comment engi
 
   - [Install](#install)
     - [Backend](#backend)
-      - [With Docker](#with-docker) 
-      - [Without docker](#without-docker) 
+      - [With Docker](#with-docker)
+      - [Without docker](#without-docker)
       - [Parameters](#parameters)
         - [Required parameters](#required-parameters)
       - [Register oauth2 providers](#register-oauth2-providers)
@@ -76,7 +76,7 @@ _this is the recommended way to run remark42_
 #### Without docker
 
 * download archive for [stable release](https://github.com/umputun/remark/releases) or [development version](https://remark42.com/downloads)
-* unpack with `gunzip` (Linux, macOS) or with `zip` (Windows) 
+* unpack with `gunzip` (Linux, macOS) or with `zip` (Windows)
 * run as `remark42.{os}-{arch} server {parameters...}`, i.e. `remark42.linux-amd64 server --secret=12345 --url=http://127.0.0.1:8080`
 * alternatively compile from the sources - `make OS=[linux|darwin|windows] ARCH=[amd64,386,arm64,arm32]`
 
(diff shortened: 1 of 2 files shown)


umputun:
I still don't understand what exactly was the problem and what "to watch for comment's node adding/removing dynamically via MutationObserver" supposed to fix. Probably just my ignorance and all of frontend guys get it, but anyway ;)

Reeywhaar:
Hehe, well, [History API](https://css-tricks.com/using-the-html5-history-api/) is a way for single page application to dynamically reuse current DOM without actual http navigation. Consider we visit [news.radio-t.com](http://news.radio-t.com) and click on some post [news.radio-t.com/post/some-post](http://news.radio-t.com/post/some-post). With history api server will get no `GET http://news.radio-t.com/post/some-post` request, since we didn't actually navigate from start page, we just reused current document, and replaced content dynamically with javascript.

So, the problem is that node `<div id="remark42">` appeared on the page after we declared remark [initialization script](https://github.com/umputun/remark#setup-on-your-website). Remark already failed with "no root found" error, and exited. That what I meant by "watch for comments node dynamically", remark should watch for node not only at start, but should track changes in DOM and apply init/deinit logic. That is what MutationObserver for. It's an API for watching changes in DOM, such as node adding/removing, change of node's attribute, etc...

I can modify pr to remove merge conflicts, but, hey @Guria, @igoradamenko what do you think about the whole situation? I said already that there could be multiple ways to implement api: first via `MutationObserver`, second manually via `window.REMARK.init(node)/window.REMARK.deinit(node)` methods.

Guria:
@Reeywhaar Why not to use History API events instead of MutationObserver?

Reeywhaar:
@Guria, changes in DOM not always caused by changes in history.

Reeywhaar:
Revoking this pr, but question in #221 still stands.

poohitan:
Why wasn't it merged?

Issue #207, opened 2018-10-02
Restrict max/min vote counts


umputun:
In order to prevent comment's vote section from growing too large, it makes sense to limit the maximum number of elements and ignore/rejects attempts to add another vote. This could be some configurable max with some sane default, like 1000 or so.

Reeywhaar:
Maybe rather apply transform to score to become like `1k`, `1.5k` `1m`? The `title` attribute could show precise number of votes. Then it's just a frontend issue.

umputun:
The issue is not how to show, but how to restrict the list of [`Votes`](https://github.com/umputun/remark/blob/master/backend/app/store/comment.go#L20) (technically speaking not list but map) in comments. Each key here is a unique user ID and the value indicates +/-.

PR #223, opened 2018-11-23
added maximum votes limit per comment

amelnikov-mylivn:
feel free to correct me if I did something not as expected


amelnikov-mylivn:
> Missing a change for README params

done

BTW do you think it is better to change unlimited to `0` as in other similar places? The only downside is that It will make impossible to disable voting by setting `0`.

umputun:
thx, LGTM

fix #207

umputun:
> change unlimited to 0 as in other similar places? 

to me default -1 makes sense as in this case 0 can be the real value.

PR #228, opened 2018-12-15
Dark theme

igoradamenko:
@umputun, I've created dark theme as we discussed before.

I've created this PR just as a notification, because I think that some folks want to use Remark in a “dark mode”.

Now it's possible to pass `theme` key to `remark_config` object. By default it is `light`, but also `dark` is available. If you have to, you can change theme in runtime using global function `window.REMARK42.changeTheme`. Read more about this in README.

/cc @Reeywhaar


igoradamenko:
@umputun, I think it's fine to merge this PR, and then test dark theme on news.radio-t.com and pirates.radio-t.com to collect feedback.

If you agree with me, merge it, please.

PR #233, opened 2018-12-18
simplify makeDirs()

zonescape:
I propose to simplify makeDirs() function.
(Note that current implementation of makeDirs() contains a bug on the line 124 - it does not wrap right error variable. It also does not check whether existed path is a directory.)


umputun:
LGTM. It failed some tests, but likely due to the bug on 124 as it returns the proper error now. I'll merge and adjust tests.

thx

Issue #230, opened 2018-12-16
Votes change enabled for read-only posts


umputun:
Not a big deal but feels wrong. I tend to believe it is not handled correctly on both backend and frontend sides. I'll take care of backend, but need someone to disable it on the frontend.

umputun:
this was done for the backend and any votes on RO posts now rejected. Need UI side to adopt.

Reeywhaar:
Anyone at it? May take.

umputun:
sure

Reeywhaar:
Noticed that server allows to vote on posts disabled manually. Is it intentional? Can I rely on `info.read_only` value?

umputun:
>  Is it intentional?

no, I  just missed this one.

> Can I rely on info.read_only value?

yeah. I think it already does the check to hide edit controls on RO post. The same thing should be done for arrows as well

PR #234, opened 2018-12-19
Disable voting on read-only posts

Reeywhaar:
Rel to #230 

PR disables comments voting on read-only posts.

screen shot 2018-12-19 at 10 47 18

@umputun I made it by just disabling arrows, did you mean to completely hide them?

### DIFF
--- a/web/app/components/comment/comment.jsx
+++ b/web/app/components/comment/comment.jsx
@@ -45,8 +45,7 @@ export default class Comment extends Component {
     this.isAdmin = this.isAdmin.bind(this);
     this.isCurrentUser = this.isCurrentUser.bind(this);
     this.isGuest = this.isGuest.bind(this);
-    this.getDownvoteDisabledReason = this.getDownvoteDisabledReason.bind(this);
-    this.getUpvoteDisabledReason = this.getUpvoteDisabledReason.bind(this);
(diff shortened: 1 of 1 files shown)


umputun:
I meant disabling, no need to hide them

web/app/components/comment/comment.jsx
```diff
@@ -335,10 +340,50 @@ export default class Comment extends Component {
...
+      (store.get('user') && store.get('user').id)
+    );
+  }
+
+  /**
+   * returns reason for disabled downvote
+   *
+   * @return {(string|null)}
+   */
+  getDownvoteDisabledReason() {
```
Guria: Is there a reason for two identical methods?

Reeywhaar: Aleksei, thanks for asking! I doubt too, but my intent was that disabling upvote and downvote reasons may differ, mainly for downvote, for example it you want to limit downvote or disable it at all, or want to disable downvoting admins, or mark some controversial comment as non-downvotable, etc..., so I tried to think ahead of time. Maybe there is no reason for this now, and I'm overcomplicating it.

Reeywhaar: Added commit, as an alternative.

Issue #241, opened 2019-01-06
Add support of post titles


umputun:
This is mostly for the last comments widget, but also usable for notifications as well. The goal is to keep title (page title) as a part of the comment and display them.

igoradamenko:
@umputun, looks like it works :-)

umputun:
kind of, sort of ;) We need a few more things:

- ability to reset (set) titles in some "batch migration"
- restore links on names if no title (back compatibility)
- remove hover on non-linked names

Reeywhaar:
Nice to have titles in user comments sidebar as well.

umputun:
@Reeywhaar feel free to PR ;)

PR #244, opened 2019-01-07
Show post title in user comments sidebar

Reeywhaar:
#241

Made post title block in user comments sidebar. Good? Title link and link with time are the same.

screen shot 2019-01-07 at 12 31 11


umputun:
I think we strip title in last comment widget ( with...), probably should do the same here

Reeywhaar:
@umputun, fixed

screen shot 2019-01-07 at 13 07 28

web/app/components/comment/__title/comment__title.scss
```diff
@@ -0,0 +1,17 @@
+.comment__title {
```
igoradamenko: This code affects preview mode which is used inside last comments widget, doesn't it?

igoradamenko: I would prefer to move this elem under comment_view_user modifier. Check comment_view_preview.

web/app/components/comment/__title/comment__title.scss
```diff
@@ -0,0 +1,17 @@
...
+  margin-bottom: 0.2rem;
+
+  &-link {
```
igoradamenko: And also, don't write code in this way, please. Use full selectors. Right know it's impossible to find `__title-link` using IDE search.

Reeywhaar: ok, I get it.

web/app/components/comment/__title/comment__title.scss
Reeywhaar: Oh, now I found preview mode (last comments) widget :-) Should I make the style the same as preview one?
Mine:
screen shot 2019-01-07 at 23 43 44
Same as preview mode:
screen shot 2019-01-07 at 23 44 35

Reeywhaar: @igoradamenko  check it out. Last commit can be reverted if top screenshot wins.

igoradamenko: The top screenshot looks better for me. Let's revert that two commits.

Reeywhaar: Done! Though title margin makes me doubt:

without:
screen shot 2019-01-09 at 16 53 46
with:
screen shot 2019-01-09 at 16 53 55

which is better?

igoradamenko: I don't know ¯\_(ツ)_/¯

Let's pick the second one.

Reeywhaar: Then all good :-)

Issue #248, opened 2019-01-15
Comments with replies should be non-editable


umputun:
Currently, the only limit for comment's edit is a time window. However, editing or deleting the comment after a reply to it looks very confusing. We should disable/rejects such edits on the backend side.

umputun:
I have added a check on my side and rejection. However, UI should treat such comment as read-only the same way it does based on the time window. Hopefully, it can be easily detected if any child comments/replies and hide edit and delete links.

@Reeywhaar - wanna take a look?

Reeywhaar:
@umputun will be glad to help, but not as early as tomorrow

umputun:
thx

Reeywhaar:
Am I understading right that if comment has reply from the same user it's still editable?

umputun:
this feels like a bug on my side. Probably I'll just switch any comment with any reply to RO

PR #250, opened 2019-01-16
Disable comment editing if it has reply

Reeywhaar:
Frontend side for #248

* If comment has reply then edit button is hidden.
* if user started to edit comment, before someone replied, and then clicks on `send`, the following message will be shown:
screenshot 2019-01-16 at 23 43 55
* if value of edit textarea is the same as original comment (user clicked edit, but haven't typed anything), then on `send` click edit input will be hidden (like if edit passed)


umputun:
> if user started to edit comment, before someone replied, and then clicks on send, the following message will be shown:

good idea

PR #257, opened 2019-01-18
Support restricted words

amelnikov-mylivn:


backend/app/store/service/service.go
```diff
@@ -24,6 +25,7 @@ type DataStore struct {
 	MaxCommentSize int
 	MaxVotes       int
 	TitleExtractor *TitleExtractor
+	Matcher        *Matcher
```
umputun: I'd suggest more descriptive name for both fields and strict. Maybe BadWordsMatcher. I little bit long, but easy to understand

backend/app/store/service/restricted_words.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+	"strings"
+	"unicode"
+)
+
+type Lister interface {
+	List(siteID string) (restricted []string, err error)
+}
+
+type StaticLister struct {
```
umputun: I don't think static lister should have map here. All other static implementations of similar things we have just ignore site and have a shared list

backend/app/store/service/restricted_words.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, token := range tokens {
+		_, present := data.restricted[token]
+		if present {
+			return true
+		}
+	}
+	return false
+}
+
+func (_ *Matcher) tokenize(text string) []string {
```
umputun: is it even needed? I think string pkg has smth to split with whitespaces

amelnikov-mylivn: This one also strips punctuation, spaces, tabs, emojis. See tests. For me it seems strange to have tokenizer that leaves such trash. It would be harder to match these tokens later.

backend/app/store/service/service.go
amelnikov-mylivn: Agree. Just got confused with examples from our discussion.

umputun:
the build failed, multiple linter issues

backend/app/store/service/restricted_words.go
umputun: ok, makes sense

umputun:
now the only missing thing - feature test on rest level. I.e. rest_private_test.go a test case adding restricted words to service and trying to post & update with one of those.

backend/app/rest/api/rest_private_test.go
```diff
@@ -258,6 +276,30 @@ func TestRest_UpdateNotOwner(t *testing.T) {
...
+	req, err := http.NewRequest(http.MethodPut, ts.URL+"/api/v1/comment/"+id+"?site=radio-t&url=https://radio-t.com/blah1",
+		strings.NewReader(`{"text":"What the duck is that?", "summary":"my edit"}`))
+	assert.Nil(t, err)
+	req.Header.Add("X-JWT", devToken)
+	b, err := client.Do(req)
+	assert.Nil(t, err)
+	body, err := ioutil.ReadAll(b.Body)
+	assert.Nil(t, err)
+	c := R.JSON{}
+	err = json.Unmarshal(body, &c)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

amelnikov-mylivn:
I faced a problem that by default all errors happened in `DataStore.Create` are exposed as HTTP 500 by controller, and all errors happened in `DataStore.EditComment` are exposed as HTTP 400 by controller. 
I moved restricted words check from these methods into `DataStore.ValidateComment` and `DataStore.ValidateCommentText` that are called by controllers and provide valid 400 errors in both cases.

umputun:
Confused about dbl-check for updateCommentCtrl. Why we even need the second check? I mean it checks once in service.EditComment and the second time in the rest controller

amelnikov-mylivn:
My bad. Forgot to remove it after moving switching to validate methods.

amelnikov-mylivn:
So first draft was based on exact match between tokens from comment and restricted words. The only benefit for this approach is speed (since we use map it is `O(n)` where n is number of tokens in comment).

Ideally what we want (probably) is to allow user to specify regexps as restricted words. But if we will do it in a naive way we will end up with `O(m*n*k)` complexity where m number of regexps, n number of tokens in comment, k average complexity of matching single regexp. 

So I implemented something in the middle - matching using wildcards. It allows us to use `*`, `?`, `+`, `!` (this one not so standard though) as wildcards. Algorithm is based on trie. Prefix matching worst case is `O(k)` where k is length of token we match. Suffix matching worst case is `O(k^2)`. But on average it should be much faster because algorithm tends to quit as soon as it sees that current branch does not match.

Note that `wildcardTrie.check` is a recursive function, but we fully control max depth by restricting max allowed length when adding patterns into the trie.
Also I covered this approach with tests. You can look at them for examples of wildcard usages.

What do you think? Should we keep it?

umputun:
a couple of things:

1. Those repeatable tests can be collapsed into a single table-based test like [this one](https://github.com/umputun/remark/blob/master/backend/app/cmd/cleanup_test.go#L25). I mean all `TestMatcher_Tokenize*` can be a single test and the same for `TestWildcardTrie*`
1. From a practical point of view - why we even care that much about flexible (and partially custom) wildcards? Probably the goal is to cover all forms of the word, but why not to do the simplest possible matching - strict match and "contains" match? We can have it as "bad-word" for strict matching and "/bad-word/" for contains. With all punctuation, you already suppress this will give us just enough flexibility with trivial implementation and decent speed.

_Pls note - I think the access to `patternMatchers` field is not safe. I had a similar comment in one of the reviews and suggested a concurrent test with `-race`,  but looks like you missed it._

amelnikov-mylivn:
1. Ok. I will do it. 
2. Is it ok if I just leave only one wildcard (`*`) handling and current trie structure? Or you want to match `exact` matches by map and all `contains` matches one by one? In this case I afraid that users will use `contains` for most of the matches. Because of languages like Russian have huge variety of prefixes and suffixes. This will lead basically almost to `O(m*n*k)` (because we will use contains for each pattern for each token).
3. (patternMatchers race) Best scenario will be to know all siteID on start and prebuild everything when creating Matcher (Is it easy to do or new site can be added in runtime?). 
Other solution is just to add read write lock. 
I will test with race and will do some benchmarks.

umputun:
> Is it ok if I just leave only one wildcard (*) handling and current trie structure? 

yeah, should be fine.

> or new site can be added in runtime

yes, sites can be added

umputun:
I'm still having some issues with [this part](https://github.com/umputun/remark/pull/257/files#diff-6d384e1763d29ca248214ef4267c8611R48). It expects any Listers to be static, however, this is the case for the current StaticRestrictedWordsLister implementation only. Generally, the source of restricted words could be some external store/db/service/whatever and it doesn't have to be static.

There are several things we can do to resolve it:

- provide `IsStatic` method to `RestrictedWordsLister` interface and allow your internal caching to work for static listers only
- change this internal caching to some other cache with TTL, i.e. cache everything but for a limited time only.
- get rid of the internal caching completely

My vote is for the last one. I think all of this is premature optimization and I'm not sure if the problem even real as we don't have any benchmarks proving how heavy such non-cached version is.

amelnikov-mylivn:
> get rid of the internal caching completely

Sorry I did not get this part. You want to remove `patternMatchers map[string]*wildcardTrie` completely and build new trie every time we need to check single comment? If so for me it is not a premature optimization - trie it is a pretty heavy data structure and whole idea is having it prebuild. Otherwise we can just downgrade to cycling through all restricted words and all tokens. 

A

amelnikov-mylivn:
so

amelnikov-mylivn:
> get rid of the internal caching completely

Sorry I did not get this part. You want to remove `patternMatchers map[string]*wildcardTrie` completely and build new trie every time we need to check single comment? If so for me it is not a premature optimization - trie it is a pretty heavy data structure and whole idea is having it prebuilt. Otherwise we can just downgrade to cycling through all restricted words and all tokens. 

Also is it a good idea for Lister to go into database or remote for the whole list of restricted words each time for each comment?

I understand that you value simplicity, but in this case I'd rather suggest to keep current solution and try to solve described issue in one of the following ways:

- Add GetVersion() into Lister and rebuild cached data structure when method returns bigger version. `StaticLister` will always return same version.
- Allow to provide listener into Lister (in this case it is no more a Lister) and invalidate cache each time Lister calls our listener (like when someone updates database). `StaticLister` will have a noop method implementation. 
- Your TTL idea.

amelnikov-mylivn:
> i.e check not just for != nil but for concrete error type or even text

Mostly the reason why I created ValidateCommentText was because I did not want to match text.
I will try to move restricted words check back into create and update methods of service.
Do you have any typed errors in the project already?

umputun:
> Also is it a good idea for Lister to go into database or remote for the whole list of restricted words each time for each comment?

This is just a detail of the particular Lister implementation. Sure some of them can cache response if needed.

> trie it is a pretty heavy data structure and whole idea is having it prebuilt

Do we have any benchmarks or at least some estimates showing how heavy is this in real life? Our input size is limited, so we can estimate at least the max number of tokens. Size of restricted words list harder to estimate, but we can use some worst-case scenario, let's say 1000. The question is - what overhead this will add to rest requests? Will we even notice such overhead in comparison to the relatively slow networking? If all we save here is a few microseconds I'd say - nah, doesn't worth the effort and complexity.

However, if benchmarking proves the problem, my choice will be to inject TTL cache as some generic interface with a single `Load(key string, ttl time.Duration, fn  func() (interface{}, error)) (interface{}, error)` method., smth similar to [this one](https://github.com/go-pkgz/lcw/blob/master/cache.go#L50) but with ttl.

umputun:
> Do you have any typed errors in the project already?

not sure, but I have nothing against typed error. I think I use "error const" to match in some places, i.e. `var errBlah = errors.New("blah")` and `if err = errBlah .... ` latter on

amelnikov-mylivn:
> nah, doesn't worth the effort and complexity

Ok. I think you convinced me :)

umputun:
LGTM

Issue #255, opened 2019-01-17
Telegram integration doesn't allow to send direct messages


VaultVulp:
Current integration allows sending messages only to public channels with `@ChannelName`-syntax. 

I guess it may be more useful to support direct messages and private channels. As described [here](https://core.telegram.org/bots/api#sendmessage) you can initialize the value of the `chat_id` parameter with integer id of a channel or user.

I suggest removing the code that prefixes any `chat_id` value with the `@` character before the payload is sent to the telegram API. It will allow developers to set any value (`@ChannelName` or `123455`) via the provided `NOTIFY_TELEGRAM_CHAN` environment variable.

If this proposition will be accepted, following file may be updated: [backend/app/notify/telegram.go](https://github.com/umputun/remark/blob/master/backend/app/notify/telegram.go). I suggest to remove both the trimming of the `@` prefix (line 33) and the addition of the `@` character in the process of API URL generation (line 95).

umputun:
Theoretically, such flexibility sounds like a good idea, but I don't see a practical use case for notifications to private channels and direct messages. What kind of setup you have in mind for such notifications?

VaultVulp:
I think it may help with the following problems:
- each user can only have a limited number of public channels/group names reserved for him, as of today, no more than 8 public names can be created.
- it will allow users to create a private source of notifications, in case they set up discussion platform in their internal network or behind some authentication wall, and receive notification w/o the fear of exposing their communication to someone.

Truth be told, I prefer to receive notification from my services via the private channel with my bot, and, as far as I know, multiple senders can send messages with the same token, so I would like to reuse the same token with the Remark and receive aggregated notifications through one source.

umputun:
> each user can only have a limited number of public channels/group names reserved for him, as of today, no more than 8 public names can be created.

Makes sense. I didn't know about 8 pub channels limitation

> it will allow users to create a private source of notifications, in case they set up discussion platform in their internal network

Such use case supported by remark42 partially. Currently, auth is oauth2 only. However, I can imagine firewalled instance allowing oauth2 only and limiting everything else.

The bottom line - I can see how the proposed change can be helpful. If you planned PR, pls proceed.

PR #258, opened 2019-01-21
Remove prefixing of a `@` character to the channel name

VaultVulp:
As proposed in #255 , following changes will improve versatility of a Telegram integration, thus allowing users to have a private source of notifications. 

I'm not that confident with Go, so this PR have following drawback:
- I was not able to provide test-case to cover refactored functions
- Refactored function may check the value of a `channelID` argument,  and append `@` in case it contains non-numeric literals

### DIFF
--- a/backend/app/notify/telegram.go
+++ b/backend/app/notify/telegram.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"html"
 	"net/http"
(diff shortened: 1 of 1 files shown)


umputun:
thx. I'll add a test and default to `@name` for non-numeric

PR #269, opened 2019-01-27
do not add .git folder to a build layer

nagimov:
`.git` folder isn't present in tarballs pulled from releases, e.g. `wget https://github.com/umputun/remark/archive/v.1.2.1.tar.gz`, and not used during the build anyways

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,6 @@ ARG MONGO_TEST
 
 WORKDIR /go/src/github.com/umputun/remark/backend
 ADD backend /go/src/github.com/umputun/remark/backend
-ADD .git /go/src/github.com/umputun/remark/.git
 
 # run tests
 RUN \


umputun:
not sure why it has anything to do with tarballs. They supposed to use $DRONE_TAG to form the version, but not .git. This part never worked properly because I'll need some sort of artifact storage on drone side and upload those artifacts to release. I have tried to implement it, but obviously, something is not right here and it doesn't keep/publish them to github properly.

regarding the conclusion - I agree, .git seems to be an artifact of some legacy method to get version number for local builds, but we don't use it anymore (defaulted to "local").

bnoctis:
It's still here on master: https://github.com/umputun/remark42/blob/20ca0896a610e40b37ad42c401ef6662bbfc862d/Dockerfile#L11

If it's just for a revision number, could it be done using the git command?

umputun:
git command won't work without .git folder.

bnoctis:
How about running the scripts locally then copying the results over?

umputun:
it means more complicated build process. I don't really see a reason to overcomplicate things. Build layer doesn't affect the final size of the container anyway

nagimov:
my two cents: having `ADD .git` breaks this workflow:
```
VER=1.10.0
mkdir ${VER}
wget https://github.com/umputun/remark42/archive/refs/tags/v${VER}.tar.gz
tar -xvf *.tar.gz -C ${VER}
cd ${VER}/*
docker build -t ${VER} .
```
since `.git` folder isn't included in release tarballs:
```
...
Step 9/53 : ADD .git/ /build/backend/.git/
ADD failed: file not found in build context or excluded by .dockerignore: stat .git/: file does not exist
```

umputun:
ah, I see. Have not thought anyone actually using tarballs instead of git checkout, but it is a valid concern for sure. I don't want to create some special make's target for the primary docker build and would rather prefer some special target to build container from the non-hit source.

nagimov:
For reference, `.git` directory was removed from the Dockerfile by this PR (#269) back in 2019 since it wasn't used in the Dockerfile then. It was re-introduced by aea77247694c1bd8d51313eb33422d7ed9559e32 that started using `git-rev.sh` to fetch the git version and tag the build with it.

I see two ways of solving this (assuming it actually needs "solving"):
1. stop using `git-rev.sh` and always tag builds with `version="local"` the way it used to work prior to aea77247694c1bd8d51313eb33422d7ed9559e32
2. make `git-rev.sh` fall back to `version="local"` if `.git` isn't found, and make `Dockerfile` copy `.git` conditionally only if it exists (not supported by docker officially although [hacky workarounds](https://stackoverflow.com/questions/31528384/conditional-copy-add-in-dockerfile) exist)

I fixed this for myself with `sed 's/...' Dockerfile` before building it :smile:

bnoctis:
Yeah, @nagimov made the point. I was trying to deploy it to Railway.app, and saw the same error, but the case wasn't as persuasive as tarballs.

Anyway, if all you need is `branch-hash-timestamp`, it could be done locally using `$(git branch --show-current)-$(git log -1 --format=%h)-$(date +%Y%m%dT%H:%M:%S)`. You are sending the same work directory to build after all. This could be run before `docker build`, setting an env var, then `ENV BUILD_TAG` in Dockerfile.

nagimov:
+1 for @bnoctis cleaner solution. Version ENV can be set via git in the Makefile and passed to `docker build --build-arg ...`
I would also add something like `[[ -z "$version" ]] || version="local"` to the Dockerfile to allow for unset ENV

PR #275, opened 2019-02-09
Hanlde limiter rejection on UI side

AnyRoad:
Issue https://github.com/umputun/remark/issues/256

- Adds common function to get error message from the http response
- Adds error message for the voting request failure displayed under the voting buttons
- Adds restoring of voting data after voting request failed

Example of the error message:

image

Handling error when user has been blocked:
image

### DIFF
--- a/web/app/components/comment/comment.jsx
+++ b/web/app/components/comment/comment.jsx
@@ -8,6 +8,7 @@ import { url } from 'common/settings';
 import store from 'common/store';
 import copy from 'common/copy';
 import debounce from 'utils/debounce';
+import { extractErrorMessageFromResponse } from 'utils/errorUtils';
 
 import Input from 'components/input';
 
@@ -23,6 +24,7 @@ export default class Comment extends Component {
       isEditing: false,
       isUserVerified: false,
       editTimeLeft: null,
+      voteErrorMessage: null,
     };
 
     this.votingPromise = Promise.resolve();
@@ -49,6 +51,8 @@ export default class Comment extends Component {
     this.isGuest = this.isGuest.bind(this);
     this.getUpvoteDisabledReason = this.getUpvoteDisabledReason.bind(this);
     this.getDownvoteDisabledReason = this.getDownvoteDisabledReason.bind(this);
+    this.handleVoteError = this.handleVoteError.bind(this);
+    this.sendVotingRequest = this.sendVotingRequest.bind(this);
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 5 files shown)


web/app/components/comment/comment.jsx
```diff
@@ -264,6 +268,27 @@ export default class Comment extends Component {
     }
   }
 
+  handleVoteError(e, originalVotingState) {
+    let errorMessage = extractErrorMessageFromResponse(e.response);
+    if (errorMessage.indexOf('user dev_user already voted for') === 0) {
```
igoradamenko: It's strange that you have dev stuff in the code. Is it correct here? Why?

web/app/components/input/input.jsx
```diff
@@ -92,16 +93,11 @@ export default class Input extends Component {
...
-        if (
-          e.response &&
-          e.response.data &&
-          typeof e.response.data.error === 'string' &&
-          e.response.data.error.indexOf("parent comment with reply can't be edited") === 0
-        ) {
-          this.setState({ isErrorShown: true, errorMessage: 'Comment has reply, editing is not possible' });
-          return;
+        let errorMessage = extractErrorMessageFromResponse(e.response);
+        if (errorMessage.indexOf("parent comment with reply can't be edited") === 0) {
```
igoradamenko: It looks like it's better to ask @umputun to return with error message some error code, because it's better to just have a map `code => text` than rely on text comparison.

web/app/components/comment/comment.jsx
AnyRoad: Sorry, it was my mistake. I've already fixed it and pushed the corrent version.

web/app/components/input/input.jsx
AnyRoad: I agree, it would be much easier to understand what happened if we have error code.
Another option is to have user-friendly error message we've got from the server side.

umputun: just to make sure: are you asking for an extra code to be added to [this json response](https://github.com/umputun/remark/blob/master/backend/app/rest/httperrors.go#L18) to map it on UI side to smth custom? And the reason for this? I mean why not to show the error text as-is, directly from the `details` field.

AnyRoad: Correct. If no languages beside the English are supported details should work fine. But seems like we can make it a little bit more UI friendly. Now values for the **details** are something like **"user blocked"** or **"old post, read-only"**. How about at least to make first letter capital and add dot in the end? Or even make it sound like complete sentence (ex. "You can not edit old posts"). Also sometimes the **details** field does not contain real error cause. For instance when user tried to edit comment having reply we've got following error:

```
{
   "details":"can't update comment",
   "error":"parent comment with reply can't be edited, a5bf336f-cb42-4b7d-980d-ed7c21aaebc0"
}
```
In order to provide better explanation for user we have to check the **error** as well.

umputun: makes sense

umputun: I have added `code` field to the error response. This is the list of all codes https://github.com/umputun/remark/blob/master/backend/app/rest/httperrors.go#L15

AnyRoad: I have changed the frontend part so it uses the `code` value.
@umputun I've also changed 1 go file. I believe it was typo, please check.
@igoradamenko Please review the code again.

Issue #276, opened 2019-02-11
Upgrade webpack-dev-server to version 3.1.11


umputun:
This is the alert from github:

<img width="855" alt="22kf4_20190210_200632 dwxdh" src="https://user-images.githubusercontent.com/535880/52543447-62940b80-2d6f-11e9-8f24-56b0ce8a70ce.png">

ping @Reeywhaar @igoradamenko

umputun:
tried to fix it in a naive way, by bumping the version to 3.1.11, but probably not a good idea to make backend dev to update npm deps ;) 

reverted

Reeywhaar:
If it soothes you, it is dev dependency that doesn't affect resulting bundle: for cve to be expoitable developer must be in public network, and only thing attacker can steal is remark's js source which is open-source. The minor annoyance is github banner, which is understandable :-).

The problem with this dep, is that webpack itself is highly outdated (remark's ^3.12.0 vs current 4.29.0) so to update dev-server, webpack probably must be updated as well by major version. Maybe not, but I can't find compatibility table.

PR #277, opened 2019-02-15
Update dependencies

Reeywhaar:
to #276 

Updated dependencies. Major difference is webpack 3 replaced with webpack 4.

Everything seems to be working fine, but I think there may be a lot of hidden things which I missed. So @igoradamenko, need your review


Reeywhaar:
well, build failed because of lint errors

PR #278, opened 2019-02-17
Add "contradiction index" metric

AnyRoad:
Issue https://github.com/umputun/remark/issues/274
* show the value of `controversy` in the hover on scores: value if it exists,
  "No controversy." if value is zero, "No votes yet." if there are no votes.
* add a new sorting mode "controversial" to "sort by" dropdown

image

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,6 +489,7 @@ type Comment struct {
     Locator   Locator         `json:"locator"` // post locator
     Score     int             `json:"score"`   // comment score, read only
     Votes     map[string]bool `json:"votes"`   // comment votes, read only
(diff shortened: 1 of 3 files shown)


umputun:
Not sure I understand what is "No controversy" and "No votes yet." and why we show it. If the value is 0 (or none) it should be just "Controversy: 0.00", why even make a special case for this?

AnyRoad:
Initially I thought it would be better to show some explanation why there is no controversy (no votes or all votes up/down), but now I agree that the reason is too straightforward to make special case. I will change to just display zero.

Issue #297, opened 2019-03-25
Voting response improvements 


umputun:
The current implementation of the voting delegates too much work to the client, sends a lot of useless data and exposes too many details. In fact, it sends back to the client the full list of all votes with user ids and +/- status to allow UI handle voting status, colors and so on. Such a list can be huge and we had to restrict maximum size (see #207).

The response (i.e. Comment) can be simplified by stripping `votes` map completely and replacing it by singular `vote` field indication +/- for the current user.

@igoradamenko & @Reeywhaar - do you see any potential problem I can cause to UI by making the change? Sure, it will need adjustment to read `vote` status, but to me, it sounds like a trivial change and even simplification of UI logic.

Reeywhaar:
I've thought recently that it possible for a user to find who up/downvoted them by looking at response and matching user id's. I think, yes, `{vote: {score: number, delta: -1/0/+1}}` value is enough to handle voting.

umputun:
I didn't plan to touch the existing `score` field, but just adding `score` : [0, -1, 1]

umputun:
@Reeywhaar - I have a functional version in `vote` branch with `votes` field stripped and new `vote` =1,-1,0 added

I can't merge it to master because it will break voting on UI side, but you may add your PR and test with `vote` branch. We have to release both at the same time.

PR #300, opened 2019-04-07
support new voting api in ui

Reeywhaar:
to #297 

Adds ui support for new voting api.

Backend side doesn't seem to work, `vote` is always 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,7 +507,7 @@ type Comment struct {
     User      User            `json:"user"`    // user info, read only
     Locator   Locator         `json:"locator"` // post locator
     Score     int             `json:"score"`   // comment score, read only
-    Votes     map[string]bool `json:"votes"`   // comment votes, read only
+    Vote      int             `json:"vote"`    // vote for the current user, -1/1/0.
     Controversy float64       `json:"controversy,omitempty"` // comment controversy, read only
     Timestamp time.Time       `json:"time"`    // time stamp, read only
     Edit      *Edit           `json:"edit,omitempty" bson:"edit,omitempty"` // pointer to have empty default in json response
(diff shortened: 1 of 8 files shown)


umputun:
did you test with `vote` branch?

Reeywhaar:
Yep, though I have tested against db left from 'anon users' branch, not with a new one. [remark.db.zip](https://github.com/umputun/remark/files/3051936/remark.db.zip)

umputun:
this is odd. I had a rich set of tests covering as much related to new `vote` functionality as I can think of and even added more. Pls note (just in case) - `vote` is visible (i.e. non-zero) for a regular (not an anonymous) authenticated user only and only for his own votes.

Reeywhaar:
oh, I've found the problem, I fetch 'GET' data without sending credentials.

umputun:
one more thing - the build failed due to some strange changes on travis side causing very slow start of tests. This had been addressed on master, but vote bench will fail.

umputun:
>  I fetch 'GET' data without sending credentials.

does it mean UI is working, but some manual tests you did with curl failed?

Reeywhaar:
Yeah, UI working, but comments fetched as from anonymous user.

Reeywhaar:
Updated pr.

Now something strange happening. There is like some sort of caching in `/api/v1/find` endpoint. If I login, upvote someone, then logout, and refresh comments (clean cookie, no XSRF, etc), response still contains `vote` on comment that I upvoted. Then if I login again, and upvote another comment, state seems to be updated to correct one, or to current–1 version :/

Screenshot 2019-04-08 at 01 19 37

```
GET /api/v1/find?site=remark&url=https://remark42.com/demo/&sort=-active&format=tree HTTP/1.1
Host: 127.0.0.1:9000
User-Agent: ///
Accept: application/json
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
Referer: http://127.0.0.1:9000/web/iframe.html?site_id=remark&url=https%3A%2F%2Fremark42.com%2Fdemo%2F
Content-Type: application/json
X-XSRF-TOKEN: 
DNT: 1
Connection: keep-alive
```

umputun:
This was an unexpected side effect of the introduced state. For now, I have added another element to the caching key making the cache user-specific but this is strange solutions as per-user cache for find doesn't make much sense. 

All of this seems to work correctly now (pls check too), however, I'll need to figure a better way to alter cached response with proper support of `vote` flag

Reeywhaar:
Rebased on top of your recent commits. Seems to be working fine.

Issue #299, opened 2019-04-05
Buttons for MD formatting 


umputun:
Having a typical toolbar for the post composer can be helpful. Something like the one we see on github or gitlab

<img width="497" alt="x9vfh-201904-05115349-babeo" src="https://user-images.githubusercontent.com/535880/55643799-a583c580-5799-11e9-82a3-194d13c52802.png">

<img width="456" alt="aiq3o-201904-05115519-gnh6k" src="https://user-images.githubusercontent.com/535880/55643823-b59ba500-5799-11e9-9f1c-fec6c3f6fe4b.png">

Mavrin:
Could we use a more user-friendly editor. Of course, it’s a good idea to limit the formatting options for the end user. It seems to me, it will a good solution, if the mark will use something looks like https://prosemirror.net/examples/markdown/

Mavrin:
Oh, I see, you use server render for preview. And client render is not appreciated solution for that case.

umputun:
- Currently, we do the server-side rendering for preview. With such editor, it could be a non-trivial task to keep preview and  WYSIWYM in full sync
- It is unclear to me how much this thing extendable for things like image drag-drop, counts and limits
- I'm not even sure if rich formatting for comments is a thing. I think 90% of users expecting a place to write a plain text, and maybe another 10% may want to drop a picture.

Mavrin:
I suggest using WYSIWYG as your default editor.
I see problem only with  your first point.
The second point possible solve by plugin https://prosemirror.net/examples/upload/ (drag-drop can be add)
Prosemirror generate pure plain text (try switch radio button between markdown and  WYSIWYM  in exapmle). If we are going add formatting button,  why we should show intermediate view for user.

Mavrin:
Whatever, it is not simple add synced server and client render. And it is good idea add simple button for textarea.

PR #305, opened 2019-04-11
#299 add toolbar for markdown formatting

Mavrin:
Solution for #299
It work the same way as it works on github. 
Underhood we use package from github team.

How it looks:
Screen Shot 2019-04-12 at 10 34 56 PM
Screen Shot 2019-04-12 at 10 35 15 PM


Mavrin:
Could anyone review it?  @Guria, @umputun, @igoradamenko

Reeywhaar:
Why silence here? Good job. Though I don't like that you store icons as react components. Better store them as plain svg and take advantage of https://www.npmjs.com/package/svg-inline-react

Mavrin:
@Reeywhaar I see only one advantage, that you can open icon in some image viewer. 
if I use real react component, I can change icon by props. e. g. `size`, `className`, `fill`. 
And I don't need add dependencies. 
I think for current case `svg-inline-react` is over-engineering

Reeywhaar:
Yep, this way it can be not only opened in viewer, but edited in redactor, given to designer, reused in some other place, etc.

`size`, `classname`, and `fill` (via css) all can be set on top element. I understand that sometimes loading this way can be handy if svg is highly dynamic. But this is not the case, it just icon.

I think better to load images as `require(`!raw-loader!some-image.svg`)` and make tiny functional component
```
function SVG(props) {
return <span className={"icon "+props.className} dangerouslysetinnerhtml={{_html: props.svgstringorsomething }}></span>
}
```
or something. I agree that `svg-inline-react` can be overload, but there lot of alternatives, for example https://www.npmjs.com/package/react-svg-loader which is bundle time thing.

Mavrin:
How is often this project change icons? For me, It is not problem covert 8 raw icons to svg manually.
`react-svg-loader` looks better. But It need add one more loader.

umputun:
with an extra toolbar the edit field looks too small. I'd suggest to make it 1 or even 2 lines higher

Mavrin:
4 lilines looks enough:
Screen Shot 2019-04-14 at 11 19 29 AM

5 lines:
Screen Shot 2019-04-14 at 11 21 46 AM

umputun:
> 4 lilines looks enough

agree

Mavrin:
Could you merge this request?

umputun:
I have merged, and it is live on https://remark42.com/demo/

However, it looks kind of strange to me. This new toolbar makes the edit section visually unbalanced.  Before the spaces on right, left and on the top were the same and it looked nice, but now it is not that nice anymore. 

I'm not sure how to fix it, open for suggestions/ideas.

Reeywhaar:
Not going far with design, but how about this? Considering mobile, and all things.
Screenshot 2019-04-15 at 07 00 06

umputun:
an interesting idea, but formatting toolbar on the bottom is unusual. Maybe the same but on top?

Reeywhaar:
Same on top
Screenshot 2019-04-15 at 07 05 50


Unusual, but I think it's convenient as with long message you will have to drag cursor more to top than to bottom.

umputun:
still odd. I may have a better idea - how about removing the additional toolbar completeley and move it to the bottom instead of the current text, i.e.
xtdcv_20190414_234347 ryrw8

Subscription action can go to some new "subscribe" button (with a dropdown) and the same for MD help.

Reeywhaar:
Unfortunately, on mobile it looks ugly then, conflicts with buttons on the left.

umputun:
maybe just collapse them all to a single button on mobile?

umputun:
another possibility is to hide this part on mobile, like github does

Reeywhaar:
Well, let's hide it then, I think burger button is to clumsy to use, and to close to "send button". But still I'm unsure about removing guide text, it contains useful info

Reeywhaar:
Well, I tried, and it seems there is too many "if". I think buttons look like they unrelated to text area.

bad
Screenshot 2019-04-15 at 09 01 21
bad
Screenshot 2019-04-15 at 09 01 10
fine
Screenshot 2019-04-15 at 09 01 00
passes
Screenshot 2019-04-15 at 09 00 54

Reeywhaar:
Hmm, I guess swapping is better.
Screenshot 2019-04-15 at 09 19 35
Screenshot 2019-04-15 at 09 19 50

Issue #285, opened 2019-03-10
Add drag and drop for images


umputun:
The goal is to allow direct drag-and-drop of pictures in github/gitlab style. On the backend will need a new api for upload and UI side will call it and insert md for the image.

umputun:
the code for the backend in `images` branch and it seems to work. However such protected upload and open download functionality introduce potential abuse vector - the bad guy may upload files and never submit the comment. 

I'm not sure how to fight this or if this even a real problem. If it is I can think of:
- some sort of "garbage collector" queueing all newly submitted images and making sure in the next 5 minutes we have a comment referencing the image
- make image upload two-step transaction and completed on a comment's post. Probably we will have to have some "temporary status" for images until a comment posted.
- allow access to images from remark42 iframe only
- some other way?

Reeywhaar:
Thought of it recently, then found drag'n'drop actually works, but it just pastes image url, so then you have to wrap it in \!\[\]\(%url\)

The easy way, I think, is to store an image as data url while preview, and upload only on sending comment. But its still needs management for case of deleted comment and such.

umputun:
some changes in place:

- image name changed from user-defined file (hash) to UUID
- added a top-level directory for users to allow easy removal of all images on "delete me" request

planned changes:

- make image upload 2-step process with the current `Save` storing pictures in a staging area and comments `Create` moving them to the permanent store. Need a way to get a list of all stored images from the comment.
- Add background removal for images sitting in the staging store for too long (EditDuration)
- rest API for image retrieval should be changed to try both long-term and staging. This is needed to allow preview rendering.
- Add parsing of comment to retrieve the list of images 
- Add `Remove` method for images

open question(s):

- not sure yet how to handle image removal from the long-term storage. For example, If comment edited and some links to images removed. Probably the second step should be delayed, i.e. Create call will move comments after EditDuration expired.

umputun:
backend part in place

PR #307, opened 2019-04-12
Image upload ui

Reeywhaar:
Adds ui support for #285

upload_screencap

Somehow even delay of 1500ms still hits rate limiter, anyway I set 5s delay between Image uploads, looks kinda reasonable.

### DIFF
--- a/web/app/common/fetcher.ts
+++ b/web/app/common/fetcher.ts
@@ -6,28 +6,45 @@ import { getCookie } from './cookies';
 export type FetcherMethod = 'get' | 'post' | 'put' | 'patch' | 'delete' | 'head';
 const methods: FetcherMethod[] = ['get', 'post', 'put', 'patch', 'delete', 'head'];
 
-type FetcherInit =
-  | string
-  | {
-      url: string;
-      body?: string | object | Blob | ArrayBuffer;
-      overriddenApiBase?: string;
-      withCredentials?: boolean;
-    };
+interface FetcherInitBase {
+  url: string;
+  overriddenApiBase?: string;
+  withCredentials?: boolean;
+}
+
+interface FetcherInitJSON extends FetcherInitBase {
+  contentType?: 'application/json';
+  body?: string | object | Blob | ArrayBuffer;
+}
+
+interface FetcherInitMultipart extends FetcherInitBase {
+  contentType: 'multipart/form-data';
+  body: FormData;
+}
+
+type FetcherInit = string | FetcherInitJSON | FetcherInitMultipart;
 
 type FetcherObject = { [K in FetcherMethod]: <T = unknown>(data: FetcherInit) => Promise<T> };
 
 const fetcher = methods.reduce<Partial<FetcherObject>>((acc, method) => {
   acc[method] = <T = unknown>(data: FetcherInit): Promise<T> => {
-    const { url, body = undefined, withCredentials = false, overriddenApiBase = API_BASE } =
-      typeof data === 'string' ? { url: data } : data;
+    const {
+      url,
+      body = undefined,
+      withCredentials = false,
+      overriddenApiBase = API_BASE,
+      contentType = 'application/json',
+    } = typeof data === 'string' ? { url: data } : data;
     const basename = `${BASE_URL}${overriddenApiBase}`;
 
     const headers = new Headers({
       Accept: 'application/json',
-      'Content-Type': 'application/json',
       'X-XSRF-TOKEN': getCookie('XSRF-TOKEN') || '',
     });
+
+    if (contentType !== 'multipart/form-data') {
+      headers.append('Content-Type', contentType);
+    }
 
     let rurl = `${basename}${url}`;
 
@@ -39,7 +56,9 @@ const fetcher = methods.reduce<Partial<FetcherObject>>((acc, method) => {
     };
 
     if (body) {
-      if (typeof body === 'object') {
+      if (contentType === 'multipart/form-data') {
+        parameters.body = body as FormData;
+      } else if (typeof body === 'object' && !(body instanceof Blob) && !(body instanceof ArrayBuffer)) {
         parameters.body = JSON.stringify(body);
       } else {
         parameters.body = body;
--- a/web/app/components/input/input.tsx
+++ b/web/app/components/input/input.tsx
@@ -7,13 +7,15 @@ import './styles';
 import { h, Component, RenderableProps } from 'preact';
 import b, { Mix } from 'bem-react-helper';
 
-import { User, Theme } from '@app/common/types';
+import { User, Theme, Image, ApiError } from '@app/common/types';
 import { BASE_URL, API_BASE } from '@app/common/constants';
 import { StaticStore } from '@app/common/static_store';
 import { siteId, url, pageTitle } from '@app/common/settings';
 import { extractErrorMessageFromResponse } from '@app/utils/errorUtils';
 
 import TextareaAutosize from './textarea-autosize';
+import { sleep } from '@app/utils/sleep';
+import { replaceSelection } from '@app/utils/replaceSelection';
 
 const RSS_THREAD_URL = `${BASE_URL}${API_BASE}/rss/post?site=${siteId}&url=${url}`;
 const RSS_SITE_URL = `${BASE_URL}${API_BASE}/rss/site?site=${siteId}`;
@@ -33,15 +35,22 @@ interface Props {
   getPreview(text: string): Promise<string>;
   /** action on cancel. optional as root input has no cancel option */
   onCancel?: () => void;
+  uploadImage: (image: File) => Promise<Image>;
 }
 
 interface State {
   preview: string | null;
   isErrorShown: boolean;
+  /** error message, if contains newlines, it will be splitted to multiple errors */
   errorMessage: string | null;
+  /** prevents error hiding on input event */
+  errorLock: boolean;
   isDisabled: boolean;
   maxLength: number;
+  /** main input value */
   text: string;
+  /** override main button text */
+  buttonText: null | string;
 }
 
 const Labels = {
@@ -50,7 +59,10 @@ const Labels = {
   reply: 'Reply',
 };
 
+const ImageMimeRegex = /image\//i;
+
 export class Input extends Component<Props, State> {
+  /** reference to textarea element */
   textAreaRef?: TextareaAutosize;
 
   constructor(props: Props) {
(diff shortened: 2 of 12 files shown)


umputun:
>  does /api/v1/picture support multiple files?

nop

> it very easy to hit request limit.

yeah, I know. in my todo list

Reeywhaar:
Do you plan to change upload api to allow multiple files? I guess then it's rather hard to reason about what size should be allowed, overall or per image, what to return in case of error, etc. I guess it's easier for me to upload consequently with 0.5 delay.

umputun:
> Do you plan to change upload api to allow multiple files? 

No,  and I don't think we should even support multiple files upload on UI side either. If someone need to drag-drop multiple images let him to do it one-by-one.

Reeywhaar:
It's just a state of things, I can't forbid user to select multiple files and drop them. After it happened I can either show error, like "Please one file at a time", or process them consequently. Second one is just more convenient.

Am I understanding right that setting `IMAGE_MAX_SIZE` to 0 will disable upload completely? I think it will be usefull to add data related to upload to `/api/v1/config` (I think something like `image_upload_size: int`) so ui can behave accordingly, i.e don't handle drop on disabled upload. Or checking for file size on client side.

umputun:
> Am I understanding right that setting IMAGE_MAX_SIZE to 0 will disable upload completely? 

There is no direct way to disable image upload, however setting size to 0 will effectively reject any image

> it will be usefull to add data related to upload to /api/v1/config

I have added `max_image_size` field to `/config` response

umputun:
>  Second one is just more convenient.

ok, makes sense.

Reeywhaar:
updated pr

umputun:
very nice, it works. However, I have minor visual-style issues:

- preview of (pre)generated  md images in the background looks a little bit strange
- progress indicator also looks to technical and not styled

To me, the github's way looks like a decent compromise between being not too technical, simple and somewhat useful. However, if such dynamic change from uploading to the final md line is too hard to implement, I'd suggest tweaking your progress screen to make it simpler and cleaner. Maybe non-transparent background during upload, not "preview" for final image md and a better position and better font style for the progress message.

Reeywhaar:
Well, yep, agree, think I need to try harder on this one.

Reeywhaar:
Updated

Reeywhaar:
Fixed

Reeywhaar:
One thing I noticed. Uploading svg passes, which can be dangerous, as it basically active document https://blog.online-convert.com/svg-file-and-its-danger/

web/app/components/input/input.tsx
```diff
@@ -148,10 +172,163 @@ export class Input extends Component<Props, State> {
       });
   }
 
-  render(props: RenderableProps<Props>, { isDisabled, isErrorShown, errorMessage, preview, maxLength, text }: State) {
+  /** appends error to input's error block */
+  appendError(...errors: string[]) {
```
Mavrin: Could you move this login to HOC? Like this  http://uidevdaily.com/2018/react-drag-and-drop-file-upload-hoc/ ?

web/app/components/input/input.tsx
```diff
@@ -148,10 +172,163 @@ export class Input extends Component<Props, State> {
...
+        const file = data[i];
+        const isFirst = i === 0;
+        const placeholderStart = this.state.text.length === 0 ? '' : '\n';
+
+        if (file.size > StaticStore.config.max_image_size) {
+          this.appendError(`${file.name} exceeds size limit of ${maxImageSizeString}`);
+          continue;
+        }
+
+        !isFirst && (await sleep(uploadDelay));
```
Mavrin: `sleep` for me it is `wait`

web/app/components/input/input.tsx
```diff
@@ -148,10 +172,163 @@ export class Input extends Component<Props, State> {
...
+    this.setState({
+      errorLock: true,
+      errorMessage: null,
+      isErrorShown: false,
+      isDisabled: true,
+      buttonText: 'Uploading...',
+    });
+
+    // fallback for ie < 9
+    if (!isSelectionSupported) {
```
Mavrin: Do you really need support IE less than 11

web/app/components/input/input.tsx
```diff
@@ -148,10 +172,163 @@ export class Input extends Component<Props, State> {
...
+      });
+
+      !isFirst && (await sleep(uploadDelay));
+
+      const result = await this.props
+        .uploadImage(file)
+        .catch(
+          (e: ApiError | string) =>
+            new Error(
+              typeof e === 'string'
```
Mavrin: Code duplication

web/app/components/input/input.tsx
Reeywhaar: Example you provided is overcomplicated for this current case. I don't need to bind to window, to dragleave, gragenter. Making it HOC assumes that I should provide props like `allowsMulptiple`, "allowsFiletype" and so on to make it generic for further reuse, which I think is meaningless now. If demand will appear, sure, it can be rewritten. But now let it be two methods.

web/app/components/input/input.tsx
Reeywhaar: and also `delay`, what purpose of renaming?

web/app/components/input/input.tsx
Reeywhaar: it's not to hard to make simple fallback, but yeah, purpose unclear.

web/app/components/input/input.tsx
Mavrin: The main idea extract logic which responsible for upload file to separate function. It is not about reuse code or parametrize. Link was just example. I don't recommend it implement it as is.

web/app/components/input/input.tsx
Reeywhaar: removed code duplication, also noticed that `this.textAreaRef.isSelectionSupported` returned `false` if selectionStart is 0, ouch.

web/app/components/input/input.tsx
Mavrin: I like `delay` too.
`sleep` has connotation in others language. Is is about suspends the execution of the current thread until the timeout interval elapsed.

Mavrin: I think logic about sequence upload file could be moved to UploadService (or something like this). It should not be in presentation level.
Let's imagine, if we add file input for upload, we need copy the same login from DND file.

Mavrin: If you move it to service you could add unit tests for conditions.

web/app/common/fetcher.ts
```diff
@@ -39,7 +48,9 @@ const fetcher = methods.reduce<Partial<FetcherObject>>((acc, method) => {
     };
 
     if (body) {
-      if (typeof body === 'object') {
+      if (contentType === 'multipart/form-data' && body instanceof FormData) {
```
Mavrin: Maybe  `contentType === 'multipart/form-data'` will be enough. Look like  `body instanceof FormData` is redundant.

web/app/components/input/input.tsx
Reeywhaar: That can be made not as HOC, but additional function which takes `File[]`, and processes them. Seems reasonable as after #305 this button should be added as well.

As for service, right now it tightly coulpled with ui, it constantly changes textarea value, so I don't think it's reasonable. You can add unit tests by mocking `Props.uploadImage`

web/app/common/fetcher.ts
Reeywhaar: you're right

web/app/components/input/input.tsx
Mavrin: > As for service, right now it tightly coulpled with ui
You are right. And it problem. We can create service which will notify about progress of upload files and in component we subscribe on that events. The the simple way it trigger redux action and add midleware.

Reeywhaar: As for `sleep` vs `delay`, it's so minor, that it pointless to discuss

umputun:
> Uploading svg passes

This was actually worse. Server side didn't care at all about data (except the size limit). I have added detection of image type to allow png, gif and jpeg only via this call.

Reeywhaar:
@umputun, please add webp support as well, as it always ignored (even by github) but actually is well supported

web/app/components/input/input.tsx
Reeywhaar: I think i'll leave it, as it mainly for ie, but also it just a simple fallback for those browsers that don't support `selectionStart`, browsers with support just get fancy textarea value manipulations.

web/app/components/input/input.tsx
Reeywhaar: > You are right. And it is problem. We can create service which will notify about progress of upload files and in component we subscribe on that events. The the simple way it trigger redux action and add midleware.

If there will be demand, sure, right now uploading presented only in Input component.

> The the simple way it trigger redux action and add midleware.
Nah, it's not simple, what is the point of redux and middleware here?

Mavrin: If you want decouple data and presentation layer. You need implement `Observer` for this case. But as we use redux for state management now we can use it for upload files.
how it could be look:
```js
//action Creator
function(files) {
  dispatch({type: 'UPLOAD_FILES_REUEST', PAYLOAD: {}})
  for(const file of files) {
      dispatch({type: 'UPLOAD_FILE_REUEST', payload: {fileName: 'fileName'}})
      fetcher
     .uploadFile(file)
     .then((res) => {
           dispatch({type: 'UPLOAD_FILE_REUEST_COMPLETED', payload: {fileName: 'fileName'}})
      })
     .then((res) => {
           dispatch({type: 'UPLOAD_FILE_REUEST_FAILED', payload: {fileName: 'fileName'}})
      })
     .then(() => {
          return delay(1500);
     });
  }
 dispatch({type: 'UPLOAD_FILES_REUEST_COMPLETED', PAYLOAD: {}})
}

```
```js
// in store we add state
{
...
files:{...}
}
```
And connect this state to input.
Yes middleware don't need here.
And this code will be easy reuse for upload from file input

Mavrin: IMHO. It is not about requirements. It is more about readability and maintainability.

Reeywhaar: what purpose `store.files` has?

Action you provided does different logic. It delays after every upload, but I need to delay only if there are several items, item is not first and on start of upload. And I don't even understand how to use it, should I use `store.files` as a queue? Hardly doubt that it increases readability and maintainability.

as for file input, it can be done easily as
```
class Input extends ... {
  ...
  onFileInputChange(e){
    const files = filterImages(Array.from(e.target.files))
    if (files.length === 0) return;
    this.uploadImages(files)
  }
  ...
}

umputun:
looking good. Is this version a final one I can merge, or anything else should be done?

Reeywhaar:
It's ok, though as #305 have been merged, one additional button (image upload) can be added to toolbar. Or it can be done via separate pr

umputun:
separate pr fine. merged

umputun:
the rejection message is ... strange. I guess this was nginx side limit caused this error, but probably client can handle it better

shxhj_20190414_230550 q5g1v

Reeywhaar:
hm, well, at least it not crashed :-) It's not particular about this pr, but for error handling as a whole, Right now api expects "{code: number, details: string, error: string}" json in case of error, and fallbacks to spawning whole body as an error text. I dunno, what should I do, parse every HTTP code or fallback to "Oops something wrong?" in case if error is not expected struct?

umputun:
yep, " something went wrong" will be ok message for such cases

umputun:
another problem - attempt to upload an image as anonymous rejected by backend and I think API returns rejection as StatusForbidden. However, UI shows 502 Bad gateway error (in the same strange way as above). Not sure if this is smth on backend or frontend, but either way I think the rejection of anonymous image upload can be done directly on UI

Reeywhaar:
Oh, strange but I haven't taken anonymous into account at all. Yep, you're right.

PR #308, opened 2019-04-15
Ui fixes

Reeywhaar:
* fixed unexpected api response, falls back to "Something went wrong"
* anonymous users unable to upload images
* adjusted toolbar style
  Screenshot 2019-04-15 at 09 31 05
  Screenshot 2019-04-15 at 09 31 17

Sorry for single pr, but I'm in hurry and will be available only next weekends.

### DIFF
--- a/web/app/@types/markdown-toolbar-element/index.d.ts
+++ b/web/app/@types/markdown-toolbar-element/index.d.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 declare namespace JSX {
   interface IntrinsicElements {
     'markdown-toolbar': {
--- a/web/app/common/fetcher.ts
+++ b/web/app/common/fetcher.ts
@@ -81,7 +81,9 @@ const fetcher = methods.reduce<Partial<FetcherObject>>((acc, method) => {
           try {
             err = JSON.parse(text);
           } catch (e) {
-            throw text;
+            // eslint-disable-next-line no-console
+            console.error(err);
+            throw 'Something went wrong.';
           }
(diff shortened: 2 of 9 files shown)


umputun:
the toolbar on the bottom looks good. However, text on top hurts my feelings and triggers OCD ;) 

Is the bottom part of the mobile version? If so, maybe we can add a toolbar button, like the one on github's right corner to show subscribe options. Generally, RSS subscription is not that essential to display it permanently and I don't see why not to put it under the button. I still think moving all this text under a button or two (md help and subscribe actions) will be nice and clear. And if those  extra buttons too much for mobile we may hide them, they not that important in the mobile version of UI

umputun:
another suggestion - how about we drop it for now :) The original is not that bad, and I think I can live with this. Could you pls revert changes related to the toolbar from the pr.

Reeywhaar:
Removed commit regarding toolbar style

umputun:
ef5m9-201904-21122129-ooqfe

is it expected?

Reeywhaar:
Oh, well, this must be rate limiter error (which one, remark's or nginx?) which must be returned as string (non-json) body, and I changed to show all non-json responses as "something went wrong" as you may remember

umputun:
yeah, all limiter errors returns text `You have reached maximum request limit.`, not json. However they have distinct code `429` you can recognize and display less generic error

Reeywhaar:
Yeah, I guess I should check for 429 especially, not sure, maybe some other codes that need to be handled that way?

umputun:
I'd say both auth related codes as well - `401` and `403`

Issue #310, opened 2019-04-16
Hard to understand that worst comment has replies.


Labutin:
Take a look at the worst comment here: https://radio-t.com/p/2019/04/02/prep-644/#remark42
It is not intuitive to click two times to collapse icon to be able see replies.
I think if replies were hidden automatically good idea to show additional icon/link to show hidden replies/comments.

Reeywhaar:
It's a bug, there should be plus sign, and comment should be collapsed

umputun:
@Reeywhaar is this an unexpected side effect from #281 ?

Reeywhaar:
Yep, I'm already fixed it, pr coming soon

PR #312, opened 2019-04-20
Fix useless comment collapse

Reeywhaar:
* fix for #310 

There is a thing. Comment below critical score being unfolded does not saves its collapse state in local storage, and with page reload will be collapsed again. This is because collapsed comments stored as plain list, and if comment is in it, then it collapsed. Useless comment being unfolded just doesn't save any thing to local storage. I don't know, should I fix it, or let we assume that this is "by design"?

Fix will be to store not in list, but in `map[comment id]boolean`.

### DIFF
--- a/web/app/common/api.ts
+++ b/web/app/common/api.ts
@@ -133,6 +133,7 @@ export const getUser = (): Promise<User | null> =>
     .get<User | null>({
       url: '/user',
       withCredentials: true,
+      logError: false,
     })
     .catch(() => null);
 
--- a/web/app/common/fetcher.ts
+++ b/web/app/common/fetcher.ts
@@ -10,6 +10,8 @@ interface FetcherInitBase {
   url: string;
   overriddenApiBase?: string;
   withCredentials?: boolean;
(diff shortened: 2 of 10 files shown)


umputun:
> Useless comment being unfolded just doesn't save any thing to local storage. I don't know, should I fix it, or let we assume that this is "by design"?

I'm trying to parse it and not sure I did it right. Are you asking if manually unfolded "bad comment" should stay unfolded after page refresh? I don't see why we may want it especially if some coding/support required for this.

Reeywhaar:
> I'm trying to parse it and not sure I did it right. Are you asking if manually unfolded "bad comment" should stay unfolded after page refresh? I don't see why we may want it especially if some coding/support required for this.

Exactly. Well, let it be that, then.

umputun:
some day we may add "user level settings" with various view customization, like "don't fold bad comment", "hide comments from user-defined list of users", etc. However even with this to-be-defined settings, such a granularity won't be needed. It will be all or nothing, i.e. unfold by default or not

PR #317, opened 2019-04-21
Rename "web"  to "frontend"

Reeywhaar:
I hope I got it right, though additional check from you will be helpful.

Assets still hosted with `http://{host}/web` prefix


Dockerfile.artifacts
```diff
@@ -9,19 +9,20 @@ ARG DRONE_BRANCH
 ENV SKIP_FRONTEND_TEST=true
 
 RUN apk add --no-cache --update git
-ADD web/package.json /srv/web/package.json
-ADD web/package-lock.json /srv/web/package-lock.json
+ADD frontend/package.json /srv/frontend/package.json
+ADD frontend/package-lock.json /srv/frontend/package-lock.json
 RUN cd /srv/web && CI=true npm ci
```
umputun: I think you missed this one. Btw, you can test this dockerfile by doing `make bin`

Reeywhaar:
Fixed `Dockerfile.artifacts `

PR #318, opened 2019-04-23
Code style fix in remark/web/counter.ejs

necinc:


umputun:
can you pls reapply on top of the current master to avoid a conflict?

Issue #324, opened 2019-05-12
Deleteme request doesn't work anymore


umputun:
_mentioned in this PR https://github.com/umputun/remark/pull/322#issue-277784617_

I don't see any regression on backend side and it's passing tests just fine. The manual attempt to call the method (`POST {{host}}/api/v1/deleteme?site_id={{site}}`) worked as expected, however it doesn't work from UI.

@Reeywhaar - to me it looks like somehow cookie not sent. As I inspect other requests, like `POST http://127.0.0.1:8080/api/v1/comment` i can see the cookie:

<img width="909" alt="405kc-201905-12131755-da973" src="https://user-images.githubusercontent.com/535880/57586209-5e83a100-74b8-11e9-8b24-42fcb87cefb0.png">

and this is what I see on deleteme:

<img width="948" alt="kiekf-201905-12131843-8yajz" src="https://user-images.githubusercontent.com/535880/57586224-7c510600-74b8-11e9-8614-00f348f1af33.png">

I have no clue why it may happen. Actually, this is very confusing - why browser didn't send the cookie in this case? maybe some iframe related issue?

Reeywhaar:
Made PR. There is confusion. Cookies not being sent is my bad, xhr request was sent without credentials, and I've fixed it in #322 and now in #326. My point starts later, see this steps:
* I log in as someone, leave comment
* request data removal, mail draft pops up
* log out, log in as dev_user (admin), follow the link from mail (`http://{remark}/web/deleteme.html?token={token}`)
* there I have message `forbidden` and request `GET http://{remark}/api/v1/admin/deleteme?token={token}` responds with 403, even though I've sent credentials.

What is proper way to approve data deletion, is `GET http://{remark}/api/v1/admin/deleteme?token={token}` right endpoint?

umputun:
> What is proper way to approve data deletion, is GET http://{remark}/api/v1/admin/deleteme?token={token} right endpoint?

yes, this is the correct one. Are you sure this one is sending cookie? Just fyi - token passed in GET is not for auth but for request verification only, it should send the regular jwt cookie as well

Reeywhaar:
Yep, I'm sure, cookies with JWT and `X-XSRF-TOKEN` header are being sent.

```
[...]

PR #326, opened 2019-05-12
Fix "delete me" feature

Reeywhaar:
to #324

### DIFF
--- a/frontend/app/common/api.ts
+++ b/frontend/app/common/api.ts
@@ -145,11 +145,13 @@ export const deleteMe = (): Promise<{
 }> =>
   fetcher.post({
     url: `/deleteme?site=${siteId}`,
+    withCredentials: true,
   });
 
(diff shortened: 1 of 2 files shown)


frontend/app/common/api.ts
```diff
@@ -150,7 +150,7 @@ export const deleteMe = (): Promise<{
 
 export const approveDeleteMe = (token: string): Promise<void> =>
   fetcher.get({
-    url: `/admin/deleteme?token=${token}`,
+    url: `/admin/deleteme?token=${encodeURIComponent(token)}`,
```
umputun: is it really needed? jwt supposed to be encoded already and token is jwt

Reeywhaar: yep, seems like overprotection from my side

Reeywhaar:
Checked again, everything works fine this time.

Issue #291, opened 2019-03-21
Hide comments from specific user


malamud:
As a user I would prefer not seeing comments from certain users.

Do I want to see answers on hidden user’s comments? No, I don’t.  
Do I need a possibility to make that user visible again? No, I don’t.

umputun:
This is not a bad idea, but I think such action should be reversible, i.e. "make that user visible again" better to be supported. Another thing to keep in mind - this functionality may require partial hiding of some subtrees.

As of now, we don't have any support for lists of muted users and no way to attach it to other users. I think, implementing backend side for this should be very doable and even not too hard. Not sure about UI side of this.

Another idea - maybe all of this can be done completely on UI side? @Reeywhaar - what do you think? Maybe even the list of such user can be stored locally in the browser and UI will do the actual hiding. I'm not sure how easy this to make on the browser side but doesn't sound like a big deal to me. Or, maybe an in-between solution with server-side storage API and hiding on UI side.

malamud:
I think it can be done completely on UI side. The list of muted users can be placed in browser storage or in cookies thereby muted user can be made visible by clearing storage or cookies. Maybe it’s not an elegant solution but the simplest one. Nevertheless if it possible to make list of muted user as part of UI it would be nice.

Reeywhaar:
Nothing complex here. Client only way is to store in localStorage.

Additionaly few endpoints can be added, so `/api/v1/user` will return additional list of blocked user ids, so list can be shared across multiple devices. Also, I haven't looked at #279 feature yet, what response `/api/v1/user` has this way.

umputun:
I don't think this has to support cross-device functionality. The reason for such "local filtration" can be not just "I hate this guy", but also "he's writing too many and too wordy comments  I don't want to see on iphone, but fine on mac". 

I suggest making it frontend-only function without any API changes, at least in the short term.

PR #334, opened 2019-05-19
Hide user comments

Reeywhaar:
to #291 

Adds ability to hide comments made by certain user.

Guest user:
Screenshot 2019-05-20 at 02 48 20

Logged in user:
Screenshot 2019-05-20 at 02 48 05

Admin:
Screenshot 2019-05-20 at 02 47 47

Right now it just plain button, there is no interface to manage hidden users. Also it's unclear what to do when user has logged out, and logged in as someone else — should list of hidden users be "per browser" or "per account"?

UPD: added settings interface for managing both hidden and blocked users, which replaces "blocked users" component

Screenshot 2019-06-10 at 01 30 46
--
Screenshot 2019-06-10 at 01 48 27
--
Screenshot 2019-06-10 at 01 50 26

There is no username in the list of blocked users right now because of #341, so just in case I've added "unknown"


umputun:
> should list of hidden users be "per browser" or "per account"?

per browser fine

>  there is no interface to manage hidden users

smth as simple as a list of users and ability to "unhide", similar to "ui" for blocked users we have for admin, should be enough 

acva4_20190519_190101 oujzh

Reeywhaar:
I think then both "blocked" and "hidden" manage buttons should be moved somewhere else, because "show hidden users" will be visible to guest and is bit confusing and I like how clear is interface when you not logged in. On second thought both these buttons can be visible only if user did actually hide or block someone. Also, these views should be merged. Also, I don't like that guest user has only one "hide" button in comment. Well, need to think on ui.

umputun:
any progress with this one?

Reeywhaar:
Unfortunately, I will not have time until next weekend

Reeywhaar:
I see strange output of `/api/v1/admin/blocked`, `id` has value of `name`, and the name is `empty`
```
[{"id":"banned_joe","name":"","time":"2019-06-16T15:02:16.860926708-05:00"}]
```

?

Reeywhaar:
filed an issue #341

Reeywhaar:
updated PR

Issue #346, opened 2019-06-15
Preview is broken


umputun:
@Reeywhaar this is smth recent, probably related to hiding functionality

Attempt to preview in UI shows nothing, the log has this:

<img width="1528" alt="86jlp-201906-15125929-tptki" src="https://user-images.githubusercontent.com/535880/59554761-bc028600-8f6d-11e9-8148-80e697f0d2f8.png">

_no error on the server side, good looking 200 response_

umputun:
@Reeywhaar let me know if you don't have time to fix it. I'd rather revert PR to restore preview because it is much more important functionality.

umputun:
related: don't we need a test on the frontend side to catch such regressions?

Reeywhaar:
Probably, I‘ll have time tomorrow’s evening, msk time.

PR #347, opened 2019-06-16
Fix ui errors (preview and image upload)

Reeywhaar:
to #346 

Apparently not only preview was broken, but image upload too :-)

I missed the thing that I was trying to connect functions which are not `actions`, but straight api calls via react-redux `connect`.

### DIFF
--- a/frontend/app/components/comment/comment.tsx
+++ b/frontend/app/components/comment/comment.tsx
@@ -19,6 +19,7 @@ import { Input } from '@app/components/input';
 import { AvatarIcon } from '@app/components/avatar-icon';
 import Countdown from '../countdown';
 import { boundActions } from './connected-comment';
+import { getPreview } from '@app/common/api';
 
 export type Props = {
   user: User | null;
@@ -43,6 +44,7 @@ export type Props = {
   theme: Theme;
   level?: number;
   mix?: string;
+  getPreview?: typeof getPreview;
 } & Partial<typeof boundActions>;
 
 export interface State {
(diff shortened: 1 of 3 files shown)


umputun:
Thx a lot. Is this thing can be tested automatically somehow?

Reeywhaar:
You mean end-to-end? Sounds interesting, in fact, as whole app is dockerized. Right now, unit-testing is barely exists, and I need to to think about the harness. As of now we test only plain `not-connected` components, and not the intermediate connection of api and components.

umputun:
I don't think the issue was fixed :( On demo site, it seems to work, but getting errors  on radio-t

pkfmj-201906-16161052-tj979

Reeywhaar:
Hmmm, how can I reproduce?

umputun:
whatever it seems to happen in safari, chorme is fine

umputun:
I got it by typing a new comment and clicking Preview

umputun:
hmm, maybe it just me/caching? not sure. Tried to reload the page multiple time and it started to work. So, maybe it was fixed ;)

Reeywhaar:
Seems to be cache issues, safari likes it. We don't have any cache policy, neither `Cache-Control`, nor `Etag`

Reeywhaar:
Oh, we have `Last-Modified`, but safari seems to ignore it

umputun:
> You mean end-to-end?

Not sure, have no clue what type of tests can be done on UI side. If end-to-end is the only way so be it.

PR #350, opened 2019-06-17
show threads

sw-double:
https://github.com/radio-t/radio-t-site/issues/94

Screenshot 2019-06-17 at 13 33 11

@Reeywhaar

### DIFF
--- a/frontend/app/components/comment/_level/comment_level.scss
+++ b/frontend/app/components/comment/_level/comment_level.scss
@@ -1,17 +1,10 @@
-$step: 24px;
-
-@for $i from 1 to 5 {
-  .comment_level {
-    &_$(i) {
-      padding-left: calc($i * $step);
-
-      &.comment_replying {
-        .comment__input {
-          @media (-moz-touch-enabled: 1) and (max-width: 768px), (pointer: coarse) and (max-width: 768px) {
-            margin-left: calc(-$i * $step);
-          }
-        }
-      }
-    }
-  }
+.thread > .thread {
+  padding-left: 17px;
+  border-left: 1px solid rgba(0, 0, 0, 0.15);
+}
+.thread_level_4 .thread > .thread {
+  padding-left: 5px;
+}
+.comment_theme_dark ~ .thread {
+  border-left-color: rgba(255, 255, 255, 0.15);
 }
--- a/frontend/app/components/comment/_theme/_dark/comment_theme_dark.scss
+++ b/frontend/app/components/comment/_theme/_dark/comment_theme_dark.scss
@@ -1,6 +1,4 @@
 .comment_theme_dark {
-  border-bottom: 1px solid #333;
-
   .comment__action_type_collapse {
     border-color: #555;
     color: #555;
--- a/frontend/app/components/comment/_theme/_light/comment_theme_light.scss
+++ b/frontend/app/components/comment/_theme/_light/comment_theme_light.scss
@@ -1,6 +1,4 @@
 .comment_theme_light {
-  border-bottom: 1px solid #fcfcfc;
-
   .comment__action_type_collapse {
     border-color: #ddd;
     color: #ddd;
--- a/frontend/app/components/comment/comment.scss
+++ b/frontend/app/components/comment/comment.scss
@@ -2,8 +2,7 @@
   display: block;
   position: relative;
   overflow: hidden;
-  margin-bottom: 16px;
-  padding-bottom: 8px;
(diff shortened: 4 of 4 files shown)


Reeywhaar:
Wow, thank you so much!

At first glimpse I see problem: Comments on level 5 and below should not be indented (think of mobile), yet somehow show that they are replies 🤔

umputun:
I think the line should be aligned to the left side of the first symbol. I know Reddit does it in the middle of the first one, but they have arrows here. 

>  Comments on level 5 and below should not be indented (think of mobile), yet somehow show that they are replies 🤔

I think the current indication on a deep level is fine but will be nice to fix it to actually jump to the parent comment

wbfcn-201906-17111237-qw6g2

umputun:
this is an example as NYT shows threads
7of6y-201906-29115551-hid8w

sw-double:
fixed alignment

umputun:
@sw-double & @Reeywhaar - what is the status of PR? is it still WIP or good to be merged?

sw-double:
not sure that it handles dark theme in an idiomatic way

otherwise LGTM on my end

umputun:
@sw-double cool, thx!

@Reeywhaar - need your blessing

Reeywhaar:
I'm working on ui and stream api, plan to end in next 3 hours, then I'll watch this PR.
@umputun hope you don't rush to make it before the show :-)
@sw-double, thank you so much 👍

umputun:
sure, take your time. thx

frontend/app/components/comment/_level/comment_level.scss
```diff
@@ -1,17 +1,10 @@
...
-          @media (-moz-touch-enabled: 1) and (max-width: 768px), (pointer: coarse) and (max-width: 768px) {
-            margin-left: calc(-$i * $step);
-          }
-        }
-      }
-    }
-  }
+.thread > .thread {
+  padding-left: 17px;
+  border-left: 1px solid rgba(0, 0, 0, 0.15);
```
Reeywhaar: Is it me, or lines color is too active in light theme?

Screenshot 2019-06-30 at 02 14 01

I'd rather mute it to rgba(0,0,0,.07)
Screenshot 2019-06-30 at 02 17 33

... or even to rgba(0,0,0,.05)
Screenshot 2019-06-30 at 02 14 15

... or another variant to make border dotted:
Screenshot 2019-06-30 at 02 13 29

frontend/app/components/comment/_level/comment_level.scss
```diff
@@ -1,17 +1,10 @@
...
+}
+.thread_level_4 .thread > .thread {
+  padding-left: 5px;
+}
+.comment_theme_dark ~ .thread {
```
Reeywhaar: I'd rather move this selector to `comment_theme_dark.scss` and make it like 
```
.comment_theme_dark {
  ...
  & ~ .thread {...}
}

Reeywhaar:
In general I see no crimes here, concise work, @sw-double, thank you, once again! So, @umputun, if you like it then great, I can fix some minor issues myself.

Reeywhaar:
I think we should make something with mobile layout
Screenshot 2019-06-30 at 02 36 32

umputun:
Generally, I like it, thx @sw-double . Two issues to address:

1. All internals "levels" don't need any visible thread lines. On such a deep level it is already impossible to understand the tree and adding zebra doesn't make it any better
1. I agree with @Reeywhaar - those lines are too visible. Your dotted variant looks better to me

4wxbw_20190629_191748 a4ra6

Reeywhaar:
I'm afraid without the lines it becomes even more unclear
Screenshot 2019-06-30 at 03 50 06

... I can make level5+ lines even more muted
Screenshot 2019-06-30 at 03 52 04

umputun:
Is there a reason why 5+ level indentations are any different? Or why we even have any for such deep levels?

Reeywhaar:
Not sure I follow. 5+ are indented less mainly for mobile and for big discussions. In old version, as you may remember, levels 5+ were unindented at all. Dunno what helps there actually, I think comments this deep are mess no matter what. What do you suggest?

Reeywhaar:
One another solution is to set min-width on thread, and if some discussiton breaks this limit then we get horizontal scroll
ezgif-2-1289a7ea05e1

umputun:
My suggestion is to revert 5+ to unindented

Reeywhaar:
ok, agree

PR #356, opened 2019-06-29
Fix: settings widget fails because it expects list but gets null

Reeywhaar:
When user is logged in, has no blocked users, and tries to open settings widget operation fails because reducer expects array of blocked users but api (`/admin/blocked`) returns null


umputun:
I'm not sure if null on empty response fro blocked is a new thing, but it wasn't intentional. Backend is trying to keep empty lists response in such cases. I'll fix it on my side as well

Issue #253, opened 2019-01-16
New comment fetch


Reeywhaar:
It's 2k19, and I think it will be good if remark have option to fetch new comments, just like in chat, where you don't have to reload a page to see new comment. Github also shows new comments immediately after it was posted.

So, idea from the design view looks like this:
![screenshot 2019-01-17 at 00 26 13](https://user-images.githubusercontent.com/884649/51280525-e701a180-19f0-11e9-8d9b-4876140d1066.png)
Notice thread margin on the left, and "2 new comments" button on the right, which upon click shows comments which appeared after user loaded the page.

From the backend/frontend side its:
* server option `autorefresh` or with values `off(default)|socket` or something.
* make use of https://github.com/googollee/go-socket.io which is basically websockets on steroids with option to downcast to http long polling and other features.
* `off` value can be used to turn off autorefresh, as lot of websocket connections can kill cheap server.
* on the frontend still https://socket.io/ may be used

Just my 2¢

umputun:
I have added `/stream/info` API allowing to subscribe to update events. See #336 for details

Reeywhaar:
Hope I'll get the time this weekend

umputun:
Do you think we need a way to control stream/no-stream from my side? For example, we can have this streaming optional globally and populate "streams": true/false as a part of `/config` response.

PR #357, opened 2019-06-29
Ui support stream api

Reeywhaar:
to #253 

Right now it's not ready (in terms of design), as it should be merged after #350


Reeywhaar:
Updated pr, it's more of working now:
Screenshot 2019-06-30 at 22 04 57

So there still some issues with this:
* Design: as you can see it's kinda ugly and cryptic, these triangles notify that top "dksdk" comment and its parent have updates. This is best I can think of, with the limitation of that indicator must not add vertical space or comments will be jumping which is irritating.
* I've added attribute `data-update-interval` to last comment widget (default 1 minute) so if there a lot of comments, one could read at least a bit. Don't know, maybe there should be checkbox `autoupdate` instead?
* Same thing with main widget. Maybe we should have same "autoupdate" checkbox which disables stream listening. This can be handy if post has a lot of comments. Right now demo.remark42.com post payload is ~100kb/500kb (zip/unzip), and this is what I have to fetch after every event. I'm concerned such constant fetching/rerenders can bury perfomance.
* Editing and deleting comments doesn't trigger stream updates.
* Noticed that, yeah, nginx closes stream after some time, so installation must require some additional actions to tame nginx
```
remark42-dev | 2019/06/30 18:48:59.658 [DEBUG] {api/stream.go:55 api.(*Streamer).Activate} stream closed due to timeout
```

So everything feels kinda raw right now, I'm not sure if it adds any value actually, kinda dissapointed :-)

Think I better to put it on a "for better times" shelf, and concentrate on performance, demo.remark42.com challenging me.

umputun:
> it's kinda ugly and cryptic

yeah, agree ;) maybe just a text label saying "comments were updated" or something like this? Can be a small font (to avoid extra vertical space) on some background? or maybe some  floating "box" / dialog?

> Don't know, maybe there should be checkbox autoupdate instead?

I don't think we need such flexebily. Whatever update interval we set on your side will be used by all users.

>  Maybe we should have same "autoupdate" checkbox which disables stream listening.

My opinion the same as above. Adding the ability to enable/disable updates can be confusing for the user and doesn't really solve the problem. However, I don't really understand the part  "I have to fetch everything after every event". Why do you need to fetch it? I think all you do is showing update info for the given thread and refresh it by user's click. Currently, you may lose some updates due to the limit=1 we have discussed, but this can be easily resolved on my side and each event may include the full list of updated comments. 

> Editing and deleting comments doesn't trigger stream updates.

Deleting supposed to trigger stream update, and I even recall testing it. Is it about the last comments or post's comments? Regarding edit - I have not considered edit as an event here, but this can be done if needed.

umputun:
@Reeywhaar - is this one still work in progress? or waiting for something from the backend to provide/fix/clarify?

Reeywhaar:
Yeah, it's kinda stalled. My general idea was to show new comments like this:
Screenshot 2019-07-24 at 02 15 19

I wanted to show, that some comment got reply or replies, and by clicking on it it would unwrap only this thread and not reload the whole comments.

And it seems stream api doesn't provide me with necessary things. What I have now is that I know that comment count changed. To actually get changed comments I have to fetch full post via `find` and diff with what I already have. Maybe I'm missing some part of api?

Right now api is suitable only to show that something changed and to show reload button, like habr has
Screenshot 2019-07-24 at 02 21 35
... but this button can be shown even without api. And it loses my main idea that I wanted a user have an ability to watch thread as it updates live. Thread should not jumping irritatingly while updates.

To solve this I can imagine adding another one endpoint (or extend `/find`) like this:
```
GET /find?url=someurl&since=128392834327

[
	{
		id: "1",
		text: "new comment",
		... // Comment struct 
	},
	{
		id: "2",
		delete: true
		... // this comment have been deleted
	},
]
```

... basically it's like `/find?format=plain` (or `/last` but for post) that returns only changed comments since some timestamp.

I think it's not trivial to implement on backend and need to take into account:
* User can edit comment >= 1 times since posting (only last edit should get into response)
* User can be banned (and then unbanned) since timestamp
* Comment can be deleted, etc., etc...

So my flow then:
* I get event that count changed
* I fetch response from endpoint above
* Applying changes

So, I'd be glad to hear your thoughts on this api, once again maybe I'm missing something.

Another one difficulty is design, but I think I know what to to with it.

umputun:
I think I have proposed a relatively easy change to stream api in some ticket or in some discussion, maybe just planned to propose. Anyway, this is what backend can do to help it:

1. Add full comment to the event. This is probably needed for `/stream/info` only (post's stream) but can be done for `/stream/last` as well. We may make it optional with some extra param like "comments=true" or smth like this
1. Make sure API doesn't drop "intermediate" updates to ensure UI won't miss anything if subscribed to the stream

With these two changes, you should be able to implement partial refreshes without hitting anything else

umputun:
I have played with a flavor of `/stream/info` returning full comment updates and I think it may not be such a great idea as I initially hoped. This stream will transfer comments to UI regardless if the user asked for a partial refresh or not, and may send potentially large data stream even if the user switched to another tab. Btw, can UI realize "tab in the background' and terminate streams? I'm not saying we want to deal with this, just curious.  

I think going with your suggestion by adding `since` to plain find will save us traffic, potentially a lot of traffic. However, I don't really understand why you think "endpoint will be fetched way more often than /find or /last". I thought you will hit this endpoint as user clicks on "2 new comments" link.

Regarding the rest of your concerns - well, we don't have to be perfect with this streaming thing, at least not right away. Having just new comments indication is a good start and we could deal with tricky cases, like edits/removal/bans latter on.

Reeywhaar:
Yes, we can stop looking for updates on tab blur.

I think returning full comments in stream is lot harder than doing it via stream and rest call, though possible. Main problem is that I need a way to ask server for some comments I skipped (either during network fail or while tab was incative)

> I thought you will hit this endpoint as user clicks on "2 new comments" link.

No, "2 new comments" is a result after I have parsed new data. Comments (data) already there, it just they marked as new and rendered that way. When user clicks they unwrap instantly.

> However, I don't really understand why you think "endpoint will be fetched way more often than /find or /last"

It will be fetched after every event received from stream api. Imagine hot topic where comments get posted every like 5 seconds, and considering there is 100 clients watching page, endpoint will be fetched 100 times every 5 second. That a lot more than just initial `/find` on page load that we have now.

umputun:
I see what you mean, and to me, it sounds too ambitious ;) How about something simpler and safer, like "2 new comments" on a floating text box appears on the top (or whatever place it can be visible) and the click just refresh the whole tree? I think this how discuss did it back in days. if you want to make it fancy, you could hit the new `find` with `since` and alter the tree in-place (partial refresh).

In this case, no extra API activity involved and no potentially heavy calls initiated.

Reeywhaar:
Well, to refresh partially I need to know which thread has new comments, chicken and egg problem.

What I don't like with refreshing whole tree is that it makes current thread you interested in to jump down, so you lose attention.

However, yes, we could make user click first at floating bar, and then instead of replacing tree in a dumb way, insert pointers to new comments. This way I can add a little of vertical space between comments as it only be inserted after user action, so I know he at least have read the thread he interested in.

umputun:
>  to refresh partially I need to know which thread has new comment

confused. if you click on this floating bar and fetch just with `/find?since` you may apply the change only and you will know which thread to update as each comment has parentid.  

> What I don't like with refreshing whole tree is that it makes current thread you interested in to jump down, so you lose attention.

this is not something we force user to do, he can click "update" or "dismiss". And we probably can try to restore the position on the screen after update

Reeywhaar:
Floating bar indicates new comments count for the whole post. To know that particular thread has new childs I have to hit `/find?since` first. Or you mean to update only thread that is in view? This is not trivial and should involve AI and Musk's neurointerface to determine which thread exactly user inteterested in.
Screenshot 2019-07-25 at 02 10 20

umputun:
> To know that particular thread has new childs I have to hit /find?since first.

yes. All I meant to say - it should be possible to hit /find?since instead of full /find and refresh only some elements of the tree instead of a full refresh. I'm not saying this optimization is necessary and I don't see anything wrong with full /find on click and full refresh of the tree.

Generally, my suggestion is to make it as simple as possible, whatever way you prefer technically. It can be full refresh with /find, applying diff with full /find or applying changes from /find?since. I any case this is about common floating bar not specifying particular thread, and refresh after the click not trying to be too nice/smart. If we can restore tree position even approximately it will be nice. If too much trouble - well, we can live with lost position.

paskal:
@Reeywhaar what is the state of this PR?

Reeywhaar:
@paskal it's stalled. Do not have spare time to work on it right now.

Issue #265, opened 2019-01-26
email notifications


umputun:
see #15 

I would like to implement a generic way to notify by email with customizable providers. This will require users to enter the email and we will have to store it somehow and associate with the user ID. We should also remove email if the user disabled notifications. Probably we may store it on the client side to be friendly and not to make users rentering email every time they turn it on/off.

umputun:
If anyone wants to deal with this -  let me know. This is not a trivial enhancement and we better to discuss it first.

paskal:
I'll try to do this one.

bronislav:
I'd like to work on it as well.

Issue #349, opened 2019-06-17
Email auth


umputun:
With https://github.com/go-pkgz/auth/issues/23 in place, we can add email authentication sending confirmation token back to the user. 

1. Implement `provider.Sender` for SMTP
1. add UI support (similar to what @Reeywhaar did for https://github.com/go-pkgz/auth/issues/25)
1. add options to activate and configure email auth
1. add SMTP container to compose example

umputun:
I have added backend support for `email` auth provider.  @Reeywhaar - the code is not merged to master but under email branch. To turn it on this [bunch of params](https://github.com/umputun/remark/blob/email/backend/app/cmd/server.go#L79) should be defined, however for testing the only needed one is `Enable` (AUTH_EMAIL_ENABLE = true). It should print to server log/out auth-token the same way as it does for the example app.

umputun:
Deployed to demo, seems to work. Gmail accepted the email, gravatar fetched.

@Reeywhaar - my email template produces an extremely ugly html, can you pls deuglify it a little bit and make more friendly (i.e., easier to select the token for example). This is [the current template](https://github.com/umputun/remark/blob/master/backend/app/cmd/server.go#L523)

umputun:
<img width="676" alt="e6m8r_20190706_184455 w1llp" src="https://user-images.githubusercontent.com/535880/60762114-3890fd80-a01e-11e9-8f1c-28d37282a655.png">

added to radio-t, frame looks strange

umputun:
also some space between inputs will be nice

umputun:
and, maybe, we better show providers as a dropdown ?

Reeywhaar:
Well, something should be done with dropdown positioning but it's not like easy. The whole top panel should be remade to adapt mobile. I guess it's a matter of another issue.

PR #362, opened 2019-06-30
[WIP] Enable email notifications

paskal:
This PR is enabling email notifications option, closing #265.


backend/app/cmd/server.go
```diff
@@ -565,6 +572,13 @@ func (s *ServerCommand) makeNotify(dataStore *service.DataStore) (*notify.Servic
 			return nil, errors.Wrap(err, "failed to create telegram notification destination")
 		}
 		return notify.NewService(dataStore, s.Notify.QueueSize, tg), nil
+	case "email":
+		email, err := notify.NewEmail(s.Notify.Email.Server, s.Notify.Email.Port, s.Notify.Email.Username,
```
umputun: way too many params, better to add `notify.EmailParams` struct

backend/app/notify/email.go
```diff
@@ -0,0 +1,114 @@
...
+			res.server, res.port, res.username, err)
+	}
+	err = tmpConn.Close()
+	if err != nil {
+		return &res, errors.Errorf(
+			"[WARN] error closing connection to %s:%d: %s",
+			res.server, res.port, err)
+	}
+
+	go func() {
```
umputun: I would extract this anon func to the normal/named, i.e. `func (e *Email) activate(ctx context.Context)`

umputun: we need a way to interrupt this goroutine from the outside

umputun: One more thing - we better keep email functionality as generic as possible, i.e. outside of notifier. Currently, there is one more ticket (#349) could leverage "send email" function. It uses very high-level interface for this:

```
// Sender defines interface to send emails
type Sender interface {
	Send(address string, text string) error
}
```

As long as your mailer implements this interface I could use it for other things

backend/app/notify/email.go
```diff
@@ -56,48 +65,11 @@ func NewEmail(server string, port int, username string, password string, keepAli
...
-							res.server, res.port, res.username, err)
-					}
-					open = false
-				}
-			}
-		}
-	}()
+	go res.activate()
 
 	// TODO: do we need to close this?
```
umputun: we do. At least to be able to properly test this. We can extend `Destination` interface with Close method and delegate the call to notify.Close

umputun:
I just noticed two new dependencies you've introduced to send emails (gopkg.in/gomail.v2 and gopkg.in/alexcesaro/quotedprintable.v3) Do we really need it? Sending email with stdlib directly seems to be easy enough https://golang.org/src/net/smtp/example_test.go

umputun:
in addition to the basic functionality, the most critical path to test is all async things. Tests also should make sure we are not leaking goroutines. Another test should simulate some heavy loads and make sure we not opening an endless number of goroutines.

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+				if s, err = d.Dial(); err != nil {
+					// Problems with connection, returning error and considering connection not established.
+					e.errChan <- errors.Errorf(
+						"error connecting to %s:%d with username %s: %s",
+						e.server, e.port, e.username, err)
+					break
+				}
+				open = true
+			}
+			if err = gomail.Send(s, m); err != nil {
```
umputun: I think, I have mentione it before - it makes sense to retry failed send. The same for failed Dial

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+				e.errChan <- errors.Errorf(
```
umputun: I see this patter in several place and don't think adding an extra line adds eny clarity. But it looks unusual for sure

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+					"error sending to %s:%d: %s",
+					e.server, e.port, err)
+			}
+			// Email sent successfully.
+			e.errChan <- nil
```
umputun: I don't see how you handle errors in errChan, but sending nil on seccesufl operation looks strange

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+	//m.SetAddressHeader("To", req.Address, req.Name)
+	m.SetHeader("Subject", fmt.Sprintf("New comment for \"%s\"", req.comment.PostTitle))
+	m.SetBody("text/html", messageBody)
+
+	// Wait for ability to send message and return error from error channel after sending it.
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case e.sendChan <- m:
+		return <-e.errChan
```
umputun: hmm, here you effectively enforce synced send. If your goal to send is syncroniosely why even bother with channel?

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+				close(e.errChan)
+				return
+			}
+			if !open {
+				if s, err = d.Dial(); err != nil {
+					// Problems with connection, returning error and considering connection not established.
+					e.errChan <- errors.Errorf(
+						"error connecting to %s:%d with username %s: %s",
+						e.server, e.port, e.username, err)
+					break
```
umputun: breaks inside of for look within select (or switch) are very confusing. It takes time to figure what it actually breaks. Will be better to avoid it

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+// Send email from request to address in settings
+func (e *Email) Send(ctx context.Context, req request) error {
+	log.Printf("[DEBUG] send notification via %s, comment id %s", e, req.comment.ID)
+
+	messageBody := prepareBody(req)
+
+	// Create message.
+	m := gomail.NewMessage()
+	// TODO: figure out where "from" and "to" addresses come from
+	//m.SetHeader("From", "no-reply@example.com")
```
umputun: from should be an input param for the constructor

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+	// Test connection before starting a daemon.
+	tmpConn, err := gomail.NewDialer(res.server, res.port, res.username, res.password).Dial()
+	if err != nil {
+		return &res, errors.Errorf(
+			"[WARN] error establishing test connecting to '%s':%d with username '%s': %s",
+			res.server, res.port, res.username, err)
+	}
+	if err = tmpConn.Close(); err != nil {
+		return &res, errors.Errorf(
+			"[WARN] error closing test connection to %s:%d: %s",
```
umputun: errors don't need levels, i.e. [WARN] part is for logging only, not error wrapping/reporting

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+		select {
+		case m, ok := <-e.sendChan:
+			if !ok {
+				close(e.errChan)
+				return
+			}
+			if !open {
+				if s, err = d.Dial(); err != nil {
+					// Problems with connection, returning error and considering connection not established.
+					e.errChan <- errors.Errorf(
```
umputun: you could use errors.Wrapf to add the context. This dependency already in use and makes error's wrapping nice and easy

backend/app/notify/email.go
```diff
@@ -0,0 +1,162 @@
...
+					"error sending to %s:%d: %s",
+					e.server, e.port, err)
+			}
+			// Email sent successfully.
+			e.errChan <- nil
+		// Close the connection to the SMTP server if no email was sent in the keepAlive period.
+		case <-time.After(e.keepAlive):
+			if open {
+				if err := s.Close(); err != nil {
+					// Problems with closing connection, considering connection still established.
```
umputun: I'm not sure this is a good idea. If we failed to close connection we can't consider that connection well and alive

backend/app/notify/email.go
paskal: To keep the connection alive and close it after a period without activity, we need a goroutine with a timer inside: I don't know another good way to close the connection by time in the event-based system.
If we have a goroutine which closes the connection, it makes sense to establish a connection and send emails from there as well.
From what I know communication between goroutines achieved with channels.
The only way around I come with so far is establishing a connection from Send (maintaining lock around it), and closing it right away if no one else waiting for the lock. It's still synced, but more complex in my opinion.
Is having multiple "event receivers" and single "event executer" in different goroutines and communicating through channels is a problem? And if so what is a better approach for keeping the stateful connection?

backend/app/notify/email.go
paskal: It's basically a reply: stop whatever we do and shut down the server because of the error. I'll do it differently after connecting to email server would be operated with retries.

backend/app/notify/email.go
paskal: It was a mistake from the time it was a log.Printf, fixed now.

backend/app/notify/email.go
paskal: Currently, it's somehow similar to `panic` (as it was originally): we just shut down the server in case of connection problems. That's why it have `break` inside.

backend/app/notify/email.go
```diff
@@ -0,0 +1,156 @@
...
+	if err = tmpConn.Close(); err != nil {
+		return &res, errors.Wrapf(err, "error closing test connection to %s:%d",
+			res.server, res.port)
+	}
+
+	// Activate server goroutine.
+	go res.activate()
+	// TODO: server goroutine never dies! This exiter should be run once first time we get context
+	// Close the channel to stop the mail daemon
+	// go func() { <-ctx.Done(); close(res.sendChan) }()
```
paskal: I plan to defer server shutdown only after the first request: beforehand no one is waiting to send an email and there is nothing to shut down. Only after the first request, it makes sense to monitor context.
As it turns out, a program currently is not designed in a way which would allow bringing context to `NewEmail` (called from `makeNotify` called from `newServerApp`).

backend/app/notify/email.go
```diff
@@ -0,0 +1,156 @@
...
+func (e *Email) Send(ctx context.Context, req request) error {
+	log.Printf("[DEBUG] send notification via %s, comment id %s", e, req.comment.ID)
+
+	messageBody := prepareBody(req)
+
+	// Create message.
+	m := gomail.NewMessage()
+	m.SetHeader("From", e.from)
+	// TODO: figure out where "to" addresses come from
+	//m.SetAddressHeader("To", req.Address, req.Name)
```
paskal: I have a question here on the scope of the changes: where should I get this `to` field from? Should I create a separate API method on subscribing to some change which will write email->post\author relationship into a database, so then on new comment `notify` will check it and generate multiple email events? (Last part can be altered to have database query and multiple send logic inside `email`, not on higher `notify` level). Or email module just need to send whatever comes to it, same as telegram counterpart for now?

backend/app/notify/email.go
```diff
@@ -0,0 +1,137 @@
...
+	res.connection, err = res.client()
+
+	return &res, err
+}
+
+// Send email from request to address in settings
+func (e *Email) Send(ctx context.Context, req request) error {
+	log.Printf("[DEBUG] send notification via %s, comment id %s", e, req.comment.ID)
+	msg := e.buildMessageFromRequest(req, "test@localhost")
+	return e.SendEmail(msg)
```
umputun: I think you missed auth. Not sure if you planned to do it in Send or other place, but will need smth like this anyway https://github.com/go-pkgz/auth/blob/master/provider/sender/email.go#L78

backend/app/notify/email.go
```diff
@@ -0,0 +1,137 @@
...
+	link := fmt.Sprintf("↦ <a href=\"%s\">original comment</a>", req.comment.Locator.URL+uiNav+req.comment.ID)
+	if req.comment.PostTitle != "" {
+		link = fmt.Sprintf("↦ <a href=\"%s\">%s</a>", req.comment.Locator.URL+uiNav+req.comment.ID, req.comment.PostTitle)
+		subject += fmt.Sprintf(" for \"%s\"", req.comment.PostTitle)
+	}
+	from := req.comment.User.Name
+	if req.comment.ParentID != "" {
+		from += " → " + req.parent.User.Name
+	}
+	// TODO: message looks bad, review it
```
umputun: well, caller can make it as nice as we want with html/template and injected template. Smth like this https://github.com/go-pkgz/auth/blob/master/provider/verify.go#L161

backend/app/notify/email.go
```diff
@@ -0,0 +1,137 @@
...
+	msg := fmt.Sprintf("%s\n\n%s\n\n%s", from, req.comment.Orig, link)
+	return e.BuildMessage(subject, msg, to, "text/html")
+}
+
+//BuildMessage generates email message
+func (e *Email) BuildMessage(subject, msg, to, contentType string) (message string) {
```
umputun: is there a plan to call this Build directly? I don't think so, but not 100% certain. If no such plans we should keep it private

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,90 @@
+package notify
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/umputun/remark/backend/app/store"
+)
+
```
umputun: see https://github.com/go-pkgz/auth/blob/master/provider/sender/email_test.go I think those tests will provide better coverage, especially this one https://github.com/go-pkgz/auth/blob/master/provider/sender/email_test.go#L68 doing end-to-end, except the actual smtp ops

umputun:
note: somehow I have removed "Draft" status. Not sure how, but it wasn't intentional

backend/app/notify/email.go
paskal: I've created it to make email more generic as was asked at the beginning of this PR. Made methods private, we'll make them public later if necessary.

backend/app/notify/email_test.go
paskal: I will do that after `Email.Send` will be written. Please don't review code until it's done: the main logic is yet absent.

backend/app/notify/email.go
```diff
@@ -0,0 +1,241 @@
...
+func (e *Email) sendEmail(message, to string) error {
+	e.count++
+	e.sendMutex.Lock()
+	client := e.SMTPClient
+	if client == nil { // if client not set make new net/smtp
+		c, err := e.client()
+		if err != nil {
+			return errors.Wrap(err, "failed to make smtp client")
+		}
+		client = c
```
golangcibot: ineffectual assignment to `client` (from `ineffassign`)

paskal:
Please feel free to review current state, two things are unresolved in the current code:
1. `To` field is now hardcoded
2. Multi-process sending is not tested now, I wanted first to see if that looks good or I should rewrite it for queue usage.

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,174 @@
...
+	fakeSMTP := &fakeTestSMTP{}
+	waitCh := make(chan int)
+	var waitGroup sync.WaitGroup
+	e := Email{EmailParams: EmailParams{}, SMTPClient: fakeSMTP}
+	for i := 1; i <= 10; i++ {
+		waitGroup.Add(1)
+		go func() {
+			// will start once we close the channel
+			<-waitCh
+			_ = e.sendEmail(fmt.Sprint(i), fmt.Sprint(i))
```
golangcibot: loopclosure: loop variable i captured by func literal (from `govet`)

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+// tmplData store data for message from request template execution
+type tmplData struct {
+	From      string
+	To        string
+	Orig      string
+	Link      string
+	PostTitle string
+}
+
+const emailConnectionTimeOut = 10 * time.Second
```
umputun: calling it defaultEmailConnectionTimeout or just defaultEmailTimeout will clarify the nature of the const

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+		auth := smtp.PlainAuth("", e.Username, e.Password, e.Host)
+		if err := c.Auth(auth); err != nil {
+			return nil, errors.Wrapf(err, "failed to auth to smtp %s:%d", e.Host, e.Port)
+		}
+	}
+
+	return c, nil
+}
+
+var msgTemplate = `{{.From}}{{if .To}} → {{.To}}{{end}}
```
umputun: I'd suggest defaultEmailTempl

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+
+//buildMessageFromRequest generates email message based on request
+func (e *Email) buildMessageFromRequest(req request, to string) (message string) {
+	subject := "New comment"
+	if req.comment.PostTitle != "" {
+		subject += fmt.Sprintf(" for \"%s\"", req.comment.PostTitle)
+	}
+	msg := bytes.Buffer{}
+	// we don't expect valid template to fail
+	_ = e.template.Execute(&msg, tmplData{
```
umputun: I would check for an error anyway. Not sure in what cases template execution may fail but generally if called function returns an error the consumer should handle it

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+	msg := e.buildMessageFromRequest(req, to)
+	err := repeater.NewDefault(5, time.Millisecond*250).Do(ctx, func() error {
+		return e.sendEmail(msg, to)
+	})
+	return err
+}
+
+// sendEmail sends message, prepared by Email.buildMessage
+func (e *Email) sendEmail(message, to string) error {
+	e.count++
```
umputun: use mutext with counter ti implement smeaphore-like syncronization is not needed, there is a sempathore https://godoc.org/golang.org/x/sync/semaphore

in addition, ++ on shared counter is not thread safe

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+		log.Printf("[WARN] failed to send quit command to %s:%d, %v", res.Host, res.Port, err)
+	}
+	if err := testSMTPClient.Close(); err != nil {
+		log.Printf("[WARN] can't close smtp connection, %v", err)
+	}
+	return &res, err
+}
+
+// Send email from request to address in settings
+func (e *Email) Send(ctx context.Context, req request) error {
```
umputun: is it supposed to be thread safe? If so, we better clarify in the comment

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+	to := "test@localhost"
+	msg := e.buildMessageFromRequest(req, to)
+	err := repeater.NewDefault(5, time.Millisecond*250).Do(ctx, func() error {
+		return e.sendEmail(msg, to)
+	})
+	return err
+}
+
+// sendEmail sends message, prepared by Email.buildMessage
+func (e *Email) sendEmail(message, to string) error {
```
umputun: Generally it's not clear to me what sendMail is doing, epscial the use of count. If there is a legit reason for this logic it should be expeined in the top comment to the function. However reading it I fail to see how this count increments/decerements help to achive the goal. I gues the goal is to queue it somehow, but not sure

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+		// count == 1 means that no one else waiting to send the message at this moment,
+		// e.SMTPClient is nil if Quit() call passed because it's closing connection as well
+		if e.count == 1 && e.SMTPClient != nil {
+			if err := e.SMTPClient.Close(); err != nil {
+				log.Printf("[WARN] can't close smtp connection, %v", err)
+			}
+			e.SMTPClient = nil
+		}
+		e.count--
+		e.sendMutex.Unlock()
```
umputun: this smells like a bug/dead-lock to me becaus unlock called above in some cases

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,173 @@
...
+	require.EqualError(t, err, "can't make email writer: failed")
+
+	assert.Equal(t, "from@example.com", fakeSMTP.mail)
+	assert.Equal(t, "to@example.com", fakeSMTP.rcpt)
+	assert.Equal(t, "", fakeSMTP.buff.String())
+	assert.False(t, fakeSMTP.quit)
+	assert.True(t, fakeSMTP.close)
+}
+
+func TestEmailMultipleSend(t *testing.T) {
```
umputun: in case if email send supposed to be thread safe we better test it concurrently to, with parallel sends.

backend/app/notify/email.go
```diff
@@ -0,0 +1,242 @@
...
+	return &res, err
+}
+
+// Send email from request to address in settings
+func (e *Email) Send(ctx context.Context, req request) error {
+	log.Printf("[DEBUG] send notification via %s, comment id %s", e, req.comment.ID)
+	// TODO: decide where to get "to" email from
+	to := "test@localhost"
+	msg := e.buildMessageFromRequest(req, to)
+	err := repeater.NewDefault(5, time.Millisecond*250).Do(ctx, func() error {
```
umputun: injecting repeater as dependency can be a good idea, especially for tests

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,173 @@
+package notify
```
umputun: tests should pass with the reace detector on, currenty it fails in TestEmailMultipleSend

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,173 @@
...
+	fakeSMTP := &fakeTestSMTP{}
+	waitCh := make(chan int)
+	var waitGroup sync.WaitGroup
+	e := Email{EmailParams: EmailParams{}, SMTPClient: fakeSMTP}
+	for i := 1; i <= 10; i++ {
+		waitGroup.Add(1)
+		go func() {
+			// will start once we close the channel
+			<-waitCh
+			_ = e.sendEmail("", "")
```
umputun: this is the test, skipping error check here is not the best idea. assert.NoError(t, e.sendEmail) feels better even it makes a little sense with fake agent

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func TestEmailSendErrors(t *testing.T) {
+	var err error
+	e := Email{EmailParams: EmailParams{FlushDuration: time.Second}}
+	e.template, err = template.New("test").Parse("{{.Test}}")
+	assert.NoError(t, err)
+	assert.EqualError(t, e.Send(context.Background(), request{}),
+		"error executing template to build message from request: template: test:1:2: executing \"test\" at <.Test>: can't evaluate field Test in type notify.tmplData")
+	e.template, err = template.New("test").Parse(defaultEmailTemplate)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,129 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	assert.EqualError(t, e.Send(ctx, request{}),
+		"canceling sending message to \"test@localhost\" because of canceled context")
+}
+
+// TODO: test writing is in process, more tests to come
+
+type fakeTestSMTP struct {
```
golangcibot: U1000: type `fakeTestSMTP` is unused (from `unused`)

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,129 @@
...
+func (f *fakeTestSMTP) Close() error         { f.close = true; return nil }
+
+func (f *fakeTestSMTP) Data() (io.WriteCloser, error) {
+	if f.fail {
+		return nil, errors.New("failed")
+	}
+	return nopCloser{&f.buff}, nil
+}
+
+type nopCloser struct {
```
golangcibot: U1000: type `nopCloser` is unused (from `unused`)

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,296 @@
...
+		6: {smtp: &fakeTestSMTP{}, amount: 1, quitCount: 0, waitForTicker: true}, // one message sent by timer
+	}
+	for i, d := range testSet {
+		email, err := NewEmail(EmailParams{BufferSize: 3, From: "test_sender", FlushDuration: time.Millisecond * 200})
+		assert.Error(t, err, "error match expected on test run #%d", i)
+		assert.NotNil(t, email, "email returned on test run #%d", i)
+
+		email.smtpClient = d.smtp
+		waitCh := make(chan int)
+		ctx, cancel := context.WithCancel(context.Background())
```
golangcibot: lostcancel: the cancel function is not used on all paths (possible context leak) (from `govet`)

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,296 @@
...
+			time.Sleep(time.Millisecond * 100)
+			// wait for another batch of email being sent
+		}
+		assert.Equal(t, d.quitCount+1, d.smtp.readQuitCount(), "connection closed expected amount of times on test run #%d", i)
+		if d.waitForTicker {
+			cancel()
+			assert.Equal(t, d.quitCount+1, d.smtp.readQuitCount(), "connection closed expected amount of times on test run #%d", i)
+		}
+	}
+}
```
golangcibot: lostcancel: this return statement may be reached without using the cancel var defined on line 148 (from `govet`)

paskal:
Code and tests for email sending are done, please review.
There are two known problems in the code, marked as `// TODO:` one corner case test and the fact that `to` field is now hardcoded.

paskal:
I've figured out the problem with tests, it's a hard-to-test concurrent case which is tested with other cases anyway and was removed.
The module is ready and the hardest part left, we need to figure out where to get "to" email from.

umputun:
>  we need to figure out where to get "to" email from.

I can see 2 ways to get this info:

1. Extend the info stored for the user (`users` bucket) with email and add a method to extract this info. Generally speaking, this may work for any auth using/supporting email directly or as a part of the scope. But I don't like this solution as it forcing us to store emails for all users, even those who don't care about email notifications. Another problem - this will be breaking change and some migration will be required.
1. Add a new API and new store.engine/store.service method(s) to save, load and clean email for the user. I like this one much more as we keep only necessary emails. Adding email should be two-step process with email confirmation code/link, smth like go-pkgz/auth does for `VerifProvider`

paskal:
I'll recreate this draft a little later if you don't mind: it would be too much to have 29 commits when only half of the work is done.

Issue #371, opened 2019-07-12
Add a button for image upload


umputun:
Apparently, users have no idea pictures can be dropped into comments. In addition, there is no way to do it on the mobile version. Will be nice to add an image upload button to the toolbar with a hint saying something about drag-and-drop as well.

umputun:
ping @Mavrin

Mavrin:
What do you think about this 
<img width="663" alt="Screen Shot 2019-07-12 at 11 35 02 AM" src="https://user-images.githubusercontent.com/1193817/61114539-31904200-a499-11e9-9b42-547dbe43ffc9.png">

umputun:
I'm not sure why this button has a text label

Mavrin:
Because it not clear, what that icon will do. For example, gitlab and github describe how to upload files.
Of course it is not problem to remove text label

umputun:
well, other buttons are not clear the same way, but the tooltip is enough to figure.

umputun:
@Reeywhaar is cut-and-paste images via clipboard suppose to work? I think it worked once, but can be wrong

Reeywhaar:
No, there is no any code related to this.

umputun:
> No, there is no any code related to this.

Can we add some code related to this? I mean if not a big deal will be nice to support it. To me, such functionality looks exotic, but somehow users expect it to work with the clipboard as well

Reeywhaar:
Yeah, I am myself find this feature good to have. Github has this, so I think it's totally possible.

mohammedi-haroune:
The upload button is still missing from the demo ? is it available for self-hosted installation ?
![remark42](https://user-images.githubusercontent.com/23096288/77619593-d4f68980-6f38-11ea-9259-df23eb6dce9a.png)

umputun:
I can see it on the demo site:

<img width="695" alt="buxvqovxat-20200326-022447" src="https://user-images.githubusercontent.com/535880/77620856-4b7b9300-6f09-11ea-8612-6978bb8c5f67.png">

The only reason I can think of - you logged as anonymous. In this case image upload disabled by design https://github.com/umputun/remark42/blob/master/backend/app/rest/api/rest.go#L320

mohammedi-haroune:
Thank you @umputun for your response.

PR #372, opened 2019-07-13
Add a button for image upload

Mavrin:
#371 How it looks
Screen Shot 2019-07-13 at 5 08 05 PM

### DIFF
--- a/frontend/app/components/input/input.tsx
+++ b/frontend/app/components/input/input.tsx
@@ -354,7 +354,11 @@ export class Input extends Component<Props, State> {
         onDrop={this.onDrop}
       >
         <div className="input__control-panel">
-          <MarkdownToolbar textareaId={this.textareaId} />
+          <MarkdownToolbar
+            allowUpload={Boolean(this.props.uploadImage)}
+            uploadImages={this.uploadImages}
+            textareaId={this.textareaId}
+          />
         </div>
         <div className="input__field-wrapper">
           <TextareaAutosize
--- a/frontend/app/components/input/input.tsx
+++ b/frontend/app/components/input/input.tsx
@@ -136,11 +136,12 @@ export class Input extends Component<Props, State> {
     });
   }
 
-  onPaste(e: ClipboardEvent) {
-    if (e.clipboardData && e.clipboardData.files.length > 0) {
-      const files = (e.clipboardData.files as unknown) as File[];
-      this.uploadImages(files);
+  async onPaste(e: ClipboardEvent) {
+    if (!(e.clipboardData && e.clipboardData.files.length > 0)) {
+      return;
     }
+    const files = Array.from(e.clipboardData.files);
+    await this.uploadImages(files);
   }
 
   send(e: Event) {
--- a/frontend/app/components/input/input.tsx
+++ b/frontend/app/components/input/input.tsx
@@ -92,6 +92,7 @@ export class Input extends Component<Props, State> {
     this.appendError = this.appendError.bind(this);
     this.uploadImage = this.uploadImage.bind(this);
     this.uploadImages = this.uploadImages.bind(this);
+    this.onPaste = this.onPaste.bind(this);
   }
 
   componentWillReceiveProps(nextProps: Props) {
@@ -133,6 +134,13 @@ export class Input extends Component<Props, State> {
       preview: null,
       text: (e.target as HTMLInputElement).value,
     });
+  }
+
+  onPaste(e: ClipboardEvent) {
+    if (e.clipboardData && e.clipboardData.files.length > 0) {
+      const files = (e.clipboardData.files as unknown) as File[];
+      this.uploadImages(files);
+    }
   }
 
   send(e: Event) {
@@ -363,6 +371,7 @@ export class Input extends Component<Props, State> {
         <div className="input__field-wrapper">
           <TextareaAutosize
             id={this.textareaId}
+            onPaste={this.onPaste}
             ref={ref => (this.textAreaRef = ref)}
             className="input__field"
             placeholder="Your comment here"
--- a/frontend/app/components/input/input.tsx
+++ b/frontend/app/components/input/input.tsx
@@ -140,6 +140,7 @@ export class Input extends Component<Props, State> {
     if (!(e.clipboardData && e.clipboardData.files.length > 0)) {
       return;
     }
+    e.preventDefault();
     const files = Array.from(e.clipboardData.files);
     await this.uploadImages(files);
   }
--- a/frontend/app/components/input/markdown-toolbar.tsx
+++ b/frontend/app/components/input/markdown-toolbar.tsx
@@ -34,7 +34,7 @@ const codeLabel = 'Insert a code';
 const linkLabel = 'Add a link <cmd-k>';
 const unorderedListLabel = 'Add a bulleted list';
 const orderedListLabel = 'Add a numbered list';
-const attachImageLabel = 'Attach an image';
+const attachImageLabel = 'Attach the image, drag & drop or paste from clipboard';
 
 export default class MarkdownToolbar extends Component<Props> {
   constructor(props: Props) {
--- a/frontend/app/components/input/markdown-toolbar.tsx
+++ b/frontend/app/components/input/markdown-toolbar.tsx
@@ -13,12 +13,12 @@ import OrderedListIcon from './markdown-toolbar-icons/ordered-list-icon';
 
 interface Props {
   textareaId: string;
-  uploadImages: (files: File[]) => void;
+  uploadImages: (files: File[]) => Promise<void>;
   allowUpload: boolean;
 }
 
 interface FileEventTarget extends EventTarget {
-  readonly files: File[] | null;
+  readonly files: FileList | null;
   value: string | null;
 }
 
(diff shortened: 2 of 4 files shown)


umputun:
LGTM, works fine, thank you. The only thing I'd like to change a little bit - the image icon looks larges compared to others. Maybe make it smaller or use a different icon?

umputun:
another thing - currently hint is saying "Attach the image". Will be nice to change "Attach image, drag & drop, or paste from clipboard"

Mavrin:
I changed hint and decrease icon size. But before all icon had 16px height. 
image

frontend/app/components/input/markdown-toolbar.tsx
```diff
@@ -22,8 +34,21 @@ const codeLabel = 'Insert a code';
...
+const attachImageLabel = 'Attach the image, drag & drop or paste from clipboard';
 
 export default class MarkdownToolbar extends Component<Props> {
+  constructor(props: Props) {
+    super(props);
+    this.uploadImages = this.uploadImages.bind(this);
+  }
+  uploadImages(e: Event) {
+    const currentTarget = (e as FileInputEvent).currentTarget;
+    if (currentTarget && currentTarget.files && currentTarget.files.length > 0 && this.props.allowUpload) {
```
Reeywhaar: It's maybe my optinion but early return should be better
```js
  async uploadImages(e: Event) {
    const currentTarget = (e as FileInputEvent).currentTarget;
    if (!this.props.allowUpload || !currentTarget || !currentTarget.files || currentTarget.files.length === 0) return;

    const files = currentTarget.files;
    this.props.uploadImages(files);
    currentTarget.value = null;
  }
```

Also, `event.preventDefault()` would be good here, just in case

frontend/app/components/input/input.tsx
```diff
@@ -135,6 +136,13 @@ export class Input extends Component<Props, State> {
     });
   }
 
+  onPaste(e: ClipboardEvent) {
+    if (e.clipboardData && e.clipboardData.files.length > 0) {
```
Reeywhaar: Early return :-)

frontend/app/components/input/input.tsx
```diff
@@ -135,6 +136,13 @@ export class Input extends Component<Props, State> {
...
+      const files = (e.clipboardData.files as unknown) as File[];
```
Reeywhaar: Use Array.from to avoiding force casting

frontend/app/components/input/input.tsx
```diff
@@ -135,6 +136,13 @@ export class Input extends Component<Props, State> {
...
+      this.uploadImages(files);
```
Reeywhaar: You should use `e.preventDefault` or else code works incorrectly. Attempting to upload file with name `Screenshot 2019-07-15 at 23.40.48.png` results in `.40.48.png` being inserted in input.

```
  onPaste(e: ClipboardEvent) {
    if (!e.clipboardData || e.clipboardData.files.length === 0) return;

    e.preventDefault();
    const files = Array.from(e.clipboardData.files);
    this.uploadImages(files);
  }
```

frontend/app/components/input/markdown-toolbar.tsx
Reeywhaar: hmm, found out, that if you're trying to paste more that one image, then pasting doesn't work because `clipboardData.files.length === 0` in that case. Even on github. Strange :-)

frontend/app/components/input/input.tsx
Reeywhaar: hmm, found out, that if you're trying to paste more that one image, then pasting doesn't work because clipboardData.files.length === 0 in that case. Even on github. Strange :-)

frontend/app/components/input/markdown-toolbar.tsx
```diff
@@ -48,6 +73,12 @@ export default class MarkdownToolbar extends Component<Props> {
           <md-link className="input__toolbar-item" title={linkLabel} aria-label={linkLabel}>
             <LinkIcon />
           </md-link>
+          {this.props.allowUpload ? (
+            <label className="input__toolbar-item" title={attachImageLabel} aria-label={attachImageLabel}>
+              <input class="input__toolbar-file-input" type="file" onChange={this.uploadImages} />
```
Reeywhaar: you can add `multiple={true}` to allow multiple uploads

frontend/app/components/input/markdown-toolbar.tsx
```diff
@@ -7,11 +7,23 @@ import ItalicIcon from './markdown-toolbar-icons/italic-icon';
 import QuoteIcon from './markdown-toolbar-icons/quote-icon';
 import CodeIcon from './markdown-toolbar-icons/code-icon';
 import LinkIcon from './markdown-toolbar-icons/link-icon';
+import ImageIcon from './markdown-toolbar-icons/image-icon';
 import UnorderedListIcon from './markdown-toolbar-icons/unordered-list-icon';
 import OrderedListIcon from './markdown-toolbar-icons/ordered-list-icon';
 
 interface Props {
   textareaId: string;
+  uploadImages: (files: File[]) => void;
```
Reeywhaar: should be `uploadImages: (files: File[]) => Promise<void>;`

frontend/app/components/input/markdown-toolbar.tsx
Reeywhaar: You are clearing currentTarget.value (why, btw?) before upload have finished, which causes only first file to be uploaded. You should `await` there

```js
  async uploadImages(e: Event) {
    const currentTarget = (e as FileInputEvent).currentTarget;
    if (!this.props.allowUpload || !currentTarget || !currentTarget.files || currentTarget.files.length === 0) return;

    const files = currentTarget.files;
    await this.props.uploadImages(files);
    currentTarget.value = null;
  }
```

Reeywhaar:
So, added a bit of review, if it's not hard to change, will be great. Otherwise good work, @Mavrin thank you so much!

frontend/app/components/input/markdown-toolbar.tsx
Mavrin: I also prefer early return, but for this case three negative statement to read more hard than three direct positive statement.

frontend/app/components/input/markdown-toolbar.tsx
Mavrin: I didn't catch, that function returns promise, because this function doesn't have any explicit return and have many return.

Reeywhaar: Yeah, I understand, it just to mark that it's asynchronous, so it won't be confusing that you await it further in "uploadImages" method below.

frontend/app/components/input/input.tsx
Mavrin: It is because clipboard data can contains different object types, this behaviour depend on  OS and browser

Reeywhaar: Yeah, I suspect it's Firefox issue

frontend/app/components/input/input.tsx
```diff
@@ -135,6 +136,14 @@ export class Input extends Component<Props, State> {
     });
   }
 
+  async onPaste(e: ClipboardEvent) {
+    if (!(e.clipboardData && e.clipboardData.files.length > 0)) {
+      return;
+    }
+    const files = Array.from(e.clipboardData.files);
```
Reeywhaar: Forgot `e.preventDefault()`. It's still buggy.
Screenshot 2019-07-16 at 02 16 06

frontend/app/components/input/markdown-toolbar.tsx
Mavrin: It is two different files, in second case IDE said me, that I called async function. Sad that typescript didn't catch my mistake.

frontend/app/components/input/input.tsx
Mavrin: Which are browser do you use?

Reeywhaar: Firefox 69b4.

Reeywhaar: But I don't think it's browser related. Upon paste I see that it pastes file's name. And uploadImages taking value of input while uploading to insert all this "...uploading" placeholders, so race appears, and I think what causes it is current flow
* image pasted
* uploadImages takes value of input
* event pastes image filename into input
* uploadImages deals with current cursor position and selection which and by this time taken value is out of sync.

Something like that, dunno. We should prevent pasting text data from event.

Mavrin: I check it on Mac
- Chrome upload only one file
- Firefox paste to file name
- Safari upload to file

Mavrin: I think it rare case and we can skip that.

Reeywhaar: Rare, Firefox users rare? 0_o

Mavrin: That user pastes two file

Reeywhaar: Oh, yeah, even github doesn't deal with it, it's fine to skip :-) It's browser's issue.

Reeywhaar: > I check it on Mac
Yeah, same, seems like it's state of api.

Mavrin: It  works without `e.preventDefault`. What is sense to make `preventDefault`?

Mavrin: I think it could be merged.

Reeywhaar: image

It works incorrectly. At least on firefox it leaves strange chunk of filename in input.

`preventDefault` prevents default which is paste of file's name.

Try to paste file with filename `Screenshot 2019-07-16 at 11.56.20.png` on firefox.

Mavrin: Hm, It works fine without prevent  in firefox on my laptop, but  if it helps I added it.

PR #381, opened 2019-07-22
Fix blocking

Reeywhaar:
To https://github.com/umputun/remark/issues/377#issuecomment-513491892

Now blocked comments rendered like on screenshot below, no matter whether user is admin or not:
Screenshot 2019-07-23 at 02 15 29

One thing I'm not sure is that if user was blocked permanently his comment (if it has childs) will be rendered like this:
Screenshot 2019-07-23 at 02 16 47
... so user traces still present. @umputun if you think this is undesired, I can tell that I'm actually cannot differentiate if user was blocked permanently or not on front, so to render such comment differently (ghosted avatar and so on) I need another one field in User struct.


umputun:
don't worry about it. I can deal with this on my side by enforcing names of permanently blocked users to  "user blocked" or smth like this

Issue #358, opened 2019-06-30
Frontend: Optimize comments reducer


Reeywhaar:
### The problem:
When adding, or editing comment we end up with full rerender of all comments tree due to how we store comments in store. We store them just like we receive them from api in a tree structure.
I see great degradation in rendering of https://remark42.com/demo/, and it will only intersify as soon we get #253

### Solution:
Split current comments reducer which has current structure:
```
type Node = {comment: {id: string, text: string, ...}, replies?: Node[]}

comments: Node[]
```
Into three objects:
```
top_comments: (Comment["id"])[], // We filter out top comments ids which allows us to keep sorting that we received from server
childs: {[key: Comment["id"]]: (Comment["id"])[]}, // Map of comment id to it's childs ids (`replies` field we get from server)
comments: {[key: Comment["id"]]: Comment}, // just map of comment's id to its content
```

Also we can add another reducer:
```
hashes: {[key: Comment["id"]]: string}, hash is just simple `${comment.id} + (comment.edit ? comment.edit.time : "")`
```
And with this we can make `shouldComponentUpdate` to check only if `nextProps.hash` equals to `this.props.hash` in such map and bailout of rerender easily.

### With these manipulations we get:

* Editing comment becomes replacing one object in `comments` reducer
* Adding comment becomes adding `Comment` in `comments` reducer, and pushing either to `childs[id]` reducer if it was reply, or pushing `id` to `top_comments`, so in case of adding of deep comment only `childs[id]` and `comments[id]` object get changed, and not the whole tree.

*I'm not sure if it really will help with performance, and even if performance is same, manupulating comments becomes much easier*

Reeywhaar:
done in #397

PR #397, opened 2019-07-29
UI Perf Improvements

Reeywhaar:
Added some improvements:
* Reconfigured reducers #358 
* Added visibility dependent logic for Comment component:
  A reason for this was poor feedback on any action on https://remark42.com/demo/. Cause of this is lot of DOM elements on page. So now when comment leaves view it gets rendered as simple hollow div. Noticeable feedback improvement for me, now "reply" button works instantly on my machine. Sorting change also a lot faster.
  Downside of this are:
    * Comments may flash during sorting change or initial load
    * CTRL-F on comments page (I mean that one that is like https://{remark}/web/iframe.html?...,  not on page with embed comments) doesn't find all data (as most of the comments are simple transparent divs, though I noticed that search doesn't work when iframe embedded on page neither)
  
  Also, not sure, if it downside or a feature, but content (images) now lazy loaded.
* Added es5 compatibility check on post build step. It ensures that code will be working on old devices.
* Updated demo page a bit to look better on mobile.


umputun:
this is a lot of changes. I have tried it locally and haven't seen anything bad/unexpected. I'll merge it and  see how this work on demo site

PR #402, opened 2019-08-01
Allow ID of container node to optionally be provided in comments widget config

johnpuddephatt:
This might not be essential for many people but it seems like a simple change and it seems sensible to allow people to change it; e.g. to follow their desired naming conventions.

This also introduces the additional possibility of having multiple sets of comments on one page (as long as different URL values are provided) – I want to use remark on a site for online consultation, with users able to comment on multiple sections of a proposal.

If there are repercussions of this change I've missed then apologies!

### DIFF
--- a/frontend/app/common/config-types.ts
+++ b/frontend/app/common/config-types.ts
@@ -15,6 +15,7 @@ export interface CommentsConfig {
   max_shown_comments?: number;
   theme?: Theme;
   page_title?: string;
+  node?: string;
 }
 
 export interface LastCommentsConfig {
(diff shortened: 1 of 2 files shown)


Reeywhaar:
> This also introduces the additional possibility of having multiple sets of comments on one page (as long as different URL values are provided) – I want to use remark on a site for online consultation, with users able to comment on multiple sections of a proposal.

Unfortunately it will not. Remark looks for global variable `remark_config` so you cannot declare it twice, thus you can't have remark instances on different ids.

PR doesn't solve any problem beside naming convention, though it revealed that Remark can't have more than one instance per page, and to solve this initialization script should be changed (not sure will it be breaking change or not yet)

PR will not break anything, so @umputun if you think its fine then merge it. One thing it misses is changes in readme regarding this option.

PR #431, opened 2019-09-08
Change url in comments, decorate approach

akosourov:
This is still wip. It shows possible decorate approach for import  reader. 
https://github.com/umputun/remark/issues/412#issuecomment-526948747


backend/app/migrator/migrator.go
```diff
@@ -24,6 +24,12 @@ type Exporter interface {
 	Export(w io.Writer, siteID string) (int, error)
 }
 
+// Mapper defines interface to convert data in import procedure
+type Mapper interface {
+	URL(url string) string
+	LoadRules(reader io.Reader) error
```
umputun: Presence of this method implicitly makes Mapper statefull and not thread safe. Practically it is not a big problem as our import/export controllers handling concurrency by itself with its own locking, however I can imagine some use cases in the future this could become a problem. The minimal thing we should do is at least mark (comment) all "unsafe" mapper function as "not thread safe", but eliminating it will be even beter. One way I can think about is to modify LoadRules to return the new Mapper struct

backend/app/migrator/mapper.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+	"io"
+	"io/ioutil"
+	"strings"
+)
+
+// UrlMapper implements Mapper interface
+type UrlMapper struct {
+	rules       map[string]string
+	prefixRules map[string]string
```
umputun: do we really need separate prefix rules? If rules handles both exact match and some form of regex I would thing they could be combined

backend/app/migrator/mapper_test.go
```diff
@@ -0,0 +1,41 @@
+package migrator
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUrlMapper(t *testing.T) {
```
umputun: I'd expect this test to be a table test checking many differnt inputs. Probably it could check all posible use case, including bad inputs

backend/app/rest/api/rest_test.go
```diff
@@ -277,7 +277,7 @@ func TestRest_parseError(t *testing.T) {
 }
 
 func startupT(t *testing.T) (ts *httptest.Server, srv *Rest, teardown func()) {
-	log.Setup(log.CallerFile, log.CallerFunc, log.Msec, log.LevelBraces)
+	log.Setup(log.CallerFile, log.CallerFunc, log.Msec, log.LevelBraces, log.Debug)
```
umputun: such a change affects all the tests and makes them all verbose. I'd suggest doing it in your test directly and not modifying output of everything else

backend/app/rest/api/rest_test.go
```diff
@@ -330,7 +330,8 @@ func startupT(t *testing.T) (ts *httptest.Server, srv *Rest, teardown func()) {
 			WordPressImporter: &migrator.WordPress{DataStore: dataStore},
 			NativeImporter:    &migrator.Native{DataStore: dataStore},
 			NativeExporter:    &migrator.Native{DataStore: dataStore},
-			Cache:             &cache.Nop{},
+			Mapper:            &migrator.UrlMapper{},
+			Cache:             memCache,
```
umputun: Is there a reason for adding memCache here instead of nop cache? It adds another element into the picture. I'm not saying this is the bad thing as rest test some kind of integration resting here, but what was your motivation? Is anything you tested didn't work without it?

umputun:
Generally I like this PR. It goes the right direction, easy to read and easy to reason about.

backend/app/rest/api/rest_test.go
umputun: ok, makes sense

backend/app/rest/api/rest_test.go
umputun: initially, I had it with log.Debug but it made such enormous build logs I turned it off

backend/app/migrator/mapper_test.go
```diff
@@ -0,0 +1,76 @@
+package migrator
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUrlMapper_URL(t *testing.T) {
```
umputun: There is one more typical patter we want to be covered with test - change protocol only, i.e. http://something to https://something. Probably it's working fine, but would be nice to add a test case for this

backend/app/cmd/server.go
```diff
@@ -305,6 +305,7 @@ func (s *ServerCommand) newServerApp() (*serverApp, error) {
 		DisqusImporter:    &migrator.Disqus{DataStore: dataService},
 		WordPressImporter: &migrator.WordPress{DataStore: dataService},
 		NativeExporter:    &migrator.Native{DataStore: dataService},
+		NewMapperFn:       migrator.NewUrlMapper,
```
umputun: Passing NewMapper as a func is a nice idea. The only confusing part (at least for me) is the name. Instead of NewMapperFn I'd suggest something less exotic, like `UrlMapperMaker` wich may communicate the intent better

backend/app/migrator/migrator.go
```diff
@@ -24,6 +24,16 @@ type Exporter interface {
...
 
+// Mapper defines interface to convert data in import procedure
+type Mapper interface {
+	URL(url string) string
+}
+
+// NewMapper defines function that reads rules from reader and
+// returns new Mapper with loaded rules. If rules are not valid
+// it returns error.
+type NewMapper func(reader io.Reader) (Mapper, error)
```
umputun: Same as in another comment - naming it MapperMaker looks nicer to me. However I would be fine if you decide to keep it, just a suggestion.

akosourov:
I've finished and fixed things what you noted. I've tested remap procedure by both command and api call running remark container which serves single siteID.
I want to note about prefix search. Do we want ability to merge comments from several ulrs to single? 
For example if remark keeps several urls (posts) 
```
https://site.com/p/1
https://site.com/p/2
```
and you want to merge them to single `https://site.com/merged/` current realisation of prefix search won't help. I mean if you set rules like
```
https://site.com/p/* https://site.com/merged/
```
the result will be 
```
https://site.com/merged/1
https://site.com/merged/2
```

umputun:
>  Do we want ability to merge comments from several ulrs to single?

I think we can live without it. The practical use case for prefix mapping I had in mind are:

- change `http://` to `https://`
- change domain, i.e `www.blah.com` to `blah.com`

Anything outside of those can be handled by user-provided mapping file for concrete urls

akosourov:
I've fixed naming a little and after it drone build failed (don't see problems in that changes)

umputun:
looks like a github-related issue- `fatal: The remote end hung up unexpectedly`

umputun:
LGTM

PR #434, opened 2019-09-20
Use GitHub actions as ci

Smolevich:
- Add support Github Actions (copy build docker image from `.travis.yml`)
- For correct work with coveralls.io in secrets should setted `COVERALLS_TOKEN`

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -1,6 +1,9 @@
 name: Github Actions CI
 
-on: [push]
+on:
+  push:
+    branches:
+    tags:
 
 jobs:
   build:
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -19,7 +19,7 @@ jobs:
       run: |
         docker build \
         --build-arg CI=$CI \
-        --build-arg COVERALLS_TOKEN=$COVERALLS_TOKEN \
+        --build-arg COVERALLS_TOKEN=$COVERALLS_TOKEN 
         .
       env:
         COVERALLS_TOKEN: ${{secrets.COVERALLS_TOKEN}}
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -9,8 +9,12 @@ jobs:
     - uses: actions/checkout@v1
     - name: Docker version
       run: docker version
-    - name: show env
-      run: echo $GITHUB_BASE_REF
+    - name: Debug
+      run: |
+        [[ "$TRACE" ]] && env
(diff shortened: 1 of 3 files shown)


.github/workflows/blank.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Docker version
+      run: docker version
+    - name: Debug if needed
+      run: |
+        [[ "$DEBUG" ]] && env
```
umputun: What does it do? Is it something for actions to report more details? What value this supposed to have true/false? Also setting DEBUG flag as a secret feels odd and very unusual

.github/workflows/blank.yml
```diff
@@ -0,0 +1,27 @@
...
+      env:
+        DEBUG: ${{secrets.DEBUG}}
+    - name: Build image
+      run: |
+        export GIT_BRANCH="${GITHUB_REF/refs\/heads\//}"
```
umputun: I'm not sure what this GIT_BRANCH  for, but will it work for tags?

Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM umputun/baseimage:buildgo-latest as build-backend
 
 ARG COVERALLS_TOKEN
 ARG CI
+ARG GIT_BRANCH
```
umputun: same question. Is it needed for coverals or smth else?

.github/workflows/blank.yml
Smolevich: it can be useful for showing environment variables of Github Actions. This secret should any value . But I can delete this step if he seems confused

.github/workflows/blank.yml
Smolevich: https://github.com/mattn/goveralls/blob/master/gitinfo.go#L78 this case. I check about using tag

Dockerfile
Smolevich: For coveralls

.github/workflows/blank.yml
umputun: that's fine to keep it. can be useful

umputun:
I have merged it but looks like DEBUG is not that optional

wuvye-201909-20135304-yhpbj

Smolevich:
@umputun, I prepare fix

umputun:
hmm, smth still confusing coveralls. For build from actions it shows 80% in some places, i.e. https://coveralls.io/jobs/53503811 The same build from travis looks correct https://coveralls.io/builds/25839972

PR #435, opened 2019-09-20
Fix problem with DEBUG variable

Smolevich:
- Fix problem with optional environment variable `DEBUG`
- Add case for tags and branches


Smolevich:
@umputun, PR is ready

PR #439, opened 2019-09-29
Update deps 2019 09

Reeywhaar:
Updated frontend deps.

#427 can be closed


umputun:
buld seem [to fail](https://github.com/umputun/remark/pull/439/checks?check_run_id=240750039#step:5:441) in linter

image

Reeywhaar:
yep, fixing it, new rules appeared :-)

Issue #70, opened 2018-06-04
Add twitter oauth provider


anatolym:
I'm gonna add Twitter OAuth provider to the project. This issue is for tracking progress.
(following the Umputun's comment https://github.com/umputun/remark/pull/64#issuecomment-394486570)

umputun:
@anatolym - I have tried to add twitter once, but it wasn't as direct as I hopped. Don't recall details, but I think they don't even support oauth2 with 3-legs. All examples I have seen use oauth1 for twitter. If this is the case for real and not my misunderstanding - integrating Twitter as another provider can introduce some substantial changes to `provider.go`. If you can figure a nice way to do it - go ahead, will be great to have twitter login, but if too much code and too much trouble we can wait till twitter add oauth2 ;)

anatolym:
@umputun, it seems Twitter gave up on implementing OAuth2 protocol. They heave some weird [3-legged authorization](https://developer.twitter.com/en/docs/basics/authentication/overview/3-legged-oauth) which just doesn't make sense.
So current `provider.go` abstraction won't work and we have to implement oauth1 flow for Twitter API.

Also this discussion with OAuth2 feature request https://twittercommunity.com/t/oauth-2-0-support/253 started 7 years ago and was closed with no explanation by Twitter stuff last December.
Do we want twitter login that much?

umputun:
yeah, this is what I thought. Kill it, we will be just fine without twitter auth.

umputun:
There is a thing in the same area of auth - avatar proxy. Right now it's keeping the original image as-is and then serves it to users from `AvatarStrore`. It works, but the solution is not optimized. You can try to make it better by resizing images to the desired size and keep those tiny pics only instead of full originals. I think this resizing can happen synchronously at the time of login. In case this procedure is heavy for large images (I hope it is not that heavy) you may need some workes to do it in separate goroutines.

umputun:
with https://github.com/go-pkgz/auth/releases/tag/v0.8.0 in place, we can finally have twitter auth

umputun:
@Reeywhaar - I have added twitter auth provider, but probably you have to do it on your side as well.

umputun:
@Reeywhaar - pinging. Twitter support in place on my side, but need it to be added to your list of providers

```
GET http://admin:password@127.0.0.1:8080/api/v1/config?site=remark

HTTP/1.1 200 OK
App-Name: remark42
App-Version: local
Author: umputun
[...]

PR #442, opened 2019-10-08
add twitter to the list of providers on frontend #70

umputun:


umputun:
@Reeywhaar - if you have a minute pls approve. As you know, not really my area of expertise, but based on other providers I have added twitter the same way.

PR #443, opened 2019-10-09
Fix docs

smaant:
Looks like commands and options have be changed a bit.
Also, mongo storage seems to be no longer available out of the box.


umputun:
Thx

Issue #445, opened 2019-10-13
"bolt" image storage type isn't supported


smaant:
[Readme](https://github.com/umputun/remark/blob/master/README.md#parameters) and [source code](https://github.com/umputun/remark/blob/master/backend/app/cmd/server.go#L121) mention `bolt` image storage type. But looks like it isn't supported atm (https://github.com/umputun/remark/blob/master/backend/app/cmd/server.go#L485).

umputun:
yeah, looks like I never implemented this storage for images. Should be easy to do, I'll mark it as "good-first-issue" and hopefully, someone will help with this.

mdjhd:
Would like to give it a try

mdjhd:
First time trying to contribute to open source, any remarks and suggestions are very much appreciated

Vdimir:
I'll try to implement.  I'll notify if I give up in few days.

smaant:
oops, didn't see that ^ PR and just submitted mine https://github.com/umputun/remark/pull/450

PR #446, opened 2019-10-13
"bolt"-image-storage-type-isn't-supported-#445

mdjhd:
added "bolt" image storage type to server.go

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -482,6 +482,20 @@ func (s *ServerCommand) makeAvatarStore() (avatar.Store, error) {
 
 func (s *ServerCommand) makePicturesStore() (*image.Service, error) {
 	switch s.Image.Type {
+	case "bolt":
+		if err := makeDirs(path.Dir(s.Image.Bolt.File)); err != nil {
+			return nil, err
+		}
+		return &image.Service{
+			Store: &image.FileSystem{
+				Location:   s.Image.Bolt.File,
+				MaxSize:    s.Image.MaxSize,
+				MaxHeight:  s.Image.ResizeHeight,
+				MaxWidth:   s.Image.ResizeWidth,
+			},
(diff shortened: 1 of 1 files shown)


umputun:
Thx for trying to deal with this, however, the PR doesn't address the issue #445. The goal is not to support `bolt` parameter as an alias to `fs` but implement the actual `image.Bolt` store.

This store should be added as `bolt_store.go` inside of [backend/app/store/image](https://github.com/umputun/remark/tree/master/backend/app/store/image) and implement [Store interface](https://github.com/umputun/remark/blob/master/backend/app/store/image/image.go#L31) for boltdb.

There is a similar store in [auth packages](https://github.com/go-pkgz/auth/blob/master/avatar/bolt.go) for avatars, which may give you some ideas.

mdjhd:
So just to clarify is it supposed to provide the same way of storage as for avatars? Does that mean that code for avatars in auth can be reused with some modification or am I missing something? Also thank you for giving hints. I am trying to see if I am up to the task.

umputun:
It supposed to store images to the key-value store (boltdb) instead of the file system. It has some (many) common parts with the avatar store, but they not the same.

I'd suggest reading first how `fs_store` does it and then adopt the same logic to the new bolt store. fs_store code is an easy read, nothing tricky is going on here. The only part which can be confusing is two locations - staging and final. The idea behind this separation - we don't want to keep images not linked to some comment. Users may upload such images by starting to write a comment, drop the image in and never save the comment. This is why the image first stored in the staging area and moved to the final (permanent) area on commit only and removed from the staging on sime timeout. All of this handled outside of the store itself by `image.go` but the store has to be aware of this in a few places: `Load` should try both permanent and staging, `Save` puts to staging only and `Commit` moves from staging to permanent.

mdjhd:
Tried to get familiar with boltdb reading docs and all but no progress. I get the logic but can't get my head around working with boltdb so I will be closing this pull request. :(

Issue #399, opened 2019-07-30
Extend image proxy to keep local copy of pictures


umputun:
Currently, we have an optional proxy to prevent http/https mixed content. However, often, users post images hosted on some external place and, sometimes, those pictures stored temporarily. To avoid missing pictures, we can extend image proxy functionality or even bridge it to the local image store.

_This is going to be optional mode, off by default._

smaant:
Added PR https://github.com/umputun/remark/pull/447

paskal:
`IMAGE_PROXY_CACHE_EXTERNAL` is the parameter responsible for that, resolving this issue.

PR #447, opened 2019-10-14
Preserving external images (issue #399)

smaant:
Added preserving external images into a local storage.

## Configuration
Added `preserve-external-images` parameter (`PRESERVE_EXTERNAL_IMAGES` envvar) which is `false` by default.

## Implementation details
Images Preserver acts as a `CommentConverter` and utilizes default image storage as it configured by Remark admin. Images Preserver extracts all external images (sourced outside of `REMARK_URL`), downloads and stores them into local storage, after that image URL is replaced with a local version. If at any moment of downloading/storing image any problems occur (for example image url returns 404, or image has unsupported format, or too big, etc) then original image url will be kept.

## Considered and rejected
1. Configuring it as an Image Proxy mode (e.g. `IMG_PROXY=[none,proxy,store]`). For some reason, even if implementation is somewhat close to Image Proxy, it feels that from a user (Remark admin) perspective Image Proxy and Image Preserving are two different things. Plus after storing an image Image Preserver no longer acts as a proxy.
2. Configuring it as an Image Store option. Doesn't feel like an Image Store kind of property, more like an engine behavior in general.

## Future improvements
1. Given that previously all stored images were directly uploaded by users, it kind of reduced chances of storing multiple duplicates. With Images Preserving enabled, people constantly referring to the same memes/gifs/etc might fill local store with duplicates pretty quickly. Some kind of deduplication might help.


smaant:
Looks like deprecation warning from `golangci-lint` brakes json for GolangCI

smaant:
Option is renamed to `timeout` https://github.com/golangci/golangci-lint/pull/793

umputun:
Thx for the PR.

My primary concern is this - " rejected: Configuring it as an Image Proxy mode" but to me, it sounds like the right way to do it. The current proxy already does most of the work of mapping picture urls, converting comment and serving back different (proxied) content on the fly. Preserving images on the FS (or other stores) is a natural extension of such a proxy. i.e. some form of cached proxy. Btw, if we treat stored images as persistent cache, we can limit the size of the cache (number of stored images or total volume on the disk) with LFU or LRU.

Besides, it feels like rest-level functionality and not the engine-level. I think with less code, it can be done less intrusively with an additional `proxy.Image` property and a few, relatively minor changes in `Convert` and `Handler` function. After Convert got the list of images we can store them with the original url encoded somehow into the file name, and Handle could serve from those files instead of reading from the remote URL. By doing this, we won't have images segmented by the user name, and it will allow native deduplication.

One potential disadvantage with this simplified logic - we won't be able to utilize submit+commit technic used in `image.Service` and won't be able to drop unlinked images from the storage. However, your implementation doesn't use it either (I think you use Service just as a Store), and this issue can be addressed as a part of LFU logic some day.

Another potentially promising solution is to combine `image.Store` part with image proxy solution described above. I.e. inject `image.Store` directly into `proxy.Image` and discard uder_id completely by passing in something like "system_images" as a user name.

smaant:
It might be a long answer, apologize in advance :)

I think there're two key things to clarify:
1. What kind of guaranties this "local copy" should provide? I assumed they should be treated just like uploaded pictures therefore stored permanently. I may probably argue that cache approach won't solve stated problem ("images hosted on some external place and, sometimes, those pictures stored temporarily") because it's very much possible that Remark's cache for a particular image will be purged around the same time as image disappears from external place.
2. When you say that this feature sounds very much like an extension for a proxy functionality, do you mean from implementation perspective only or from configuration pov also? In my opinion they should be configured separately because they solve two completely different problems. From a user (remark admin) perspective:
   * Proxy — "I want to make sure user isn't getting warnings about insecure connection when external image is coming from HTTP site" (and proxy is just one of the possible solutions for this problem);
   * Preserving images — "I want to make sure comments are preserved in original state even when they refer to external resources".

Besides these two things a few more comments:
> After Convert got the list of images we can store them with the original url encoded somehow into the file name

Imho pretty dangerous approach, urls are mutable, tomorrow the same url may refer to a completely different picture but remark's storage/cache will continue serving old one.

> However, your implementation doesn't use it (submit+commit technic) either

In fact it does use it. Because I put images into a permanent storage and replace urls with `/api/v1/pictures/...` all of them will be committed during comment posting. And cleanup is just an infinite loop which removed everything from staging older than TTL (no extra actions from me are required).

> discard uder_id completely by passing in something like "system_images" as a user name

Yeah, I thought about it too. One thing that stopped me, eventually some remark installation will get an actual user called "system_images" :) Who knows what other features remark will have then, for example you may add returning all images for a user, imagine surprise of this "system_images" user :)

Let me know what you think, after all it's your project and I'm totally fine changing implementation in a way you see it.

umputun:
> It might be a long answer, apologize in advance :)

No worries, mine will be even longer ;)

**TLDR: up to you with a couple of extras**


>  it's very much possible that Remark's cache for a particular image will be purged around the same time as image disappears from external place.

Sure. I mentioned LRU/LFU type of cleanup as a hypothetical "possible someday" kind of extension. Anything unlimited may run out of space someday, but let's not care about it now, like I ignored this problem for images directly uploaded to remark42 with a hope to always have big enough disk to keep them forever.

> When you say that this feature sounds very much like an extension for a proxy functionality, do you mean from implementation perspective only or from configuration pov also?

Both. I don't see the ability to serve the response from the stored file as something disqualifying the thing to be a proxy. One of the most popular proxies is squid, and this is exactly what it does. Currently, the proxy we have in remark42 limited to provide http->https "forwarding" but this forwarding is for the same external assets, and I don't see anything conceptually different between serving request by calling `http.Get` to load a picture or some `store.Load` func reading the picture from FS. I still treat this thing as some sort of cached proxy. We agreed to cache forever, but the retention period doesn't make much difference technically or conceptually. 

> Imho pretty dangerous approach, urls are mutable, tomorrow the same url may refer to a completely different picture but remark's storage/cache will continue serving old one.

Well, yes and no. I mean, this is a typical tradeoff between size and accuracy. We may address it in different ways without killing deduplication completely, for example, checking the size of the remote image after some TTL and including size into the encoded filename. But I don't think we care about this too much, at least for now, because currently, we including images exactly the same way by URL. However, I don't have any fundamental problem with saving images under user_id if you think it may help somehow.

> In fact it does use it. Because I put images into a permanent storage and replace urls with 
> /api/v1/pictures/... all of them will be committed during comment posting

Ok, agree.

>  eventually some remark installation will get an actual user called "system_images"

No, this is not possible with user_id because ids formed as <provider>_sha, i.e. `github_123456779`. Even if someday we add a provider `system` we won't be able to make sha what looks like "images" ;)

> Let me know what you think, after all it's your project and I'm totally fine changing implementation in a way you see it.

I don't have a major technical problem with your current implementation. It is fine and clear code, decently covered. I didn't like the change of converter interface initially but can live with it. To me, passing user_id felt too artificial, but I guess, in some cases such ability can be handy. 

Handling imgPreserver as a convertor was the most controversial idea here (from my pov), and the main reason why I started this discussion. However, I can accept it and, maybe, may begin to like it one day. It still feels like the wrong level to me, but it won't wake me up in the middle of the night screaming and crying ;)

The bottom line - I don't want you to spend too much time redoing this. If you have some bright ideas on how to keep the most of it but address the "wrong level" issue - it will be nice. If not - I'll gratefully accept the current implementation.

_Some additional things to consider:_

- I don't see how you handle the lack of stored file. I think currently it has no way to extract the original url. Maybe adding encoded orig url to `resImgURL` (as a query param) may help here. Not a big deal, but it means if the user changed his mind and decided not to serve already preserved images, it will be impossible to do. 
- both user-uploaded images, as well as proxied images, will be stored side by side with no way to realize which is what. Not an immediate problem, but can be someday as some big site will ask to remove all images remark42 downloaded from them. Smth to think about.

smaant:
Haha, thank you very much for a very detailed response! I actually think that "proxy" idea is growing on me, I'll see what I can do with it. I'm going to spend 5 hours on a plane this weekend, may just as well do something useful :)

backend/app/rest/proxy/image.go
```diff
@@ -105,36 +143,73 @@ func (p Image) Handler(w http.ResponseWriter, r *http.Request) {
...
+// download an image. Returns a Reader which has to be closed by a caller
+func (p Image) downloadImage(ctx context.Context, imgURL string) (io.ReadCloser, error) {
+	log.Printf("[DEBUG] downloading image %s", imgURL)
+
+	timeout := 60 * time.Second // default
+	if p.Timeout > 0 {
+		timeout = p.Timeout
+	}
+
+	ctx, _ = context.WithTimeout(ctx, timeout)
```
golangcibot: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (from `govet`)

backend/app/rest/proxy/image.go
```diff
@@ -53,48 +99,40 @@ func (p Image) Handler(w http.ResponseWriter, r *http.Request) {
...
-		rest.SendErrorJSON(w, r, http.StatusBadRequest, err, "can't get image "+string(src), rest.ErrAssetNotFound)
+		rest.SendErrorJSON(w, r, http.StatusBadRequest, err, "can't parse image url "+imgURL, rest.ErrAssetNotFound)
 		return
 	}
-	defer func() {
-		if e := resp.Body.Close(); e != nil {
-			log.Printf("[WARN] can't close body, %s", e)
+	if p.CacheExternal {
+		imgReader, _, err = p.ImageService.Load(imgID)
+		if err != nil {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

umputun:
so, you had 2 months to think about it ;) any conclusions?

smaant:
@umputun doh, I'm so sorry for disappearing! Yes, I actually pushed my changes implementing caching in proxy back in October (see the last commit https://github.com/umputun/remark/pull/447/commits/e4db45abb5d79148e5d9bb9dfd2f92065fa9d599), now need only rebase it to master and add/update tests (and probably add support into BoltDB storage). I think I'll try to do it during Christmas holidays, would it work?

umputun:
sure. somehow I missed that push completely

smaant:
Yeah, it's my fault probably, I was planning to finish everything quickly so didn't add any comment after pushing. Do you want to take a look at the implementation now (I don't expect anything to change while rebasing/adding tests) or would prefer to review finished version?

umputun:
let's do it on rebased version. I don't recall what even changed in store/image package and caused the conflict, so it's hard to tell how transparent this merge is going to be

backend/app/rest/proxy/image_test.go
```diff
@@ -92,8 +95,69 @@ func TestImage_Routes(t *testing.T) {
...
+	ts := httptest.NewServer(http.HandlerFunc(img.Handler))
+	defer ts.Close()
+	httpSrv := imgHTTPServer(t)
+	defer httpSrv.Close()
+
+	imgURL := httpSrv.URL + "/image/img1.png"
+	encodedImgURL := base64.URLEncoding.EncodeToString([]byte(imgURL))
+
+	imageStore.On("Load", mock.Anything).Once().Return(nil, int64(0), nil)
+	imageStore.On("SaveWithID", mock.Anything, mock.Anything).Once().Run(func(args mock.Arguments) { ioutil.ReadAll(args.Get(1).(io.Reader)) }).Return("", nil)
```
golangcibot: Error return value of `ioutil.ReadAll` is not checked (from `errcheck`)

smaant:
Cleaned up everything, added tests and rebased to master. Tested without frontned (with both FS and BoldDB storages) so it's all good for review. Will do end-to-end testing with frontend tomorrow.

smaant:
linter complains about using `sha1`, do we need to care given that it's not being used for anything security related? I didn't use `sha26` or higher to keep keys (filenames) getting too long an potentially hitting some limits.

umputun:
sha1 is fine for our uses and `//nolint` will be appropriate here

smaant:
Tested with frontend, all related functions seems to be working properly (posting/viewing comments with and without proxy/caching enabled).

A few notes:
 - config related to enabling http->https proxy has changed, I didn't add backward compatibility or any warnings if previous option is detected. It might make sense to do it
 - in order to guaranty that image ID doesn't change (which happened in case of resizing and FS image store) I removed storing images extensions. `content-type` for cached images is `image/*` which seems to be accepted by browsers without any problems.
 - as I mentioned in code (https://github.com/umputun/remark/pull/447/files#diff-4a38ba5920fb68398b49ccea32190da2R88-R91) turning off both proxy and caching might have unexpected for users effect, not sure if we need to do anything about it right now

README.md
```diff
@@ -184,6 +184,7 @@ _this is the recommended way to run remark42_
 | restricted-words        | RESTRICTED_WORDS        |                          | words banned in comments (can use `*`), _multi_ |
 | edit-time               | EDIT_TIME               | `5m`                     | edit window                                     |
 | read-age                | READONLY_AGE            |                          | read-only age of comments, days                 |
+| img-proxy               | IMG_PROXY               | `false`                  | enable http->https proxy for images (*deprecated*, please use image-proxy.http2https |
```
paskal: I don't know if deprecated options should be exposed in documentation (they probably should with notification in which version they will be removed?), but I think that if you truly want to deprecate some function, show a warning in the log so the user would know them need to change something in their configuration.

umputun: > I don't know if deprecated options should be exposed in documentation

I'd suggest adding another MD table "deprecated parameters" collapsed by default

> show a warning in the log so the user would know they need to change the configuration.

agree

paskal: I'd suggest showing warnings in one version after deprecation (1.5.0 for now), don't use variable and still show information on how to fix it on the second version after deprecation (1.6.0 for us) and truly remove the option in next version (1.7.0 if we are starting with 1.5.0).

This will require some "check for deprecated options" code on our side and some effort on renewing all this stuff before new releases but would be very graceful for the users. That is the behavior I would expect from the mature project, and that I believe similar to how kubernetes handles deprecations.

paskal: @smaant my changes are merged, you can move deprecated option to [appropriate section](https://github.com/umputun/remark#deprecated) in the Readme and write deprecated option handling in [HandleDeprecatedFlags](https://github.com/umputun/remark/blob/d0ddd8aa80615c20dad5ef3ad30589ac44f3d820/backend/app/cmd/server.go#L271) function.

backend/app/store/image/image.go
```diff
@@ -140,22 +141,22 @@ func (s *Service) Close() {
 // biggest side (width or height) preserving aspect ratio.
 // Returns original data if resizing is not needed or failed.
 // If resized the result will be for png format and ok flag will be true.
```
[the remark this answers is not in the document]
smaant: good catch, done

backend/app/rest/proxy/image.go
```diff
@@ -14,36 +19,77 @@ import (
...
-	return p.replace(commentHTML, imgs)
+	return commentHTML
 }
 
 // Handler returns http handler respond to proxied request
 func (p Image) Handler(w http.ResponseWriter, r *http.Request) {
-
-	if !p.Enabled {
-		http.Error(w, "proxy disabled", http.StatusNotImplemented)
+	if !p.HTTP2HTTPS && !p.CacheExternal {
```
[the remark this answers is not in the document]
smaant: yeah, that might be a solution. I don't think it's critical to fix it right now, how about I create an issue to have it fixed later in order to avoid holding this PR any longer?

backend/app/cmd/server.go
```diff
@@ -294,6 +300,10 @@ func (s *ServerCommand) HandleDeprecatedFlags() (result []DeprecatedFlag) {
 		s.SMTP.TimeOut = s.Auth.Email.TimeOut
 		result = append(result, DeprecatedFlag{Old: "auth.email.timeout", New: "smtp.timeout", RemoveVersion: "1.7.0"})
 	}
+	if s.LegacyImageProxy != false && s.ImageProxy.HTTP2HTTPS == false {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `s.LegacyImageProxy` (from `gosimple`)

README.md
smaant: great, thank you, I've rebased and added `img-proxy` into this function

backend/app/cmd/server.go
```diff
@@ -294,6 +300,10 @@ func (s *ServerCommand) HandleDeprecatedFlags() (result []DeprecatedFlag) {
 		s.SMTP.TimeOut = s.Auth.Email.TimeOut
 		result = append(result, DeprecatedFlag{Old: "auth.email.timeout", New: "smtp.timeout", RemoveVersion: "1.7.0"})
 	}
+	if s.LegacyImageProxy && s.ImageProxy.HTTP2HTTPS == false {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `!s.ImageProxy.HTTP2HTTPS` (from `gosimple`)

backend/app/rest/proxy/image.go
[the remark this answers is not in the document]
umputun: agree

.golangci.yml
```diff
@@ -0,0 +1,2 @@
+service:
```
paskal: Why this file was introduced? We have .golangci.yml in backend directory, this was not supposed to be introduced.

README.md
```diff
@@ -178,7 +178,8 @@ _this is the recommended way to run remark42_
 | restricted-words        | RESTRICTED_WORDS        |                          | words banned in comments (can use `*`), _multi_ |
 | edit-time               | EDIT_TIME               | `5m`                     | edit window                                     |
 | read-age                | READONLY_AGE            |                          | read-only age of comments, days                 |
-| img-proxy               | IMG_PROXY               | `false`                  | enable http->https proxy for images             |
+| image-proxy.http2https  |   | `false`                  | enable http->https proxy for images             |≈
```
paskal: Small typos, fixed in #570.

backend/app/store/image/fs_store.go
```diff
@@ -36,35 +36,39 @@ type FileSystem struct {
...
 	}
 
-	log.Printf("[DEBUG] file %s saved for image %s, size=%d", dst, fileName, len(data))
+	log.Printf("[DEBUG] file %s saved for image %s, size=%d", dst, id, len(data))
 	return id, nil
 }
 
+// Save data from reader for given file name to local FS, staging directory. Returns id as user/uuid
+// Files partitioned across multiple subdirectories and the final path includes part, i.e. /location/user1/03/123-4567
+func (f *FileSystem) Save(fileName string, userID string, r io.Reader) (id string, err error) {
```
paskal: Named return parameters `(id string, err error)` increase the clarity of signature but not used properly inside the code, I suggest swapping `id = path.Join(userID, guid())` with something like `tempID`, and renaming `finalID` to `id`.

backend/app/store/image/fs_store.go
```diff
@@ -36,35 +36,39 @@ type FileSystem struct {
...
-		dst = f.location(f.Staging, id)
-	}
 
 	if err = os.MkdirAll(path.Dir(dst), 0700); err != nil {
 		return "", errors.Wrap(err, "can't make image directory")
 	}
 
 	if err = ioutil.WriteFile(dst, data, 0600); err != nil {
-		return "", errors.Wrapf(err, "can't write image file %s", dst)
+		return "", errors.Wrapf(err, "can't write file")
```
paskal: Does the wrapped error have information about destination already?

backend/app/store/image/bolt_store.go
```diff
@@ -60,16 +60,14 @@ func NewBoltStorage(fileName string, maxSize int, maxHeight int, maxWidth int, o
 	}, nil
 }
 
-// Save data from reader to staging bucket in DB
-func (b *Bolt) Save(fileName string, userID string, r io.Reader) (id string, err error) {
+// SaveWithID saves data from reader with given id
+func (b *Bolt) SaveWithID(id string, r io.Reader) (string, error) {
```
paskal: I haven't checked, but aren't these changes should have their counterpart in `_example/memory_store`?

Similar changes for me required them.

.golangci.yml
umputun: I had the same concern and almost deleted the file as a dup, but I believe this was addes not for our linter but for the external one run by GolangCI

paskal: I suggest adding a comment about this file purpose then.

backend/app/store/image/bolt_store.go
umputun: As far as I recall I have never introduced remote to image.store, and as of now it has no way to be customized. Adding remote the same way we do for both store.admin and store.engine is a good thing to have, but should be handled in a separate tiicket/pr.

.golangci.yml
umputun: if this file is needed for real I'd rather move the one from the `backend` to root.

backend/app/store/image/bolt_store.go
paskal: @umputun could you please create a PR about that? I don't have enough context to do that.

paskal: #572 created, resolving this thread.

.golangci.yml
paskal: @smaant would you be so kind as to clarify the purpose of this file?

backend/app/rest/proxy/image.go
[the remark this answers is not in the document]
paskal: @smaant could you please create an issue for that?

backend/app/rest/proxy/image.go
```diff
@@ -105,36 +145,78 @@ func (p Image) Handler(w http.ResponseWriter, r *http.Request) {
...
+		if e != nil {
+			return errors.Wrapf(e, "failed to make request for %s", imgURL)
 		}
+		resp, e = client.Do(req.WithContext(ctx))
+		return e
 	})
-	return result, nil
-}
+	if err != nil {
+		log.Print(err.Error())
```
paskal: I think error should be printed with `[ERROR]` prefix or something like that, and some explanation where this error comes from.

backend/app/store/image/fs_store.go
paskal: Fixed in #581

backend/app/rest/proxy/image.go
paskal: Fixed in #580.

backend/app/store/image/fs_store.go
paskal: Fixed in #580.

backend/app/rest/proxy/image.go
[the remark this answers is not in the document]
paskal: Resolving in favor of #614.

.golangci.yml
paskal: Resolving in favor of #615.

PR #449, opened 2019-10-19
Add bolt image storage

Vdimir:
Add bolt image store support #445 

Implementation stores staging in separate bucket and copy data to main bucket on commit.
Data from staged not deleted immediately, it would be removed on cleanup.

All comments will be very appreciated.

### DIFF
--- a/backend/app/store/image/bolt_store.go
+++ b/backend/app/store/image/bolt_store.go
@@ -71,7 +71,9 @@ func (b *Bolt) Save(fileName string, userID string, r io.Reader) (id string, err
 			return errors.Wrapf(err, "can't put to bucket with %s", id)
 		}
 		tsBuf := &bytes.Buffer{}
-		binary.Write(tsBuf, binary.LittleEndian, time.Now().UnixNano())
+		if err = binary.Write(tsBuf, binary.LittleEndian, time.Now().UnixNano()); err != nil {
+			return errors.Wrapf(err, "can't serialize timestamp for %s", id)
+		}
 		if err = tx.Bucket([]byte(insertTimeBktName)).Put([]byte(id), tsBuf.Bytes()); err != nil {
 			return errors.Wrapf(err, "can't put to bucket with %s", id)
 		}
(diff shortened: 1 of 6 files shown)


backend/app/store/image/bolt_store.go
```diff
@@ -0,0 +1,152 @@
...
+	data, _ = resize(data, b.MaxWidth, b.MaxHeight)
+
+	id = path.Join(userID, guid())
+
+	err = b.db.Update(func(tx *bolt.Tx) error {
+		if err = tx.Bucket([]byte(imagesBktName)).Put([]byte(id), data); err != nil {
+			return errors.Wrapf(err, "can't put to bucket with %s", id)
+		}
+		tsBuf := &bytes.Buffer{}
+		binary.Write(tsBuf, binary.LittleEndian, time.Now().UnixNano())
```
golangcibot: Error return value of `binary.Write` is not checked (from `errcheck`)

backend/app/store/image/bolt_store.go
```diff
@@ -0,0 +1,154 @@
...
+	"io/ioutil"
+	"path"
+	"time"
+
+	bolt "github.com/coreos/bbolt"
+	log "github.com/go-pkgz/lgr"
+	"github.com/pkg/errors"
+)
+
+type Bolt struct {
```
umputun: None of public struct or public methods commented. I wonder how linter missed it, will check. However all exposed things should be command like:

// Bolt implements the store for images keeping data in 3 buckets ....

backend/app/store/image/bolt_store.go
```diff
@@ -0,0 +1,154 @@
...
+	}, nil
+}
+
+func (b *Bolt) Save(fileName string, userID string, r io.Reader) (id string, err error) {
+	data, err := readAndValidateImage(r, b.MaxSize)
+	if err != nil {
+		return "", errors.Wrapf(err, "can't load image %s", fileName)
+	}
+
+	data, _ = resize(data, b.MaxWidth, b.MaxHeight)
```
umputun: is there a reason why resized status ignores? FS store does this:

```
	if resized { // resized also converted to png
		id = strings.TrimSuffix(id, filepath.Ext(id)) + ".png"
		dst = f.location(f.Staging, id)
	}

```

Vdimir: FS store uses `resized` flag only to keep file extension consistent with actual type. Bolt do not store information about extension and I don't found reasons to use flag.

umputun: makes sense

umputun:
LGTM

PR #438, opened 2019-09-25
Email notifications

paskal:
This PR is enabling email notifications, resolving #265.
Things to do:
- [x] Implement email storage
- [x] Implement email notifications sending
- [x] Implement API for setting and deleting email
- [x] Change API to set email with receiving a verification email first
- [x] Send verification email

### Storage
Stores emails in sub-bucket in `userDetailsBucketName` top-level bucket. Currently supports only the email, but can be extended to store any kind of data in the future.

This sub-bucket can store any string data. To use it for storing lists, let's say a list of subscriptions, we would need to store them as a comma-separated string.

### Notifications sending
Notification is sent in case the user has an email set, the new comment is directly under the user's comment and the author of the new comment is another user.

Subscription per-thread right now is not supported, entering user email subscribes him to every response to his or her comments.

### API
API allow authorized user to set and delete the email address for himself, setting email address requires verification token which is sent in an email message to the specified address.

PR #469, opened 2019-11-06
implement user details storage

paskal:

It allows storing UserDetail in sub-bucket in userDetailsBucketName top-level bucket. Currently supports only the email, but can be extended to store any kind of data in the future.


backend/app/store/engine/bolt.go
```diff
@@ -617,6 +645,80 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
 	return bkt, nil
 }
 
+// checkUserDetail returns requested userDetail
+func (b *BoltDB) checkUserDetail(req UserDetailRequest) (val string) {
```
umputun: name: this is not really checkBlah but rather getBlah function

backend/app/store/engine/bolt.go
```diff
@@ -617,6 +645,80 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
...
+	key := req.UserID
+	if key == "" {
+		return "" // return nothing in case UserID is not set
+	}
+
+	bdb, err := b.db(req.Locator.SiteID)
+	if err != nil {
```
umputun: what the reason to eat errors here? I get why checkSometing may do it, but get should return the real error and not to mask it with empty string result

backend/app/store/engine/bolt.go
```diff
@@ -617,6 +645,80 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
...
+		return ""
+	}
+
+	_ = bdb.View(func(tx *bolt.Tx) error {
```
umputun: the same as above, error should be returned

backend/app/store/engine/bolt.go
```diff
@@ -617,6 +645,80 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
...
+		var bucket *bolt.Bucket
+		if bucket, err = b.userDetailsBucket(tx, req.Detail); err != nil {
+			return err
+		}
+		v := bucket.Get([]byte(key))
+		val = string(v)
+		return nil
```
umputun: assigning val here is confusing, better to return string(v) directly

backend/app/store/engine/bolt.go
```diff
@@ -617,6 +645,80 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
...
+	})
+
+	return val
+}
+
+// checkUserDetail sets requested userDetail
+func (b *BoltDB) setUserDetail(req UserDetailRequest) (res string, err error) {
+	key := req.UserID
+	if key == "" {
+		return "", errors.New("UserID is not set")
```
umputun: in all other places errors are not Camel, but lower case. Should be consitent

backend/app/store/engine/bolt.go
```diff
@@ -617,6 +645,80 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
...
+			}
+			res = req.Update
+		}
+		return nil
+	})
+
+	return res, err
+}
+
+func (b *BoltDB) userDetailsBucket(tx *bolt.Tx, userDetail UserDetail) (bkt *bolt.Bucket, err error) {
```
umputun: function name as well as parameter confusing. Took me time to figure how userDetail in 3 places mean 3 different things in one function. 

This gets a bucket for subset of userDetails and if you can't figure better name you need at lest to comment what it does

backend/app/store/engine/bolt.go
```diff
@@ -16,12 +16,13 @@ import (
...
 
 // BoltDB implements store.Interface, represents multiple sites with multiplexing to different bolt dbs. Thread safe.
-// there are 5 types of top-level buckets:
+// there are 6 types of top-level buckets:
 //  - comments for post in "posts" top-level bucket. Each url (post) makes its own bucket and each k:v pair is commentID:comment
 //  - history of all comments. They all in a single "last" bucket (per site) and key is defined by ref struct as ts+commentID
 //    value is not full comment but a reference combined from post-url+commentID
 //  - user to comment references in "users" bucket. It used to get comments for user. Key is userID and value
 //    is a nested bucket named userID with kv as ts:reference
+//  - users details in "user_details" top-level bucket. Each detail makes its own bucket and each k:v pair is userID:value
```
umputun: generally aproach of limiting user detail value to a singe "thing" (email in this example) and support sub-buckets for each one feels odd and hard to maintain. I'd rather have a single, top-level bucket with userID as a key and struct (serilaized json for example) as the value. This way everyting can be much easier to change/extend. This is epseically valid concern for migrator support

backend/app/store/engine/engine.go
```diff
@@ -73,11 +74,12 @@ const (
 	FlagFalse  FlagStatus = -1
 )
 
-// Enum of all flags
+// Enum of all flags and details
 const (
 	ReadOnly = Flag("readonly")
 	Verified = Flag("verified")
 	Blocked  = Flag("blocked")
+	Email    = UserDetail("email")
```
umputun: looks like an alien here. You are preparing to have different flavours of UserDetails, so don't mix itto another family but make it's own const group (with one for now)

paskal:
@umputun everything is fixed, the only part left to be done is migrator.
I believe existing won't change after migrator will be written and would be grateful for review of the current code.

paskal:
Migrator is implemented, the only thing left is to:
- [ ] Cover migrator-related code with tests

backend/app/store/engine/remote_test.go
```diff
@@ -173,6 +204,6 @@ func testServer(t *testing.T, req, resp string) *httptest.Server {
 		require.NoError(t, err)
 		assert.Equal(t, req, string(body))
 		t.Logf("req: %s", string(body))
-		fmt.Fprintf(w, resp)
+		_, _ = fmt.Fprintf(w, resp)
```
golangcibot: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (from `staticcheck`)

paskal:
I've removed `engine.ListDetails` interface (and embedded it into `engine.UserDetail`) and enabled user's details deletion for `deleteme` requests.

The only thing missing now is `store` interfaces that will be used within `rest_private`, but I'd prefer to write at the same time as API access to this data. I consider this done, please review.

paskal:
@umputun I've again thoughtfully gone trough this code and after the last cleanup of comments and clarifying constants names see no further points of improvements, please review.

backend/app/store/service/service.go
```diff
@@ -725,6 +740,12 @@ func (s *DataStore) SetMetas(siteID string, umetas []UserMetaData, pmetas []Post
 		if um.Verified {
 			errs = multierror.Append(errs, s.SetVerified(siteID, um.ID, true))
 		}
+		// this code doesn't delete user details in case they are not set in import but present in DB already
+		if um.Details != (engine.UserDetailEntry{}) && um.Details.Email != "" {
```
umputun: (engine.UserDetailEntry{}) ? why () here?

backend/app/store/service/service_test.go
```diff
@@ -738,12 +752,12 @@ func TestService_SetMetas(t *testing.T) {
 	// two comments for https://radio-t.com
 	b := DataStore{Engine: prepStoreEngine(t), EditDuration: 100 * time.Millisecond,
 		AdminStore: admin.NewStaticKeyStore("secret 123")}
-	umetas := []UserMetaData{}
-	pmetas := []PostMetaData{}
+	var umetas []UserMetaData
```
umputun: minor: I guess your IDE suggested this change, because i see no real tecnnical reason to do it. This is style question and I'd suggest not to change tt in fragments you modify unless you have a good reason

backend/app/store/engine/bolt.go
```diff
@@ -617,6 +646,140 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
 	return bkt, nil
 }
 
+// getUserDetail returns UserDetailEntry with requested userDetail (omitting other details)
+func (b *BoltDB) getUserDetail(req UserDetailRequest) (result []UserDetailEntry, err error) {
```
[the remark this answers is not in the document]
paskal: To provide a single interface both for getting single UserDetail and listing all user details for the whole site.

Previously it was two different functions with two signatures (`UserDetail` and `ListUserDetails`) but if I understood you correctly you were against extending the `engine` interface without absolute necessity and I've [merged them](https://github.com/umputun/remark/pull/469/commits/05fed5730ebe464cb0cf9f20e1441ee90831a602#diff-ce8df428a3e2494038b311335b9e8c95L29). As it's a list now, `UserDetailEntry` started having `UserID` duplicated in to prevent returning map (by  `UserID`) instead of list for `listUserDetails`.

The previous version of `UserDetail` returned just `string` (as it's a result for a single entry) and if we want to split single detail handling and list again new one would be returning single `UserDetailEntry`. Please let me know if that's necessary.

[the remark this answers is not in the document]
umputun: > To provide a single interface both for getting single UserDetail and listing all user details for the whole site.

It makes sense, just not as obvious as I read this particular method. Probably adding a comment explaining why list returned will be enough

backend/app/store/service/service.go
```diff
@@ -698,6 +699,20 @@ func (s *DataStore) Metas(siteID string) (umetas []UserMetaData, pmetas []PostMe
 		m[v] = val
 	}
 
+	// process users details
+	usersDetails, err := s.Engine.UserDetail(engine.UserDetailRequest{Locator: store.Locator{SiteID: siteID}, Detail: engine.AllUserDetails})
```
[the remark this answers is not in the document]
paskal: There we get all the site's details. Previously it was `map[string]UserDetailEntry` per `UserID` and I was under impression that we agreed to replace it with a simple slice and duplicate `UserID` inside each entry as it is clearer than having a map. I like this version more than the map one, should we leave it like this or switch to some third invariant?

backend/app/store/service/service.go
paskal: Checking against straight `engine.UserDetailEntry{}` produces syntax error, it surprised me as well. Apparently, that's the way of checking against empty (but not nil) structure. [Here is what I found on it now](https://stackoverflow.com/a/28447372/961092).

backend/app/store/service/service.go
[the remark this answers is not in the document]
umputun: nm, my bad. I missed this was a part of Metas call

backend/app/store/service/service_test.go
paskal: You are absolutely right, I've cleaned up all these changes in last few commits but missed this one.

backend/app/store/service/service.go
umputun: hmm, looks ugly to me. Maybe just check `um.Details.UserID != ""`  ?

backend/app/store/engine/bolt.go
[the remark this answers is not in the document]
paskal: Addressed by eccc90596c53cc0da95af5ffb47c17a8d7d2cb55, please let me know if that's clear enough.

backend/app/store/engine/bolt.go
```diff
@@ -617,6 +646,140 @@ func (b *BoltDB) flagBucket(tx *bolt.Tx, flag Flag) (bkt *bolt.Bucket, err error
...
+			}
+		}
+		return nil
+	})
+
+	return result, err
+}
+
+// setUserDetail sets requested userDetail
+func (b *BoltDB) setUserDetail(req UserDetailRequest) (result []UserDetailEntry, err error) {
```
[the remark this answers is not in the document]
paskal: Addressed by eccc90596c53cc0da95af5ffb47c17a8d7d2cb55.

backend/app/store/engine/bolt.go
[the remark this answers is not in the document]
umputun: LGTM

backend/app/store/engine/bolt_test.go
```diff
@@ -620,6 +620,51 @@ func TestBolt_FlagListBlocked(t *testing.T) {
 
 }
 
+func TestBoltDB_UserDetail(t *testing.T) {
```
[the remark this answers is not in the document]
paskal: Addressed by a8f29803aae878987f84364e8e5b668d89c46ef0, please see if it's better now.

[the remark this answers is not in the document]
umputun: yeah, a little bit. It still bothers me, but I can live with it.

PR #470, opened 2019-11-06
User details custom storage implementation example

paskal:

It implements a custom storage example for UserDetails, which is supposed to be added to remark in #469. Same as mentioned PR, it supports storing any strings but not slices of strings.


paskal:
Merged into #469, resolving this one.

PR #471, opened 2019-11-06
Email sender implementation

paskal:

It adds Email structure to the notify module, which allows sending emails but is not used or initialized at this moment. Email object initialization and retrieval of user's emails logic from storage will be done in separate PR, this PR only adds generic email sender.


backend/app/notify/email.go
```diff
@@ -0,0 +1,380 @@
...
+		log.Printf("[WARN] can't close smtp connection, %v", err)
+	}
+	return &res, err
+}
+
+// Send email from request to request.parentUserEmail if it's set, otherwise do nothing and return nil, thread safe
+// do not returns sending error, except following cases:
+// 1. (likely impossible) template execution error from email message creation from request
+// 2. message dropped without sending in case of closed ctx
+func (e *Email) Send(ctx context.Context, req request) error {
```
umputun: I don't think this is a good idea to make the buffered email clinet to be so much aware about higher levels. To me it looks like all it suppose to do is to accept the set of arguments in terms of email operation, i.e. no request but emails, from, to and so on. You are mixing two levels here - low-level email sender (with auto flash and buffering) and some higher level email service, transating requests and dealing with the sender.

backend/app/notify/email.go
```diff
@@ -0,0 +1,380 @@
...
+	if req.parentUserEmail == "" || req.parent.User == req.comment.User {
+		// don't send anything if there is no email to send information to
+		// or if user replied to his own comment
+		return nil
+	}
+	// start auto flush once, as this is the first moment we see the context from caller
+	e.once.Do(func() {
```
umputun: just a general point: running goroutine from once could make testing harder. If there is higher level (service mentioned above) start of the gorotine is better to delegate to that level.

backend/app/notify/email.go
umputun: note: I'm not saying this is really must-have separation. The code here is simple enought and may work as is, just my initial impression.

backend/app/notify/email.go
```diff
@@ -0,0 +1,380 @@
...
+	case <-ctx.Done():
+		return errors.Errorf("sending message to %q aborted due to canceled context", req.parentUserEmail)
+	}
+}
+
+// SendVerification sends email verification
+// do not returns sending error, except following cases:
+// 1. (likely impossible) template execution error from verify message template execution
+// 2. message dropped without sending in case of closed ctx
+func (e *Email) SendVerification(ctx context.Context, req VerificationRequest) error {
```
umputun: again, mixing resposibilty. SendVerification is a high-level concern. Having it next to the autoFlush make me concerend a little bit.

backend/app/notify/email.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+	if len(cumulativeErrors) > 0 {
+		err = fmt.Errorf(strings.Join(cumulativeErrors, "\n"))
+	}
+	return errors.Wrapf(err, "problems with sending messages")
+}
+
+// sendEmail sends message prepared by e.buildMessage to net/smtp.Client with established connection, thread unsafe
+func (e *Email) sendEmail(smtpClient smtpClient, m emailMessage) error {
```
umputun: confused with smtpClient passed in. Don't you have one as a member already?

backend/app/notify/notify.go
```diff
@@ -29,14 +29,27 @@ type Destination interface {
 	Send(ctx context.Context, req request) error
 }
 
-// Store defines the minimal interface accessing stored comments used by notifier
+// VerificationDestination defines interface for a given destination service which supports verification sending
```
umputun: I don't think it should be part of notify at all. To me Verification is rather concern from another world, probably from service.Email

backend/app/notify/notify.go
```diff
@@ -29,14 +29,27 @@ type Destination interface {
...
 type Store interface {
 	Get(locator store.Locator, id string, user store.User) (store.Comment, error)
 }
 
 type request struct {
-	comment store.Comment
-	parent  store.Comment
+	comment         store.Comment
+	parent          store.Comment
+	parentUserEmail string
```
umputun: this one is too specific for notify and it is leaking abstraction. Notify suppose to work on more generic level of abstaction - get a pair of comments ane maybe some meta and do the job. In no place it should be dealing with email directly

backend/app/notify/notify.go
```diff
@@ -67,14 +80,15 @@ func (s *Service) Submit(comment store.Comment) {
 	if len(s.destinations) == 0 || atomic.LoadUint32(&s.closed) != 0 {
 		return
 	}
+	var email string
```
umputun: see above, it should not deal with email directly

umputun:
One more thing - if you split sender and email service it will be nice to have a simple version of the sender doing the send directly, without all buffering thing. Not as a replacement for the buffered one, but additional implementation. At the same time, you may have NoOp sender, but this is already another level of nice-to-have.

backend/app/notify/notify.go
paskal: The problem is that whoever notify. Submit calls does not have any access to datastore but should have all context for sending message already.

Another solution I can see is having CommentWitheEmail structure for sending via notify. Submit, and asking the sender to fill an email for it. Would it be better?

umputun: I meant no need to make it email-related so directly. Just add some meta struct holding all info we need in addition to the comment. i.e. Submit(req Request) and Request may include the comment, email and whatever else we need in the future

backend/app/notify/email.go
paskal: Point taken. I've tried to do that in the first place but had to abandon this idea because we need to pass context, as stated in the comment. I see it making code flow more complex, but I found no better way to do it. Resolving.

backend/app/notify/email.go
```diff
@@ -0,0 +1,380 @@
...
+	EmailParams
+	smtpClient // embedded into Email for test purposes: not set in production code, used in autoFlush initialisation
+
+	msgTmpl    *template.Template // parsed request message template
+	verifyTmpl *template.Template // parsed verification message template
+	submit     chan emailMessage  // unbuffered channel for email sending
+	once       sync.Once
+}
+
+// smtpClient interface defines subset of net/smtp used by email client
```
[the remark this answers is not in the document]
paskal: It's created per occasion to establish a new SMTP connection, send the data and close it afterward. I think it's in place and doesn't see how I could improve its usage.

backend/app/notify/email.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		return &res, errors.Wrapf(err, "can't parse message template")
+	}
+	res.verifyTmpl, err = template.New("messageFromRequest").Parse(res.VerificationTemplate)
+	if err != nil {
+		return &res, errors.Wrapf(err, "can't parse verification template")
+	}
+
+	// establish test connection
+	testSMTPClient, err := res.client()
```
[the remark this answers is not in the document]
paskal: I would rather have it here to have a timely error message in case of a problem and the distinction of temporary email server connection problems from permanent ones (which is harder/impossible with lazy call).
I'm willing to pay the cost of test connection establishment on startup for not requiring the user to send an email to discover that he messed up with the configuration. What is your opinion on this?

backend/app/notify/email.go
paskal: Addressed by 03cd00694479537723f882eaa2c3356b11ab939c, is it what you had in mind?

backend/app/notify/email.go
```diff
@@ -0,0 +1,380 @@
...
+`
+
+const defaultEmailVerificationTemplate = `Confirmation for {{.User}} {{.Email}}, site {{.Site}}
+
+Token: {{.Token}}
+`
+
+//NewEmail makes new Email object, returns it even in case of problems
+// (e.MsgTemplate parsing error or error while testing smtp connection by credentials provided in params)
+func NewEmail(params EmailParams) (*Email, error) {
```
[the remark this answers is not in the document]
paskal: I'll look into switching NewEmail to Options (however I see it would make code more complex and no benefits in it taking in consideration we have single call of this function at the moment).

paskal:
>One more thing - if you split sender and email service it will be nice to have a simple version of the sender doing the send directly, without all buffering thing. Not as a replacement for the buffered one, but additional implementation. At the same time, you may have NoOp sender, but this is already another level of nice-to-have.

Re splitting sender and email service: I don't see why we would want that at this point in time without any consumers of separate email service. My vote is against separating these before there is any demand for that.

Re NoOp sender, I can't see where and how it would be used. Would you please describe your scenario? Is it logging everything instead of actually sending it, or something else?

backend/app/notify/email.go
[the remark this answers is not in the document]
paskal: Pleases see [initialisation example](https://github.com/paskal/remark/blob/email_notifications/backend/app/cmd/server.go#L342-L358): we need to pass everything from `cmd.Server` and I wonder if functional options are really necessary here.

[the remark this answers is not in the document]
paskal: https://github.com/paskal/remark/pull/3 is an illustration of how this new Email will be used.

backend/app/notify/notify.go
```diff
@@ -35,8 +35,16 @@ type Store interface {
 }
 
 type Request struct {
```
umputun: I have no better ideas how to include the extra info you need into request, but generally it bothers me a little bit. This struct became too specialized for email support. I'm mot saying this is bad, just a feeling.

backend/app/notify/email.go
```diff
@@ -0,0 +1,374 @@
...
+	BufferSize           int           // email send buffer size
+	FlushDuration        time.Duration // maximum time after which email will me sent, 30s by default
+}
+
+// Email implements notify.Destination for email
+type Email struct {
+	EmailParams
+	smtpClient // initialized only on sending, closed afterwards
+
+	msgTmpl    *template.Template // parsed request message template
```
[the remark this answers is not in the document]
paskal: `template.New` returns it and I thought it would be strange to convert it. I have no strong opinion here, please let me know if I need to unfold it from the pointer.

I can instead of storing it compile it every time during sending the message, but it would introduce unneeded computations and would delay the template compilation error, I would prefer to leave everything in the current state.

backend/app/notify/notify.go
paskal: Addressed by a2281931edaa333d884c720dab736d051715df29, now it is still ugly but at least have a decent description of what is expected and what purpose Request does serve.

Please see if it that's better than was before.

backend/app/notify/email.go
```diff
@@ -0,0 +1,374 @@
...
+
+// EmailParams contain settings for email set up
+type EmailParams struct {
+	Host                 string        // SMTP host
+	Port                 int           // SMTP port
+	TLS                  bool          // TLS auth
+	From                 string        // From email field
+	Username             string        // user name
+	Password             string        // password
+	TimeOut              time.Duration // TLS connection timeout
```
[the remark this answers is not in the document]
paskal: Good catch as it is actually TCP connection timeout. Addressed in 12b6298b1ba065a1fec3261159d5cf7ee47814af.

backend/app/notify/email.go
```diff
@@ -0,0 +1,374 @@
...
+	case <-ctx.Done():
+		return errors.Errorf("sending message to %q aborted due to canceled context", msg.to)
+	}
+}
+
+// autoFlush flushes all in-fly records in case of:
+// 1. buffer of size e.BufferSize + 1 is filled
+// 2. there are no new messages for e.FlushDuration and buffer is not empty
+// 3. ctx is closed
+func (e *Email) autoFlush(ctx context.Context) {
```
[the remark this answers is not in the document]
paskal: 014bdebe4962f3d3193b4986b3a8336417057ae0 addresses this issue, please advice if that's clear enough now.

backend/app/notify/email.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	if err := e.smtpClient.Rcpt(m.to); err != nil {
+		return errors.Wrapf(err, "bad to address %q", m.to)
+	}
+
+	writer, err := e.smtpClient.Data()
+	if err != nil {
+		return errors.Wrap(err, "can't make email writer")
+	}
+
```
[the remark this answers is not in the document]
paskal: Done in 014bdebe4962f3d3193b4986b3a8336417057ae0.

backend/app/notify/email.go
```diff
@@ -0,0 +1,374 @@
...
+	})
+	if err != nil {
+		return "", errors.Wrapf(err, "error executing template to build message from request")
+	}
+	return e.buildMessage(subject, msg.String(), to, "text/html"), nil
+}
+
+//buildVerificationMessage generates verification email message based on given input
+func (e *Email) buildVerificationMessage(user, address, token, site string) (string, error) {
+	subject := "Email verification"
```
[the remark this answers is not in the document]
paskal: I've added customization of the VerificationSubject in 014bdebe4962f3d3193b4986b3a8336417057ae0.

There is comment notification subject in the function above (and it's constructed based on the available request information), should we have ability to change it as well? It's not changable for telegram as well at this moment.

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,372 @@
...
+)
+
+func TestEmailNew(t *testing.T) {
+	var testSet = map[int]struct {
+		template bool
+		err      bool
+		errText  string
+		params   EmailParams
+	}{
+		1: {template: true, err: true},
```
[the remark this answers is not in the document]
paskal: I will fix it by the morning, adding the `name` field indeed would be sufficient. Thanks for pointing to the `t.Run()`, I've never seen it used before and it's unused in this project as well at the moment.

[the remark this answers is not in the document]
paskal: Fixed in c54edfa6563717e0392cf604fdfb27281b73c263.

paskal:
I (supposedly) fixed all the problems you mentioned but found that the code is broken after I carelessly changed the code after your remark instead of justifying its existence.
It is fixed in https://github.com/paskal/remark/pull/4, please review before reviewing this PR.

backend/app/notify/email.go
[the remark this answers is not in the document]
umputun: > template.New returns it and I thought it would be strange to convert it.

makes sense

backend/app/notify/email.go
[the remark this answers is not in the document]
umputun: fine with me

backend/app/notify/notify.go
umputun: well, it doesn't change much, but extra comments won't hurt for sure

paskal:
Last commit improves situation a lot, but on the other side introduce so many changes it's better to look at a whole module once again, my apologies.

backend/app/notify/email.go
```diff
@@ -0,0 +1,411 @@
...
+				log.Printf("[WARN] notification email(s) send failed, %s", err)
+			}
+			return
+		}
+	}
+}
+
+// sendMessages sends messages to server in a new connection, closing the connection after finishing.
+// Thread safe.
+func (e Email) sendMessages(ctx context.Context, messages []emailMessage) error {
```
golangcibot: copylocks: sendMessages passes lock by value: github.com/umputun/remark/backend/app/notify.Email contains sync.Once contains sync.Mutex (from `govet`)

backend/app/notify/email.go
```diff
@@ -0,0 +1,411 @@
...
+	Mail(string) error
+	Auth(smtp.Auth) error
+	Rcpt(string) error
+	Data() (io.WriteCloser, error)
+	Quit() error
+	Close() error
+}
+
+// smtpMaker interface defines function for creating new smtpClients
+type smtpMaker interface {
```
umputun: minor: a single-method interface better to be in sync with named, i.e. for blahMaker the mathod better be named Make. If you want to keep it create maybe just call it smtpCreator

backend/app/notify/email.go
umputun: I'm not sure how this expected to work from ctx point of view. It passes ctx once to the goroutine and ignores for any other call

backend/app/notify/email.go
umputun: yeah, it is a real one. Making pointer receiver is the right thing to do here

backend/app/notify/email.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+// String representation of Email object
+func (e *Email) String() string {
+	return fmt.Sprintf("email: from %q using '%s'@'%s':%d", e.From, e.Username, e.Host, e.Port)
+}
+
+// Create establish SMTP connection with server using credentials in smtpClientWithMaker.SmtpParams
+// and returns pointer to it. Thread safe.
+func (s emailClient) Create(params SmtpParams) (smtpClient, error) {
```
umputun: I'd make pointer receiver just for consitenncy with Email on the same file

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,394 @@
...
+				Username: "test@username",
+				Password: "test@password",
+				TimeOut:  time.Second,
+			},
+		},
+	}
+	for _, d := range testSet {
+		d := d // capture range variable
+		t.Run(d.name, func(t *testing.T) {
+			t.Parallel()
```
umputun: what Parallel for? I don't mind it generally, but see no reason to use it here

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,394 @@
...
+			rcpt: "test@example.org", request: Request{Comment: store.Comment{ID: "999"}, parent: store.Comment{User: store.User{ID: "test"}}, Email: "test@example.org"},
+			mail: "test_sender", response: strings.Repeat(emptyEmail, 9)},
+		{name: "one message: sent by timer", smtp: &fakeTestSMTP{}, amount: 1, quitCount: 0, waitForTicker: true,
+			request: Request{Comment: store.Comment{ID: "999"}, parent: store.Comment{User: store.User{ID: "test"}}, Email: "test@example.org"}},
+	}
+	for _, d := range testSet {
+		d := d // capture range variable
+		t.Run(d.name, func(t *testing.T) {
+			t.Parallel()
+			email, err := NewEmail(EmailParams{BufferSize: 3, From: "test_sender", FlushDuration: time.Millisecond * 200}, SmtpParams{})
```
umputun: this Parallel in addition to the goroutine inside makes it all look confusing

backend/app/notify/email.go
paskal: Excellent point, now `smtpClientWithCreator` contains  `smtpClientCreator`.

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,289 @@
...
+	return f.rcpt
+}
+
+func (f *fakeTestSMTP) readMail() string {
+	f.lock.RLock()
+	defer f.lock.RUnlock()
+	return f.mail
+}
+
+func (f *fakeTestSMTP) readQuitCount() int {
```
golangcibot: func `(*fakeTestSMTP).readQuitCount` is unused (from `unused`)

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,289 @@
...
+	return nopCloser{&f.buff}, nil
+}
+
+func (f *fakeTestSMTP) readRcpt() string {
+	f.lock.RLock()
+	defer f.lock.RUnlock()
+	return f.rcpt
+}
+
+func (f *fakeTestSMTP) readMail() string {
```
golangcibot: func `(*fakeTestSMTP).readMail` is unused (from `unused`)

backend/app/notify/email_test.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func (f *fakeTestSMTP) Data() (io.WriteCloser, error) {
+	if f.fail["data"] {
+		return nil, errors.New("failed to send")
+	}
+	return nopCloser{&f.buff}, nil
+}
+
+func (f *fakeTestSMTP) readRcpt() string {
```
golangcibot: func `(*fakeTestSMTP).readRcpt` is unused (from `unused`)

PR #480, opened 2019-11-23
Switch notify.Submit to receive notify.Request instead of store.Comment

paskal:
This change is necessary if we want to send more than just `store.Comment` (let's say user email or verification token) trough the `notify.Submit`.

Added as separate PR as `notify.Submit` triggers changes in many parts like `notify.Terraform` and will clutter `notify.Email` PR which I'm working on.

Change affects internal programming interface but nothing frontend-facing will be altered.


umputun:
LGTM

Issue #473, opened 2019-11-06
Allow anonymous voting 


umputun:
Currently, anonymous users can't vote. This is not a bug but expected behavior as we had no way to prevent abuse in such a case. However, with #400 in place, we may allow it.

This is going to be a new param (disabled by default) and in case of votes IP check disabled, will show a warning in logs.

_triggered by #472_

paskal:
Now there is an `anon_vote` option in `/config` endpoint and in case it's true, an anonymous user is allowed to vote.
What is missing is frontend support for it: currently, the anonymous user is always restricted from voting.

PR #501, opened 2019-12-26
Add backend support for anonymous voting

paskal:
Backend part for #473.

After merging this PR new `ANON_VOTE` option is added and propagated to `/config` endpoint, and if it's true then anonymous users can cast votes.

As mentioned in README, `VOTES_IP` should be enabled at the same time to make these new options work.

I've tested it and it works as intended, however, is unavailable from UI without appropriate frontend changes.


umputun:
Looking good, but it will be nice to have a test covering such a case.

PR #518, opened 2020-01-01
Shutdown test servers

mullakhmetov:
Сame across some unclosed test servers and fixed it. It wasn't done on purpose, is it?


paskal:
Would you be so kind as to check the same thing across all code? `httptest.NewServer` mostly created in setup functions, and I believe many calls outside of such functions might be affected.

mullakhmetov:
> Would you be so kind as to check the same thing across all code?

I've already done it. Only these three.

PR #523, opened 2020-01-04
Fix right-to-left typing in IE11/Edge

### LINKED ISSUE #484: Cursor sticks to the left when typing text (MS Edge)
Microsoft Edge 44.18362.449.0
When I type text (at podcast.umputun,com and radio-t.com) it looks like "Home" button is pressed constantly (cursor always jumps to the beginning of the line) so text appears in right-to-left way.
See https://vimeo.com/377949461

akellbl4:
This issue was closed https://github.com/developit/preact/issues/326 and now we can pass value right to textarea `value` attribute.

It should fix #484 issue.

### DIFF
--- a/frontend/app/components/input/textarea-autosize.tsx
+++ b/frontend/app/components/input/textarea-autosize.tsx
@@ -66,12 +66,6 @@ export default class TextareaAutosize extends Component<Props> {
     }
   }
   render(props: Props) {
-    return (
-      // We set text as a child of textarea and not in value property for a reason.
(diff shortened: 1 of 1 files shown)


akellbl4:
I tested it at MS Edge. I think it should be tested at MS IE11. If anyone has IE11 I will be happy If you test it.

paskal:
Ping @tabakerov, would you be so kind as to help us with the test? Instructions on how to bring web server with the dev version of code to life live there: https://github.com/umputun/remark#development

PR #524, opened 2020-01-04
[CI] Separate tests and build, frontend and backend tests

### LINKED ISSUE #521: Move frontend tests to separate CI step outside of docker
Hello, everybody!

[Currently](https://github.com/umputun/remark/blob/07a14edc3b07ba592cd2d0c78eecf1c65e25b961/.github/workflows/ci-build.yml#L67) we run frontend tests inside Docker alongside with building Go binary.

It doesn't give us anything CI-provided image can't and it makes sense to separate frontend tests and image build into separate steps.

It would be nice if someone will create a new [job](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobs) with everything frontend-related, then I would separate backend and frontend and will be able to try not running backend part if the code was not altered (and vice versa with frontend).

[setup-node](https://github.com/actions/setup-node) might be the thing we need.

paskal:
Fix #521.
Separates the CI pipeline into the build, test for backend, and test for frontend. Also, introduce conditional build, so backend won't be tested on frontend changes.

Results of the change can be observed [here](https://github.com/paskal/remark/commits/ci_test_branch) after `separate build and test CI workflow, separate frontend and backend tests` commit.

### DIFF
--- a/.github/workflows/ci-test-frontend.yml
+++ b/.github/workflows/ci-test-frontend.yml
@@ -1,0 +1,27 @@
+name: test_frontend
+
+on:
+  push:
+    branches:
+    tags:
+    paths:
+    - '.github/workflows/ci-test-frontend.yml'
+    - 'frontend/**'
+  pull_request:
+    paths:
+    - '.github/workflows/ci-test-frontend.yml'
+    - 'frontend/**'
+
+jobs:
+  frontend:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-node@v1
+    - run: npm ci
+      working-directory: ./frontend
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-test-frontend.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  frontend:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-node@v1
+    - run: npm ci
+      working-directory: ./frontend
+    - run: npx run-p lint test build
```
akellbl4: It also needs ts-check

```suggestion
    - run: npx run-p lint test build
```

Issue #496, opened 2019-12-17
UI for email subscription


umputun:
With the backend part in place (thx to @paskal) we can integrate it on the UI side. This should be something similar to email-based login, i.e:

1. Add a link "subscribe to email notification"
1. Click on this link will ask for email and show confirmation dialog expecting a token
1. Token will be sent to the user by email, the same way as we send confirmations for email auth
1. Subscribed user should be able to unsubscribe from UI

API for email notifications [documented in the readme](https://github.com/umputun/remark#email-subscription) 

@Reeywhaar - let me know if you want to handle this one, should be very similar to email auth you did before.

umputun:
related: #495 #494

paskal:
1. https://www.mailgun.com/ is the easiest way to get a server for email notifications, docker-compose configuration for it would look similar to this:
    ```
            - NOTIFY_EMAIL_HOST=smtp.eu.mailgun.org
            - NOTIFY_EMAIL_USERNAME=postmaster@example.org
            - NOTIFY_EMAIL_PASSWORD=password
            - NOTIFY_EMAIL_FROM=notify@example.org
            - NOTIFY_EMAIL_PORT=465
            - NOTIFY_EMAIL_TLS=true
    ```
1. [remark.rest](https://github.com/umputun/remark/blob/master/backend/remark.rest) file has working examples of requests to run against local docker-compose setup, after which you will have chosen email account verified and receiving notifications.
1. A note on the implementation: when users subscribing to email notifications, they will get all replies to all their comments on the site.
1. A small fix to apply with implementing this feature: email verification currently have text _Copy and paste this text into “token” field on comments page_ — it should be changed to something like _on email verification page_ when it will be know how it will be named.
https://github.com/umputun/remark/blob/d23d119d70e015d65aaf155713f615056181df96/backend/app/notify/email.go#L128
    <img width="360" alt="image" src="https://user-images.githubusercontent.com/712534/70978789-b1f24900-20b0-11ea-8761-75b6aead641b.png">

akellbl4:
@umputun I can do it. Where should be placed subscribe button?

paskal:
Another note: `/config` endpoint now has `email_notifications` boolean property which tells UI if notifications interface should be enabled.

umputun:
@akellbl4 sorry, missed your question. Thx for volunteering, appreciate it. 

This subscribe button (link?) should be around other subscriptions 

[...]

PR #528, opened 2020-01-05
Add email subscription info to User information

paskal:
Change for #496, add EmailSubscription flag only for `GET /user` endpoint response to use in UI.

Also, refactor `DataStore` methods which use only `SiteID` part of `store.Locator` to use only `SiteID` instead of the `Locator` object.

### DIFF
--- a/backend/app/store/user.go
+++ b/backend/app/store/user.go
@@ -15,14 +15,15 @@ import (
 
 // User holds user-related info
 type User struct {
-	Name     string `json:"name"`
-	ID       string `json:"id"`
-	Picture  string `json:"picture"`
-	IP       string `json:"ip,omitempty"`
-	Admin    bool   `json:"admin"`
-	Blocked  bool   `json:"block,omitempty"`
-	Verified bool   `json:"verified,omitempty"`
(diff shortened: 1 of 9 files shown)


backend/app/store/user.go
```diff
@@ -15,14 +15,15 @@ import (
...
-	Verified bool   `json:"verified,omitempty"`
-	SiteID   string `json:"site_id,omitempty"`
+	Name              string `json:"name"`
+	ID                string `json:"id"`
+	Picture           string `json:"picture"`
+	IP                string `json:"ip,omitempty"`
+	Admin             bool   `json:"admin"`
+	Blocked           bool   `json:"block,omitempty"`
+	Verified          bool   `json:"verified,omitempty"`
+	EmailSubscription bool   `json:"email_subscription,omitempty"`
```
umputun: unless I missed it, I don't see any place using or testing this new EmailSubscription flag.

paskal: https://github.com/umputun/remark/pull/528/files#diff-8c75f8c35267cecc93710ac2ec291accR211, I'll add a test.

paskal: Added in 4e00428.

PR #533, opened 2020-01-06
Add auto-changing theme for demo page

akellbl4:
Theme of UI is changing follows to OS settings.
I made just for demo page.

### DIFF
--- a/frontend/index.ejs
+++ b/frontend/index.ejs
@@ -10,6 +10,13 @@
         padding: 0;
         margin: 0;
         font-family: -system-ui, sans-serif;
+        color: #333;
+        background: #fff;
+      }
+
+      a {
+        color: #0aa;
+        text-decoration: none;
       }
 
       .container {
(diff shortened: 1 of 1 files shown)


umputun:
I'm not sure what "for developing view" is and how this PR affects things. Is it supposed to switch themes automatically on ios/macos? And why toggleTheme func was removed?

akellbl4:
@umputun as I see index.ejs uses for development purposes. Yes, it switch theme automatically for macOS, iOS and Android. toggleTheme func was moved to bottom script and now it subscribers by addEventListner.

The main thing is the first rendering uses OS theme, then you can change it by clicking on the button

umputun:
just to make sure - the goal is to demonstrate on `/web` page (i.e. https://demo.remark42.com/web/) how to implement OS theme support. And it does nothing else, i.e. not changing views of the real widgets outside of that page. Am I right?

akellbl4:
@umputun yes, this change affects only this page. As you can see the code is written only on this page.
And yes, I was wrong in the name of PR. This change is just for the demo page.
This change is convenient when you set automatic theme changing or your primary theme is dark.

Also, we can implement this feature on radio-t.com if you want :)

PR #540, opened 2020-01-07
Delete user email on /deleteme request

paskal:
$subj, discovered after I found out that /deleteme is not available for anonymous users and wanted to check if it's properly working for normal ones.


umputun:
would be nice to add this to test as well

PR #544, opened 2020-01-08
Run es-check by CLI

akellbl4:
Also, es-check was updated


akellbl4:
@Mavrin I agree, but it takes not so much time for check and I think we can leave it.

Reeywhaar:
@mavrin, reason for this check, is that babel skips transpilation for modules in node_modules. This way code with const, arrow functions can appear in final bundle and break devices on ios 8, ie10 and so on. We can’t trust webpack here.

Issue #162, opened 2018-07-12
Change vote buttons order 


alehano:
It would be more convinient if upvote arrow button will be on the right side. Upvote uses more often than downvote. Now it placed close to collapse button that sometimes cause accidentally wrong clicks, especially on mobile phone.

umputun:
I don't think swapping them will be a good idea as it will confuse existing users for sure.

Regarding accidental click on the collapse button - I don't like this position either but not sure what will be the better place.

@Guria - any bright ideas?

Guria:
@umputun I like Reddit solution: border on the left through all nested thread

alehano:
@Guria Agree, collapse button on the left side is more intuitive.

umputun:
@Guria reddit-style collapsing should be fine

umputun:
@sw-double & @Reeywhaar - making comments collapsable on the border line we just added can be useful 

re: #350

paskal:
I wonder what are the action items here after #545 is merged?

WinstonPoh:
What if the upvote button is changed to be on top of the number and downvote at the bottom of the number? That was both new and old users will not be confused?

PR #545, opened 2020-01-09
Collapse comment by thread border

akellbl4:
#162

### DIFF
--- a/frontend/app/components/thread/thread.tsx
+++ b/frontend/app/components/thread/thread.tsx
@@ -1,26 +1,18 @@
 /** @jsx createElement */
-import { createElement, RenderableProps, FunctionComponent } from 'preact';
-import { useStore } from 'react-redux';
+import { createElement, FunctionComponent } from 'preact';
+import { useSelector, useDispatch } from 'react-redux';
+import { useCallback } from 'preact/hooks';
 import b from 'bem-react-helper';
 
-import { ConnectedComment as Comment } from '@app/components/comment/connected-comment';
 import { Comment as CommentInterface } from '@app/common/types';
-import { getThreadIsCollapsed } from '@app/store/thread/getters';
+import { getHandleClickProps } from '@app/common/accessibility';
 import { StoreState } from '@app/store';
-import { InView } from '../root/in-view/in-view';
+import { setCollapse } from '@app/store/thread/actions';
+import { getThreadIsCollapsed } from '@app/store/thread/getters';
+import { InView } from '@app/components/root/in-view/in-view';
+import { ConnectedComment as Comment } from '@app/components/comment/connected-comment';
 
-const mapStateToProps = (state: StoreState, props: { id: CommentInterface['id'] }) => {
-  const comment = state.comments[props.id];
-  return {
-    comment,
-    childs: state.childComments[props.id],
-    collapsed: getThreadIsCollapsed(comment)(state),
-    isCommentsDisabled: !!state.info.read_only,
-    theme: state.theme,
-  };
-};
-
-interface OwnProps {
+interface Props {
   id: CommentInterface['id'];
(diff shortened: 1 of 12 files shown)


umputun:
thx. Not sure if expected, but the selection seems to work for the single (current?) segment only, i.e.:

image

Another thing - there is no visual indication of a collapsed tree

akellbl4:
@umputun, ok. I realize what I need to change.

umputun:
Not sure if you expected me to review, but if you do - I still don't understand what it's doing. It seems to collapse a part of the level sometime (also indicates incomplete level), and sometimes it collapses a level above. Just to make sure we are on the same page - this thing supposes to do exactly the same fold as the current [-] icon does.

Indication of minimized subtree odd as well (losing avatar and a part of the name). I'd suggest something simple here, like show the "title" in italic or different shade or smth like this

akellbl4:
All right. I made the right logic of comment collapsing and fixed collapse on the deepest level.
Now I am going to "make it sexy" and to do this code clean.

frontend/app/components/thread/thread.tsx
```diff
@@ -29,10 +21,21 @@ interface OwnProps {
...
+  const collapsed = getThreadIsCollapsed(comment)(state);
+
+  return { comment, childs, collapsed, theme };
+};
 
-function Thread(props: RenderableProps<Props>) {
-  const { collapsed, comment, childs, level, theme } = props;
+export const Thread: FunctionComponent<Props> = ({ id, level, mix, getPreview }) => {
+  const dispatch = useDispatch();
+  const { collapsed, comment, childs, theme } = useSelector(commentSelector(id));
```
Mavrin: Please use shallow equality https://react-redux.js.org/next/api/hooks#equality-comparisons-and-updates
 or use separate useSelector

umputun:
>  I would expect a rectangle or plus sign or something else which will tell me that it's something I can expand.

We have tried some other icons but found nothing decent. Open for suggestions/examples

> With the current design, I would expect that clicking an avatar would collapse the thread: it's not the case now.

Well, this functionality already occupied by the user's widget. I don't think changing the behavior of such a click based on view is a good idea.

> The deepest level of comments (or I think deeper than the deepest level) is noncollapsible.

They seem to be collapsable from the closest top level with a visual ident. I don't want to make it complicated and allow some smart collapses for deepest level. To me, it should match visual borders and if level 10 collapsed because user clicked on the deepest visual ident border (6?) - so be it.

paskal:
> We have tried some other icons but found nothing decent. Open for suggestions/examples

I don't think we should preserve user avatar, that's a confusing part. Even having the current icon without an avatar background would be better in my opinion.

>> With the current design, I would expect that clicking an avatar would collapse the thread: it's not the case now.

> Well, this functionality already occupied by the user's widget. I don't think changing the behavior of such a click based on view is a good idea.

Clicking avatar doesn't produce any action in the current version of remark, as far as I know. I am OK with keeping the current behavior, however.

> They seem to be collapsable from the closest top level with a visual ident. I don't want to make it complicated and allow some smart collapses for deepest level. To me, it should match visual borders and if level 10 collapsed because user clicked on the deepest visual ident border (6?) - so be it.

Good by me.

umputun:
> I don't think we should preserve user avatar, that's a confusing part. Even having the current icon without an avatar background would be better in my opinion.

interesting idea, worth to try

> Clicking avatar doesn't produce any action in the current version of remark, as far as I know.

yeah, you right. Probably click on avatar can do collapse

akellbl4:
> Probably click on avatar can do collapse

I think collapsing by click on avatar might be confusing because it is not a common behavior for avatar.

umputun:
>  by click on avatar might be confusing 

agree

PR #548, opened 2020-01-12
Fix position of "show more" button

akellbl4:
It was aligned by the left side.
This button visible only on mobile version.

### DIFF
--- a/frontend/app/components/root/__show-more/root__show-more.scss
+++ b/frontend/app/components/root/__show-more/root__show-more.scss
@@ -1,4 +1,5 @@
 .root__show-more {
+  display: block;
   width: 300px;
   max-width: 100%;
   margin: 20px auto;


paskal:
Could you please add visual on how it looks like before and after the change? By the way, I didn't ever saw it on a desktop because I think for desktop (on [demo](https://remark42.com/demo/), for example) it seems to load more comments automatically.

akellbl4:
Before:
Screenshot 2020-01-12 at 15 12 10
After:
Screenshot 2020-01-12 at 15 12 37

@paskal you can activate mobile view in your browser.
Safari (enable development mode before in preference): Develop -> Enter Responsive Design Mode
Chrome: View -> Developer -> Developer Tools and click on the icon with Phone and Tablet in the top left corner of Developer Tools.

paskal:
My apologies, I thought that "The button was shown only on a mobile." was undesired behavior and you wanted to start showing it now. Now it looks better than before.

Issue #494, opened 2019-12-17
Prettify email notifications HTML


paskal:
Email notifications are merged but are not yet enabled and it's necessary to adjust the new comment notification template before we will be able to enable it.

When subscibed user get reply to their comment they get email notification currently which look like this:
<img width="722" alt="image" src="https://user-images.githubusercontent.com/712534/70975902-2a560b80-20ab-11ea-879a-ef6dcd2960de.png">

It should have a user avatar, and text of the comment it replies to. Also, it might be a good idea to add a date to both comments it in the same way it's done on the site:
<img width="843" alt="image" src="https://user-images.githubusercontent.com/712534/70976092-93d61a00-20ab-11ea-9ffb-06231375d315.png">

Here is how Disqus handles this.
<img width="722" alt="image" src="https://uploads.disquscdn.com/images/7e423eee4b5f4816f36acd53c49131f4877883e2b3a2d3ed31f01f3fd95a234c.png">

The code of the template which should be changed: https://github.com/umputun/remark/blob/d23d119d70e015d65aaf155713f615056181df96/backend/app/notify/email.go#L97-L115

New reply and parent comment parameters which you can use in the template:

https://github.com/umputun/remark/blob/d23d119d70e015d65aaf155713f615056181df96/backend/app/store/comment.go#L12-L30

https://github.com/umputun/remark/blob/d23d119d70e015d65aaf155713f615056181df96/backend/app/store/user.go#L16-L26

Please pick human-readable names for variables in your template, there is no problem translating `parent.Comment.User.Picture` to `parentUserAvatar` when passing to HTML template.

StaroKep:
<img width="690" alt="Screen Shot 2019-12-29 at 01 27 35" src="https://user-images.githubusercontent.com/17280727/71550165-f64be600-29da-11ea-9d45-fe06f053e157.png">

Code:
1)https://paste.ofcode.org/kR3iDbUrHxPBJ8LfqwtSTd
2) https://paste.ofcode.org/t8ZKCFmY8KzGzhjpK66cC4

paskal:
@StaroKep really nice job! Would you be so kind as to prepare PR request with these changes?

From the code part, I would suggest only removing empty lines and making `style` single-lined. If it would be necessary, I'll move the code to a separate file instead of having it inline now.

umputun:
[...]

PR #550, opened 2020-01-14
Update email templates

akellbl4:
#494

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,9 @@ _this is the recommended way to run remark42_
 
 ##### Deprecated
 
-<details><summary>deprecated options</summary>
+<details>
+<summary>deprecated options</summary>
+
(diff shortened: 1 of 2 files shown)


akellbl4:
@umputun I moved templates from GO because it is more comfortable to edit HTML in a separate file. It can help to new contributors change templates as files against part of GO file. 
Also, the reply template has 66 lines (and I tried to collapse some lines to single) and I think it's pretty big.

What do you think? If you sure that templates should be placed in GO files I will move them back.

umputun:
> If you sure that templates should be placed in GO files I will move them back.

Currently, it is not the question of what is better aesthetically because templates from external files are not loaded (in case of the binary distribution), i.e. this change breaks things for sure. I suggest moving them back and, in case if we decide to introduce another step into our build process, we can always move them out.

akellbl4:
Okay, I got it. I will move separation in other PR ;)

PR #552, opened 2020-01-15
Fix score position and margin before blockquote

akellbl4:
*  fix score position
* add margin before blockquote if it is placed after text


akellbl4:
Text is too close to blockquote when it is placed above.  Especially, It is bad when a link is placed before it. 

<details>
<summary>Before</summary>

Screenshot 2020-01-15 at 11 34 05
</details>

<details>
<summary>After</summary>

Screenshot 2020-01-15 at 11 34 19
</details>

PR #554, opened 2020-01-16
Drop cache on release for remark.css/.js

akellbl4:
Now static files load by permanent link it may cause problems with caching.
I added dynamic query string to loaded files, it generates by a content hash of the files and flush cache only when the content of the file was updated.

<details>
<summary>Hashes in url were added</summary>

Screenshot 2020-01-16 at 09 57 49

</details>

### DIFF
--- a/frontend/iframe.html
+++ b/frontend/iframe.html
@@ -10,7 +10,6 @@
         window.close();
       }
     </script>
-    <link rel="stylesheet" href="remark.css" />
     <style>
       html,
       body {
@@ -127,6 +126,5 @@
 
       window.parent.postMessage(JSON.stringify({ inited: true }), '*');
     </script>
-    <script type="text/javascript" src="remark.js"></script>
   </body>
 </html>
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -165,7 +165,7 @@ module.exports = () => ({
       inject: false,
     }),
     new MiniCssExtractPlugin({
-      filename: '[name].css',
+      filename: '[name].css?[contenthash:6]',
     }),
     new webpack.optimize.ModuleConcatenationPlugin(),
     ...(process.env.CI
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -70,8 +70,8 @@ module.exports = () => ({
   },
   output: {
     path: publicFolder,
-    filename: `[name].js`,
-    chunkFilename: '[name].js',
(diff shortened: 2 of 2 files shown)


frontend/iframe.html
```diff
@@ -127,6 +126,5 @@
 
       window.parent.postMessage(JSON.stringify({ inited: true }), '*');
     </script>
-    <script type="text/javascript" src="remark.js"></script>
```
akellbl4: Links to files were removed because of webpack-html-plugin inject them automatically.

frontend/webpack.config.js
```diff
@@ -139,8 +139,14 @@ module.exports = () => ({
       'process.env.REMARK_NODE': JSON.stringify(NODE_ID),
       'process.env.REMARK_URL': env === 'production' ? JSON.stringify(remarkUrl) : 'window.location.origin',
     }),
+    new Html({
+      template: path.resolve(__dirname, 'iframe.html'),
```
akellbl4: Inject static files to ifreme.html

paskal:
I will investigate how to properly cache these things shortly. However with all libraries you get different name of the imported file based on its version, and that’s how CDN work: I think the approach of generating unique file name per hash is sane and we should implement it.

akellbl4:
@Mavrin I do not agree with it, because of my suggestion similar to `[name].[hash].js`.
1. `[contenthash]` better than just `[hash]`, because it does not change if the content of file is not changed.
2. If just add query string it is the same if we change the name of the file. Because queries cached by a path.

For example:
- We rolling update.
  Webpack generates paths:
    remark.css?12345
    remark.js?23456
- We rolling the next one in which we change only the CSS file.
  Webpack generates paths:
    remark.css?34568 - (cache of this file will be busted)
    remark.js?23456 - (this file stayed cached)

You don't need to change the name of a file for a flushing cache.
But if you change it just in time when we rolling update you can have 404 error because a browser already gets the old version of HTML but old static files disappeared with the previous instance and you don't have access to new.

That is because I want to make static files name permanent.

frontend/iframe.html
paskal: Would you be so kind as to move unrelated changes to separate PR so this one would be crystal-clear? From your commentary, I'm under impression that this change is unrelated to this PR title.

akellbl4: I think we won't use this changes.
Also, this changes needs to this PR, I removed static link for files, because webpack generates links with hashes and link it automatically.

Issue #565, opened 2020-01-18
Built in browser's search doesn't work


paskal:
With current "load when I see it" approach it's impossible to search on the comments page, as search on page (at least in Firefox) would show results only from a text which is visible on the user's screen.

This breaks user experience and causes all kinds of problems, for example [on this page](https://radio-t.com/p/2020/01/14/prep-685/) with news references users are reporting same news multiple times as they have no ability to check for it other than read all comments manually.

I don't know how hard would be to fix it, but I do think that it should be fixed at any cost.

umputun:
this is a good point. I didn't know our rendering is that dynamic.

umputun:
I don't think we have a good reason to render comments on-demand on a big screen. Back in days, it used to be complete render on desktops and i don't recall why UI switched to this smart render. Probably this was done by @Reeywhaar for some good reason, but I really don't recall all the details.

Reeywhaar:
Yeah, this was done because as React doesn't work good with long tree/list structures. On demo https://remark42.com/demo/ page there were noticeable lag when you click "reply", change sorting (like 20 seconds delay) and so on. So I switched to rendering of comment only if it is in viewport. Otherwise it renders dummy block (`<div></div>`).

Another good thing is that all images also render lazily.

But yeah, searching sucks.

akellbl4:
@Reeywhaar do you made any benchmarks connected with it?

Reeywhaar:
No, it was noticeable difference. You can try and disable it.

paskal:
Is it possible to render whatever the user looks at the moment, but don't lose anything rendered so far? That way sort speed and all other things should be fine, I think that's similar to how Reddit works.

Reeywhaar:
This will not solve the problem with searching, and will not gain anything.

I think that we can render dummy including comments message. As a compromise. But we'll lose lazy image loading, though it was side effect. If we need lazy loading, I think t can be made separately. Once concern is that comment's text itself can be complex structure (tables, images).

paskal:
From Reddit I won't expect search from all comments because I know it won't load past first ~50 before I reach the end of the page, however, I know that I can search through whatever is loaded so far as soon as I've seen it once.

For Remark it doesn't work like that: I see lazy loading, but after scrolling to the very bottom I still can't search the data I've seen so far. This is an expected behavior for me.

Reeywhaar:
Okay, let's try your solution.

PR #568, opened 2020-01-18
Do not render dummy for comment once it gets into viewport

Reeywhaar:
to #565

Stop rendering dummy block for comment once it appears in viewport

### DIFF
--- a/frontend/app/components/comment/comment.tsx
+++ b/frontend/app/components/comment/comment.tsx
@@ -52,6 +52,7 @@ export type Props = {
 } & Partial<typeof boundActions>;
 
 export interface State {
+  renderDummy: boolean;
   isCopied: boolean;
   editDeadline: Date | null;
   voteErrorMessage: string | null;
@@ -79,6 +80,7 @@ export class Comment extends Component<Props, State> {
     super(props);
 
     this.state = {
+      renderDummy: typeof props.inView === 'boolean' ? !props.inView : false,
       isCopied: false,
       editDeadline: null,
       voteErrorMessage: null,
@@ -114,6 +116,10 @@ export class Comment extends Component<Props, State> {
       scoreDelta: props.data.vote,
       cachedScore: props.data.score,
     };
+
+    if (props.inView) {
+      newState.renderDummy = false;
+    }
 
     // set comment edit timer
     if (props.user && props.user.id === props.data.user.id) {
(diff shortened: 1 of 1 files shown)


paskal:
After this change comments page becomes searchable. However it's still searchable after changing the sort method, I was thinking that after sort change we should re-render everything from scratch to prevent performance issues.

Reeywhaar:
Uhm, okay. I see two tropes:

first:
on sort change: we unmount whole tree (and show something like "loading..."). After remount comment state will be cleared and `renderDummy` will be `false` again.
pros:
- easy to do
cons:
- "loading..."

second:
we add `sortChangeCount` to redux and increment on each sort change. Then use `sortChangeCount` as part of thread's and comment's key. On key change, comment will be remounted.
pros:
- behaviour will be same as now
cons:
- err, need to manage count value and components keys.

Know something else?

paskal:
FYI: I am completely happy with how it works now (search on-page problem is solved) and I don't see performance downsides on 42 comments I've created.

I'm asking about the sort because you mentioned that it will be slow on a bigger dataset, and I was under impression that it's relatively cheap to have current performance benefits and at the same time fix page search problem, however from what I see at this very moment and from what I've read from your explanations, this PR seems to be eliminating performance benefits for someone who loaded the whole page. I might be completely wrong about it, it's just a guess from the information I have as someone who don't know how it works inside.

Reeywhaar:
I am not sure If I should change anything else as well. I do not know what expected behaviour for user after sort change (search will not be working again).

Btw, I see someone updated react-redux library (which I tried to avoid, because at some version there appeared lags between user actions), and now I feel every click has some lag. There is some difficulties caused by using of preact. Am I the one who notices it?

paskal:
> I do not know what expected behaviour for user after sort change (search will not be working again).

In analogy to Reddit, I would expect it to load first ~50 comments with a new sort method and everything I've seen so far to be lost for search and re-appear only after I start scrolling the page again.

akellbl4:
@Reeywhaar I can't see lags but maybe because I didn't see when it was better.

Also, I saw the discussion connected with adding redux in the project, and I think redux unnecessary for the project especially now when we can change it by `useReducer` and theme interface by CSS vars. I think it can be a theme for a new discussion about that.

I will check which place can cause problems with performance because I don't see any problems with it in my projects which operate with a bigger data storage.

Issue #560, opened 2020-01-17
Email confirmation with a direct link


umputun:
@paskal - I think back in days we have discussed ability tо confirm by direct click from the notification email directly. However, this seems to be gone and confirmation email provides token only.

Did we have a reason to drop it from the template, or just missed this part?

paskal:
The reason is just separation of frontend and backend: there is a separate endpoint on the backend to verify an email but it's called from the frontend, the same way as email auth works at the moment. I'll try to figure out the possible solution for this with frontend developers, my main concern here is to keep frontend and backend decoupled (so no redirects from the backend to frontend, for example).

umputun:
I see what you mean, but from another hand, email template is some sort of frontend (at least conceptually) it just happened to live in the backend universe. I don't see much a problem with adding the link generator aware of the final, UI link for this email. There are other places backend aware about such things already.

akellbl4:
In my opinion, we can do smth like that:
Frontend gives a link to a page to the backend, backend adds query param to a link. When a user goes by link, frontend catch token from a link and sends it to backend.

This solution has weak sides but it will be convenient for a user.

paskal:
Current behavior on reloading http://127.0.0.1:8080/web/remark.js without cache (Cmd + Shift + R on MacOS) with added debug logging:

    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:583 api.cacheControl.func2.1} cacheControl request for /web/remark.js
    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:475 api.addFileServer.func1} addFileServer request for /web/remark.js
    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:583 api.cacheControl.func2.1} cacheControl request for /web/remark.js
    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:475 api.addFileServer.func1} addFileServer request for /web/remark.js

For some reason request for a static asset is handled by all involved handlers twice. I've checked the content of the request, and it's the same for both requests.

On reload `addFileServer` request handler is not called because cacheControl return 304 answer, but it's still getting called twice for some reason:

    remark42-dev | 2020/01/18 06:36:07.041 [DEBUG] {api/rest.go:583 api.cacheControl.func2.1} cacheControl request for /web/remark.js
[...]

PR #573, opened 2020-01-19
Add subscription link support to notification email verification template

paskal:
Address #560, the backend part.

When `NOTIFY_EMAIL_SUBSCRIBE_URL` environment variable is provided, instead of embedding token to email it will be included as a link.

With `SITE=http://127.0.0.1` and `NOTIFY_EMAIL_SUBSCRIBE_URL=/subscribe?token=`, link will look like `http://127.0.0.1/subscribe?token=<subscription_token>`. The token is not escaped and I'll escape it if it will be necessary when the frontend part will be written.

After the frontend counterpart will be created, `NOTIFY_EMAIL_SUBSCRIBE_URL` default should be set to newly created URL location.

That's how verification email looks like when subscribe_url is supplied:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,6 @@ _this is the recommended way to run remark42_
 | notify.telegram.timeout | NOTIFY_TELEGRAM_TIMEOUT | `5s`                     | telegram timeout                                |
 | notify.email.fromAddress | NOTIFY_EMAIL_FROM      |                          | from email address                              |
 | notify.email.verification_subj | NOTIFY_EMAIL_VERIFICATION_SUBJ | `Email verification` | verification message subject          |
-| notify.email.subscribe-url | NOTIFY_EMAIL_SUBSCRIBE_URL | `` | URL parameters to add to SITE in order to include to subscription email instead of plain token |
 | smtp.host               | SMTP_HOST               |                          | SMTP host                                       |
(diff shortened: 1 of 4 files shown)


akellbl4:
I thought it well be smth like this
Screenshot 2020-01-20 at 16 12 49

Two methods at the same time.

umputun:
I think we should keep both methods

paskal:
Updated, please see new look below.

image

umputun:
LGTM, but we can't merge it till UI side is ready

umputun:
my bad, looks like `{{if .SubscribeURL}}` makes it safe to merge

README.md
```diff
@@ -155,6 +155,7 @@ _this is the recommended way to run remark42_
 | notify.telegram.timeout | NOTIFY_TELEGRAM_TIMEOUT | `5s`                     | telegram timeout                                |
 | notify.email.fromAddress | NOTIFY_EMAIL_FROM      |                          | from email address                              |
 | notify.email.verification_subj | NOTIFY_EMAIL_VERIFICATION_SUBJ | `Email verification` | verification message subject          |
+| notify.email.subscribe-url | NOTIFY_EMAIL_SUBSCRIBE_URL | `` | URL parameters to add to SITE in order to include to subscription email instead of plain token |
```
umputun: I know what you meant, but the wording is confusing.

Issue #538, opened 2020-01-07
Autosubmit on paste the token in inputs


akellbl4:
We have auth by verification via token and we know the length of the token.
It says that we can submit the form automatically by pasting token in input.

This feature can implement for email auth and email subscription.

umputun:
> and we know the length of the token.

Not really, the token is JWT and can have different lengths for different users. We know token structure (i.e. could be possible to match with regex) but not the length.

akellbl4:
@umputun good. If I am right, it looks like `^[A-Za-z0-9-_=]+\.[A-Za-z0-9-_=]+\.?[A-Za-z0-9-_.+/=]*$`

H1D:
Anyways we can [easily](https://stackoverflow.com/a/38552302) parse JWT token and auto submit when it is valid.

akellbl4:
@H1D thank you for the good advice ;)

PR #583, opened 2020-01-26
Autosubmit token by paste and a bit more

akellbl4:
- [x] Autosubmit in email auth #538 
- [x] Autosubmit in email subscribe #538 
- [x] Using official typings instead copied
- [x] Tests for JWT funcs
- [x] Tests for the interfaces


umputun:
```
ERROR in /srv/frontend/app/components/comment/comment.test.tsx
337
[tsl] ERROR in /srv/frontend/app/components/comment/comment.test.tsx(280,26)
338
      TS2344: Type '{ user: User | null; data: Comment; repliesCount?: number | undefined; post_info: PostInfo | null; isUserBanned?: boolean | undefined; isCommentsDisabled: boolean; editMode?: CommentMode | undefined; ... 8 more ...; uploadImage?: ((image: File) => Promise<...>) | undefined; } & Partial<...> & AllHTMLAttributes<...> ...' does not satisfy the constraint 'string | number | symbol'.
339
  Type '{ user: User | null; data: Comment; repliesCount?: number | undefined; post_info: PostInfo | null; isUserBanned?: boolean | undefined; isCommentsDisabled: boolean; editMode?: CommentMode | undefined; ... 8 more ...; uploadImage?: ((image: File) => Promise<...>) | undefined; } & Partial<...> & AllHTMLAttributes<...> ...' is not assignable to type 'symbol'.
340
npm ERR! code ELIFECYCLE
341
npm ERR! errno 2
342
npm ERR! remark-ui@0.1.0 build: `webpack --mode production && es-check es5 './public/*.js'`
343
npm ERR! Exit status 2
344
npm ERR! 
345
npm ERR! Failed at the remark-ui@0.1.0 build script.
```

akellbl4:
@umputun fixed. I just forgot to push one of the files with which I am working now.

umputun:
I have not tried it, but not really sure I get how all of this supposed to work. So, user pastes token and it got submitted by itself? If that's the case, why the Submit button even needed? And, in addition, this automatic submission feels kind of exotic and unusual to me.

akellbl4:
@umputun yep, if a user pastes the correct token it sends automatically.
"Send" button is needed not to confuse a user because without the button the form might look broken.

paskal:
There might be errors like "Too many requests, try later" and I would prefer having the submit button all the time to deal with them manually in case it's needed.

After automatic submit user end up either logged in (with auth) or with page "you are subscribed, click here to unsubscribe" (notifications) - this is the desired output of the process user started. It might be surprising to skip the "push submit button" step but the user won't feel lost in my opinion.

I have the same mechanics working for my bank 3-D Secure pin-code from SMS prompt, I recall being surprised in a good way when I found I don't need to push any buttons after I entered expected four digits. My vote is for introducing this feature in a way it was implemented in this PR.

umputun:
ok, convinced. As soon as we get approval from @Mavrin I'll merge

frontend/package.json
```diff
@@ -43,6 +44,7 @@
     "@types/redux-mock-store": "^1.0.1",
     "@typescript-eslint/eslint-plugin": "^2.5.0",
     "@typescript-eslint/parser": "^2.5.0",
+    "atob": "^2.1.2",
```
Mavrin: It looks unnecessary or use for node
```js 
const atob = (value) => Buffer.from(value, `base64`).toString();
```

PR #586, opened 2020-01-29
Add collapse thread control to comment controls

alexeyten:
Add “Collapse” button

image

### DIFF
--- a/frontend/app/components/comment/comment.test.tsx
+++ b/frontend/app/components/comment/comment.test.tsx
@@ -211,22 +211,24 @@ describe('<Comment />', () => {
 
       const controls = element.find('.comment__controls').children();
 
-      expect(controls.length).toBe(5);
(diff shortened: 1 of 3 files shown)


umputun:
having this new link doesn't add a new line, which is good. However, I'm not sure how this will look on small screens (iPhone) with logged-in user.

alexeyten:
image

There would be a problem for admin. I guess «Edit», «Delete» and countdown would wrap to next line. 
image

umputun:
looking at the examples above I have to admit - I don't like what I see. Without a good reason, it will change the position of the most important link (reply) for all users. To me, it feels like unnecessary complexity. In the end - if the user needs to collapse in v1.5 he has to discover a new way of doing it (border or this new link). I don't see the value of duplicating this functionality unless you have a good reason I'm missing.

PR #589, opened 2020-02-07
Replace blackfriday.v2 with blackfriday/v2

paskal:
Move this change outside of code-highlight to prevent unrelated changes creeping into that PR.


umputun:
good idea

Issue #32, opened 2018-05-15
Add github style syntax highlighting.


Shchvova:
For examle:

    ```java
    class Borsch {
      private static function borshch() {
        return new Borshch();
      }
    }
    ```
In github results in:

```java
class Borsch {
  private static function borshch() {
    return new Borshch();
  }
}
```

But in remark it just renders
```
class Borsch {
  private static function borshch() {
    return new Borshch();
  }
}
```

umputun:
yeah, we know and even tried to address. The problem is this - the regular md library we use seems to set ids only for colors but doesn't provide styles. External rendering engine I have tried brings way too many dependencies and conflicted with sanitizer.

Shchvova:
I tried to implement my change but [found out](https://github.com/umputun/remark/blob/5f6134b39094f586e5635e45f4f3dd61427a57fe/app/rest/api/rest.go#L160) that original markdown is not stored at all, and storage just contains rendered html. 

I'm not sure how editing supposed to work with such model either... Do you get to edit html after you entered markdown? If it has colors would it be HTML with all the styles and stuff. Anyway. My suggestion wouldn't work with current model, and I don't know if changing model breaks something.

umputun:
> original markdown is not stored at all

Why is it a problem for syntax highlighting? Maybe you tried to colorize it on the client side?

> I'm not sure how editing supposed to work with such model either...

Good point, me neither. Most likely this will require storing md as well.

Shchvova:
Yes, my whole comment was about post processing MD on client side with JS. This will allow to use custom post-processing. Basic MD rendering is done how it is now. When comments is loaded client can extend with callback and re-render the comment.
Example would be mathjax/katex. Like include math formulas. This is not needed for casual comments, but some sites may want to allot to use math notations in comments. Another - plug in advanced code highlighting.
Suggestion: everything works as it is now: except add `orig`(`md`) field to Comment. It can be used either in "Edit" functionality or in andvaced re-rendering on client side if required. Re-rendering this way can be made user customization, not even requiring to edit anything in Remark. All can be done with parent HTML/CSS/JS.

umputun:
@Shchvova see https://github.com/umputun/remark/issues/13 trying to decide which way to go

umputun:
@Reeywhaar - I have checked the current state of fenced code blocks and on my side, it seems to render it correctly

[...]

PR #590, opened 2020-02-07
Code colors (backend support)

paskal:
Backend part for #32.

In order to merge it, we need clearance from frontend developers that getting rid of `<code>` tag won't break anything and they will be able to add CSS to render this code properly.

Also, all review comments should be resolved before the merge.

### Example without the language specification

Comment text

    ```
    test text
    ```

Old render

```html
<pre><code>test text
</code></pre>
```

New render

```html
<pre class="chroma">test text
</pre>
```

### Example with the language specification

Comment text

    ```go
    func main(aa string) int {return 0}
    ```

Old render

```html
<pre><code class="language-go">func main(aa string) int {return 0}
</code></pre>
```

New render

```html
<pre class="chroma"><span class="kd">func</span> <span class="nf">main</span><span class="p">(</span><span class="nx">aa</span> <span class="kt">string</span><span class="p">)</span> <span class="kt">int</span> <span class="p">{</span><span class="k">return</span> <span class="mi">0</span><span class="p">}</span>
</pre>
```


backend/app/store/comment.go
```diff
@@ -109,7 +109,8 @@ func (c *Comment) SetDeleted(mode DeleteMode) {
 // Sanitize clean dangerous html/js from the comment
 func (c *Comment) Sanitize() {
 	p := bluemonday.UGCPolicy()
-	p.AllowAttrs("class").Matching(regexp.MustCompile("^language-[a-zA-Z0-9]+$")).OnElements("code")
+	p.AllowAttrs("class").Matching(regexp.MustCompile("^chroma$")).OnElements("pre")
+	p.AllowAttrs("class").Matching(regexp.MustCompile(".")).OnElements("span")
```
paskal: This rule is far too permissive, the proper rule should be "allow `class` property with a particular set of values in `span` inside `pre` tag". I haven't figured out the complete list of possible values so far: CSS file might be a good reference for a full list of possible classes.

backend/app/store/formatter.go
```diff
@@ -48,7 +50,9 @@ func (f *CommentFormatter) FormatText(txt string) (res string) {
 		Flags: bf.Smartypants | bf.SmartypantsFractions | bf.SmartypantsDashes | bf.SmartypantsAngledQuotes,
 	})
 
-	res = string(bf.Run([]byte(txt), bf.WithExtensions(mdExt), bf.WithRenderer(rend)))
+	extRend := bfchroma.NewRenderer(bfchroma.Extend(rend), bfchroma.ChromaOptions(html.WithClasses()))
```
paskal: Renderer has numerious [options](https://github.com/Depado/bfchroma#options), I wonder which set would work for us.

backend/app/store/comment.go
paskal: Oh, and the important part: it seems to me that it's impossible to build such a rule using `bluemonday`.

backend/app/store/formatter.go
paskal: From all I've read, current state should be enough and all the visual magic for white and dark themes should happen in CSS.

backend/app/store/comment.go
```diff
@@ -109,7 +109,14 @@ func (c *Comment) SetDeleted(mode DeleteMode) {
...
 	p := bluemonday.UGCPolicy()
-	p.AllowAttrs("class").Matching(regexp.MustCompile("^language-[a-zA-Z0-9]+$")).OnElements("code")
+	p.AllowAttrs("class").Matching(regexp.MustCompile("^chroma$")).OnElements("pre")
+	// this is list of <span> tag classes which could be produced by chroma code renderer
+	// source: https://github.com/alecthomas/chroma/blob/022b6f4fc2c4aa819aac18363c8de3f70619200b/types.go#L221-L316
+	const codeSpanClassRegex = "^(chroma|ln|lnt|hl|lntable|lntd|w|err|x|esc|k|kc" +
+		"|kd|kn|kp|kr|kt|n|na|nb|bp|nc|no|nd|ni|ne|nf|fm|py|nl|nn|nx|nt|nv|vc|vg" +
+		"|vi|vm|l|ld|s|sa|sb|sc|dl|sd|s2|se|sh|si|sx|sr|s1|ss|m|mb|mf|mh|mi|il" +
+		"|mo|o|ow|p|c|ch|cm|cp|cpf|c1|cs|g|gd|ge|gr|gh|gi|go|gp|gs|gu|gt|gl)$"
+	p.AllowAttrs("class").Matching(regexp.MustCompile(codeSpanClassRegex)).OnElements("span")
```
paskal: This rule is too permissive, the proper rule should be "allow `class` property with a particular set of values in `span` inside `pre` tag".

It seems that it's impossible to build the "inside `pre` tag" part of a rule using bluemonday.

I will test to see how important is it and if we can live with it.

Issue #5, opened 2018-05-12
Implement and document styles customization


umputun:
We should have a way to define external CSS for basic customizations - things like link colors, font sizes, and other styles.

On compose level this will be as simple as some volume mapping, like `- ./etc/styles.css:/srv/web/styles.css` 

Also, need instructions in the readme on how to do it and what we allow to customize (with an example)

Electrifying-Haze:
why can't there be just .css files inside the ./var directory, including the ones with the default themes?

umputun:
@akellbl4 & @Mavrin -it will be very nice if we finally could support custom css somehow. Pls share your brilliant ideas

ColasNahaboo:
A simple solution could be to add en env var ```CSS``` in ```docker-compose.yml``` that can have as value an additional CSS to use. The overriding will be left to the CSS natural cascading rules. The documentation should then tell which CSS selectors should stay invariant between remark42 releases.

```   - CSS=https://my-local-css-url.css```

On my apache web server in the meantime I have made this code that works currently, with the hack that I have proxied the urls ```/web/``` and ```/api/```   to my local server remark42 as it it was on the same domain (otherwise modern browsers wont let you inject a CSS link into remark42 iframe contents):

I added some code to the client-side javascript added to pages (after the line ```/* Add my custom CSS */```) that use a new js config field ```css```

It is usable now but of course it is not a general solution as it supposes the urls /web/ and /api/ are free to use on the host server.

```
<script>
  var remark_config = {
    host: "https://%HTTP_HOST%",    /* will be proxied to the remark42 server */
    site_id: '%WIKITOOLNAME%',
    components: ['embed', 'last-comments', 'counter'],
    url: '%SCRIPTURL{"view" topic="%BASEWEB%.%BASETOPIC%"}%',
    css: 'https://%HTTP_HOST%/pub/Main/SiteStyle/colas_remark42.css'
};
 (function(c) {
    for(var i = 0; i < c.length; i++){
      var d = document, s = d.createElement('script');
      s.src = remark_config.host + '/web/' +c[i] +'.js';
      s.defer = true;
      (d.head || d.body).appendChild(s);
    }
    /* Add my custom CSS */
    window.onload = function(){   
      var f = document.querySelector("div#remark42 > iframe");
      var d = f.contentDocument? f.contentDocument: f.contentWindow.document;
      var h = d.getElementsByTagName("head")[0];
      var s = document.createElement('link');
      s.type = 'text/css';
[...]

PR #599, opened 2020-02-14
Use css custom properties for colors

Mavrin:
There is POC of theme api. It can be merged safely. I suppose, it is not should affect current behavior for end user.
## What I did:
- I added stylelint rule which check that we use css custom properties for all color usage. Thank you @lahmatiy for help with plugin and [csstree](https://github.com/csstree). 
- I added `postcss-custom-properties` plugin for fallback for browsers which doesn't support css custom properties
- I replaced all static color on css custom properties

## How does it work?
You can uncomment in `index.ejs`
```js
        // __colors__: {
        //   "--color0": "red",
        // },
```
and notice that comment change color
Screen Shot 2020-02-14 at 22 29 28
I use `__colors__` because it not final implementation.
## What is next?
- We need reduce count of colors. Now we have 49 colors. I think, we can reduce that and make derived colors, for example 
```js
  const bgPanel = color(colorBody)
    .mix(color(colorContent), 0.025)
    .toString();
```
- We should define how api will looks for user who embed remark42. I think, it will be something like this:
```js
var remark_config = {
        site_id: 'remark',
        host: window.location.origin,
        url: 'https://remark42.com/demo/',
        components: ['embed', 'counter'],
        theme: { // if it is string fallback on old behavior
           mode: "dark",
           font: {
               family: "Helvetica, Arial, sans-serif"
           },
           colors: {
              color0: "green",
              ...
           }
        }
      };
```

Comments are welcome.


paskal:
I am not able to review this, but from the description, it seems like there are multiple changes included in single PR and generally I would advise to move them to separate PRs to simplify the reviewer job.

Mavrin:
@paskal Of course it can be split, but it doesn't make sense most of changed files is replace by full search and if I split this request on smaller, it can't help with review this request.

paskal:
It's up to other reviewers, but in such a situation I personally would prefer having two more PRs with 1 file changes in each and this with 63 files, rather than one with 65. Mixing multiple intentions in single PR usually makes my life as a reviewer harder, and it seems that this PR is a mix of custom properties added plus some style and linter improvements that are outside of that scope.

Mavrin:
Linter is not outside of that scope. You can consider it like tests, it verifies that I don't miss any css files witch  contains css colors.

umputun:
this is an interesting idea. The question I have – is it limited to colors only? From your example, it looks like fonts will be supported as well, correct me if wrong. What else (if any) can we support in the same way?

Mavrin:
Yes,`fonts` also will be supported. I am going add it in next pull request. Also we can add support padding, font-size, but it isn't easy. My first goal is colors and font-family.

umputun:
Another question - wouldn't be uploading a regular css more generic and easier to implement? I mean adding some API method allowing admins to upload css file with whatever user need? The method (smth like `POST /css?site-id=blah`) will save css overwrites per site-id and will be paired with `GET /css/site-id=blah` to serve it. We can add server-side logic to this GET, like returning default (empty) css in case if the admin hasn't posted any. At the same time, we could apply some modifications to the response, like enforcing "!important" suffixes (if needed, just an example), verify CSS validity, and so on.

_Am I missing something and this is not a good idea? To me, it sounds like almost no work on the UI side (well, maybe minimize and consolidate styles to make it more friendly for "rewrites") and mostly trivial change for the backend.  I have seen something similar in both matermost and rocket chat, both allow to enter css manually._

Mavrin:
Add support custom css  files , it can be next step. But it is not easy support css selector between remark42 version, and it will be hard admin user figure out what to need change in css for new released version. 
Look how it do `slack`
Screen Shot 2020-02-17 at 07 11 30

umputun:
>  But it is not easy support css selector between remark42 version, and it will be hard admin user figure out what to need change in css for new released version.

Theoretically this "migration issue" can be mitigated on the server-side. I.e. we provide several basic selectors and maintain proper mapping. Not sure I like this approach as it leaks parts of frontend logic to the backend side. 

Or even better - we define a list of "exposable selectors" and never change them. This will definitely make ui development less flexible as those selectors going to be a part of "public API". 

Either way - up to you, just an idea I wanted to share.

akellbl4:
I support the approach with adding CSS files.
We can make two options of customization "simple" and "advanced".
In first it is just a definition of custom properties. We can provide an example with a list of colors and demo page with few variants of color customization, also it can be an "online editor" for CSS.
The second way is an attempt to fix class names, I think it is possible that we will take care of it. For example, we can start from small UI elements like buttons and inputs and then go forward with bigger elements.
Yes, of course, some time selectors can be changed but I see only two reasons:
1. Refactoring - we can change complex parts of the interface and it will corrupt parts with layout stylization (I mean when a user adds styles to layout blocks). I think we should make rules of style customization and try to keep compatibility with previous versions.
2. New features - it is definitely a migration issue.

It is a long way to make it good enough for using and I think we can start to provide this feature like alfa or beta, because our "API" can change before we make final version.

P.S. Example with Slack is good but it is a standalone application and theming it is a minor feature for them. But for us, it can be a major feature because I think any comment system provides it.

Mavrin:
@umputun could you merge it?

ColasNahaboo:
I wws going to comment (prompted from https://github.com/umputun/remark42/issues/5#issuecomment-641260128 ) but I would only repeat what umputun has said.:

- Just support adding CSS files.
- define a set of documented selectors.
- if you must change them, document the changes in the release notes.

PR #601, opened 2020-02-17
Cleanup embed.ts and add limit-size checker

Mavrin:
Motivation:
embed.js file is remark42 loader and it should be smaller as possible.

In this request I removed unused pollyfills and fix IE11 issue with remove dom node. I checked, emed.js works even in IE11.

Size was changed from `25k` to `2.5к`

### DIFF
--- a/frontend/app/common/constants.config.ts
+++ b/frontend/app/common/constants.config.ts
@@ -1,0 +1,5 @@
+export const BASE_URL: string =
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ((window as any).remark_config && (window as any).remark_config.host) || process.env.REMARK_URL!;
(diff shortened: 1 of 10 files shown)


frontend/app/common/constants.config.ts
```diff
@@ -0,0 +1,5 @@
+export const BASE_URL: string =
```
akellbl4: mb it should be just "config.js"?

akellbl4:
LGTM, I just looked at code and didn't build and test it.

frontend/app/common/constants.config.ts
Mavrin: I don't know, I extracted these constant from `constants.ts` and I suppose it is constants. Anyway it can be removed in future.

Mavrin:
@umputun Could you merge it please.

PR #606, opened 2020-03-05
Added german translation

codiflow:
Ok everything seems fine just the **reply** got changed to **replay**.

Thats wrong and should be reply instead.


frontend/app/locales/de.json
```diff
@@ -1,167 +1,167 @@
...
+  "comment.unverify-user": "Möchtest du den Benutzer {userName} wirklich nicht mehr bestätigen?",
+  "comment.verified-user": "Bestätigter Benutzer",
+  "comment.verify-user": "Möchtest du den Benutzer {userName} wirklich bestätigen?",
+  "comment.vote-error": "Fehler beim Abstimmen: {voteErrorMessage}",
+  "commentForm.exceeded-size": "Die Datei {fileName} überschreitet das Größenlimit von {maxImageSize}",
+  "commentForm.input-placeholder": "Gib hier deinen Kommentar ein",
+  "commentForm.new-comment": "Neuer Kommentar",
+  "commentForm.notice-about-styling": "Hier kannst du <a>Markdown</a> verwenden",
+  "commentForm.preview": "Vorschau",
+  "commentForm.replay": "Antworten",
```
codiflow: replay should be **reply** instead

Mavrin: Please change it manually. Don't worry, it already fixed in original dictionary and it will work well.

PR #609, opened 2020-03-05
Small refactoring

akellbl4:
I made some small refactors during doing other tasks and I think this is time to put it on master :)
It's good to review by commits.

* visibility flag for toggling settings view was moved from redux to local store
* duplicated code for parsing query string was collapsed in one func and used in all places + tests
* wrapper with `node_id` was removed from jsx because this id was defined in html and was dublicated on page after render. 
* wrapper of root view was moved on upper level
* setSorting got tiny refac
* for pollyfills was added dummy types and was changed resolving of import


frontend/app/components/root/root.tsx
```diff
@@ -96,46 +94,64 @@ const messages = defineMessages({
...
     });
 
     window.addEventListener('message', this.onMessage.bind(this));
   }
 
-  logIn = async (p: AuthProvider): Promise<User | null> => {
-    const user = await this.props.logIn(p);
-    await this.props.fetchComments(this.props.sort);
+  fetchComments(sort = this.state.sort) {
+    return this.props.fetchComments(sort);
```
Mavrin: fetchComments() {
  const sort = this.state.sort;
  return this.props.fetchComments(sort);
}

frontend/app/components/root/root.tsx
```diff
@@ -96,46 +94,64 @@ const messages = defineMessages({
...
+  }
+
+  async changeSort(sort: Sorting) {
+    if (sort === this.state.sort) return;
+    this.setState({ isCommentsListLoading: true });
```
Mavrin: this.setState({ isCommentsListLoading: true,  sort}, async () => {
     await this.fetchComments().catch(() => {});
     localStorage.setItem(LS_SORT_KEY, sort);
     return { isCommentsListLoading: false };
});

frontend/app/store/post_info/actions.ts
```diff
@@ -1,40 +1,39 @@
...
+  await (read_only ? disableComments() : enableComments());
   dispatch(unsetCommentMode());
-  dispatch({
-    type: POST_INFO_SET,
-    info: { ...storeState.info, read_only: !state },
-  });
-  return !state;
+  dispatch(setPostInfo({ ...info, read_only }));
+
+  return read_only!;
```
Mavrin: return !read_only;

frontend/app/store/post_info/actions.ts
```diff
@@ -1,40 +1,39 @@
...
+  const { read_only } = info;
+
+  await (read_only ? disableComments() : enableComments());
   dispatch(unsetCommentMode());
-  dispatch({
-    type: POST_INFO_SET,
-    info: { ...storeState.info, read_only: !state },
-  });
-  return !state;
+  dispatch(setPostInfo({ ...info, read_only }));
```
Mavrin: read_only: !read_only

frontend/app/components/root/root.tsx
akellbl4: [This parameter used in this place](https://github.com/umputun/remark42/pull/609/files/c2f5bd44c3755491eb7894dd8c26f0da779745f1#diff-df6edac8c631ae459ffbd3dc621acd4bR138
)

frontend/app/store/post_info/actions.ts
akellbl4: `read_only` was changed twice in the previous version when it defined and when it passed to obj.

umputun:
looks like we have a conflict now

PR #612, opened 2020-03-08
set anonymous with admin names to blocked 

### LINKED ISSUE #605: Reject admin names for anonymous and (maybe) email auth
With those auth methods, bad guys can try to impersonate and pick one of admin's names. This is visible but may confuse users.

We should reject such names at least on the backend side for anonymous. Regarding email auth - unclear, because this could be a valid way for real admin to login

umputun:
fix #605


umputun:
note: this PR marks anonymous users trying to set admin names as blocked.

Issue #595, opened 2020-02-09
login on-demand 


umputun:
The idea suggested by @bobuk - if the user not logged in, click on reply should ask for login first and then open the usual comment box. 

_I have checked how other systems do it - reddit, github, gitlab and habr don't even support it, hacker news seems to jump to a different login page, disqus opens a popup._

akellbl4:
@bobuk we did it ;)

Issue #426, opened 2019-08-26
A few frontend improvement ideas


leem32:
I know there's no ETA for any frontend refactoring, but just wanted to add a few ideas for what in my opinion would improve remark42.

Show the comments box even if a user is not logged in.
Or at least add an option for this in the config parameters.
I believe this would greatly increase the chances of a user posting a comment.
The user could type their comment and when they click the submit button a popup would appear saying "Please login to comment".

Reduce the length of the sign in to comment line.
This line of text is too long IMO and looks messy for smaller viewports and if you need to place your comments in a narrow container.
Instead of: "Sign in to comment using Google, Yandex, Facebook or email" you could just have "Login: " followed by the social login/email icons similar to the way Disqus does.

Also the dropdown boxes for email login and username break outside the container at certain viewports or in narrow containers which also looks messy and obviously unreadable.

Those are the 3 things I would most like to see added/fixed in remark42.

Thanks :)

akellbl4:
> Show the comments box even if a user is not logged in.
I think it is not small feature. I will think about it later.

> Reduce the length of the sign in to comment line.
@umputun can I change "Sign in to comment using ..." to "Login via ..."?

> Also the dropdown boxes for email login and username break outside the container at certain viewports or in narrow containers which also looks messy and obviously unreadable.

@leem32 can you provide some screenshots or viewport sizes?

umputun:
> can I change "Sign in to comment using ..." to "Login: ..." or "Login via ..."?

sure, good idea

leem32:
Oddly I can't seem to reproduce the issue I was having with the username dropdown box.
I do notice, when I log into comments the sort by options box escapes its container which can cause the right side of the sort by options box to be cut off for narrow mobile viewports in a 1 column layout with no gutter.

(Screenshots using Chrome Dev tools mobile view.)
These screenshots show when a user is logged in the sort-by options box expands further to the right than when a user is not logged in. 

User not logged in: (this is fine)
![remark42 not logged in](https://i.ibb.co/N63wdWk/remark42-not-loggedin.png)

User logged in.
![remark42 logged in](https://i.ibb.co/rvr8wJN/remark42-loggedin.png)

akellbl4:
@leem32 it's not a problem :) This is the default browser select (not HTML markup). It can render differently in different browsers. On a mobile device, it looks completely different. 

<details>
<summary>For example on iOS</summary>

[...]

PR #616, opened 2020-03-11
On-demand auth

akellbl4:
This PR solve suggestions form #595 and #426
It is fast realization of on-demand auth.
I can have small bugs and maybe not so clean code.
I will happy with any suggestions connected with realization.

* auth block was moved to comment form
* if comments on a page are read-only auth form shows in the old place
* comment value in the form will be saved between refreshes (it is side effect form saving comment value between an unauth and auth states)
* sort was moved back to store because now it difficult to pass this value over comments tree

P.S. Auth block needs to refactor and maybe small redesign, I plan to do it in the near feature.

<details>
<summary>Screenshots</summary>

Top comment block:
Screenshot 2020-03-11 at 16 07 37

Comment block under any comment:
Screenshot 2020-03-11 at 16 07 50

Read only page:
Screenshot 2020-03-11 at 16 11 02

</details>


umputun:
looking good. So far the only thing I run into - image drag-drop (cut-paste) doesn't work in the entry box unless the user logged in. I understand this is due to the required auth for image operation, but user can be surprised. Maybe we should show some message saying "to add image pls login" or smth like this

frontend/app/components/auth-panel/auth-panel.tsx
```diff
@@ -522,7 +324,20 @@ function getSortArray(currentSort: Sorting, intl: IntlShape) {
   });
 }
 
-export const AuthPanelWithIntl = (props: PropsWithoutIntl) => {
+export default function(props: OwnProps) {
```
Mavrin: I recommend to use named function. It is better for development, you will see function name in call stack and react dom tree.

frontend/app/components/auth/auth.tsx
```diff
@@ -0,0 +1,198 @@
+/** @jsx createElement */
```
Mavrin: It will be good if you use hooks for this component.

frontend/app/components/auth-panel/auth-panel.tsx
akellbl4:
```suggestion
export default function AuthPanelConnected(props: OwnProps) {
```

frontend/app/components/auth/auth.tsx
akellbl4: I just moved the old component in a new place. It will be rewritten in the near future.

frontend/app/components/auth-panel/auth-panel.tsx
Mavrin: Yes it is much better. But I prefer to avoid default export, because  it is potential buggy approach.

umputun:
@akellbl4 - do you want me to merge it in or still work in progress?

akellbl4:
@umputun I think I should add notifications about image uploading before merge.

umputun:
>  I think I should add notification about image uploading before merge.

makes sense.

umputun:
the error message should disappear after the login
grfasmxqhw-20200312-135908

akellbl4:
@Mavrin done

Issue #572, opened 2020-01-19
Remote (rpc) implementation of image.store


umputun:
We have missed this one. It is going to be very similar to admin.remote and engine.remote, i.e. will have rpc-wrapped functions for image.Store interface

In addition, we will need a demo implementation in example.memory_store

re https://github.com/umputun/remark/pull/447#discussion_r368314575

PR #623, opened 2020-03-22
Image interface changes

paskal:
Multiple preparations before implementing #572 which I don't want to mess up into image RPC implementation PR.

- sort imports in `memory_store` for consistency
- add a missing copyright notice to `rpc.go` file
- make all `image.Store` interface functions public: otherwise, it's impossible to use this interface from another package:
  image
- make `image.Store.Load` return `[]byte` instead of `io.ReadCloser`
- regenerate mock files in remark42 backend (for `image` and `engine`)
- `go mod tidy` in `memory_store` changes to `go.mod`
- divide `memory_store` RPC server to multiple files by handlers, as new one would be added and single file feels to be too big for that


backend/app/store/image/remote_store_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func testServer(t *testing.T, req, resp string) *httptest.Server {
```
golangcibot: `testServer` is unused (from `deadcode`)

PR #626, opened 2020-03-23
Fix user actions

akellbl4:
Selections of comments from the store were fixed in user actions.
For all cases was added tests.


frontend/app/store/user/actions.ts
```diff
@@ -100,30 +98,26 @@ export const fetchHiddenUsers = (): StoreAction<void> => dispatch => {
 };
 
 export const hideUser = (user: User): StoreAction<void> => (dispatch, getState) => {
-  if (IS_STORAGE_AVAILABLE) {
-    const hiddenUsers = JSON.parse(getItem(LS_HIDDEN_USERS_KEY) || '{}');
-    hiddenUsers[user.id] = user;
-    localStorage.setItem(LS_HIDDEN_USERS_KEY, JSON.stringify(hiddenUsers));
-  }
-  dispatch({ type: USER_HIDE, user });
+  const hiddenUsers = JSON.parse(getItem(LS_HIDDEN_USERS_KEY) || '{}');
```
Mavrin: need try catch

PR #632, opened 2020-04-01
Small improvements

Smolevich:
- Use action shogo82148/actions-goveralls@v1 ([Link](https://github.com/shogo82148/actions-goveralls) on repo)
- Remove path for .golangci.yml for ci trigger


.github/workflows/ci-test-backend.yml
```diff
@@ -8,13 +8,11 @@ on:
       - '.github/workflows/ci-test-backend.yml'
       - 'backend/**'
       - '!backend/scripts/**'
-      - '.golangci.yml'
```
paskal, line 11: Why would you want to prevent tests from running on `.golangci.yml` change?

.github/workflows/ci-test-backend.yml
```diff
@@ -39,18 +33,19 @@ jobs:
...
-          export TZ="America/Chicago"
           date
           cd backend/app
           go test -mod=vendor -timeout=60s -covermode=count -coverprofile=$GITHUB_WORKSPACE/profile.cov_tmp ./...
           cat $GITHUB_WORKSPACE/profile.cov_tmp | grep -v "_mock.go" > $GITHUB_WORKSPACE/profile.cov
           cd ../_example/memory_store
           go test -race ./...
+        env:
+          TZ: "America/Chicago"
+          GOPATH: ${{ github.workspace }}
```
paskal, line 44: Why `GOPATH` is necessary here?

.github/workflows/ci-test-backend.yml
```diff
@@ -60,8 +55,11 @@ jobs:
...
-        working-directory: backend
+      - name: Send coverage
+        uses: shogo82148/actions-goveralls@v1
+        with:
+          path-to-profile: ${{ github.workspace }}/profile.cov
+          parallel: true
+          working-directory: backend
         env:
-          COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GOPATH: ${{ github.workspace }}
```
paskal, line 65: Why do we need GOPATH here? It's [marked as legacy](https://github.com/shogo82148/actions-goveralls#use-with-legacy-gopath-mode) in this action doc.

.github/workflows/ci-test-backend.yml
```diff
@@ -60,8 +55,11 @@ jobs:
         run: $GITHUB_WORKSPACE/golangci-lint run --config ${GITHUB_WORKSPACE}/backend/.golangci.yml ./...
         working-directory: backend/_example/memory_store
 
-      - name: submit coverage
-        run: $(go env GOPATH)/bin/goveralls -service="github" -coverprofile=$GITHUB_WORKSPACE/profile.cov
-        working-directory: backend
+      - name: Send coverage
```
paskal, line 58: A small thing, but I would prefer to preserve the original name of the step.

.github/workflows/ci-test-backend.yml
paskal: To clarify, this line is intentionally added there so when linter configuration is changed, it's tested by CI as well.

Smolevich: This file was removed from repo [here](https://github.com/umputun/remark42/commit/10c1aa939f5ad5c71e17e62978ca835e2dc16664), what is the reason to leave this line?

paskal: backend/** [matches](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) hidden files as well, all good then. Thank you!

.github/workflows/ci-test-backend.yml
Smolevich: Removed

.github/workflows/ci-test-backend.yml
Smolevich: Removed

.github/workflows/ci-test-backend.yml
```diff
@@ -61,7 +55,8 @@ jobs:
         working-directory: backend/_example/memory_store
 
       - name: submit coverage
```
umputun, line 57: what the reason to change it? it doesn't make code shorter and more readable but adds another dependency for no good reason.

.github/workflows/ci-test-backend.yml
Smolevich: I reverted old name

.github/workflows/ci-test-backend.yml
Smolevich: Because this step looks like as some that can be replaced on public action with configuration params.

umputun: I still think this is unnecessary dependency. In addition, it doesn't work or works differently. I don't see your builds on https://coveralls.io/github/umputun/remark42

Smolevich: Job #108898c84-PR-632.0
https://coveralls.io/jobs/60912751

umputun: which shows "PENDING COMPLETION", whatever it means. Maybe coveralls is slow, not sure

paskal: Might be related to parallel flag?

Smolevich: I think that it's related with fact that PR from fork repo to original

Smolevich: @umputun @paskal do you want that I will revert old definition for this step?

paskal: I am OK with that only if it works. This action is run by CI in this repo and there shouldn't be problems with a fork: fork results would go into fork coveralls account.
Can you please try to make it work first? If you'll enable coveralls for your own account, your fork results should appear there.

Smolevich: https://coveralls.io/jobs/61033004 build in my forked repo

Smolevich: https://coveralls.io/jobs/61033012 last build for PR here

paskal: @Smolevich problem is it's still in `pending completion` state. I ask again, could be it because of a parallel flag? What's the point of having it?

paskal: Other PR in Coveralls example:
image
I don't see a single coveralls run assigned to this PR (e.g. a line with `PULL #632`), in [coveralls builds](https://coveralls.io/repos/289736/builds) page so even if it works it still loses PR annotation. My point is still the same: either fix it or remove it, but in the current state of this PR, something is not right.

paskal: [It seems to work](https://coveralls.io/builds/29871429) after the last commit: I was right to be suspicious about the parallel flag. I am OK with switching to this.

paskal: I gave it some more thought and I tend to agree with @umputun : we're introducing some magic here which is less transparent than plain shell code (but does run some shell code underneath) and gives us no benefits.

umputun: Theoretically, I can see how this can be useful in the form of universal better-go-actions including inside everything needed to build, test, lint and make/submit coverage reports. However our flow, even with this piece, still half-declarative and half-imperative, so I struggle to see how this change will make anything better.

Smolevich: Action with coveralls instead shell script can be useful if this functional is needed more than one repository 
I agree that we don't have benefit for current case

paskal: @Smolevich could you then please revert goveralls change and leave only other changes with code clarification? Also there might some small things to fix after the latest changes, like a typo in `tast and lint examples`. Thank you!

Smolevich: @paskal i revert changes and fix typo

Smolevich: @umputun, are you agree with changes?

PR #635, opened 2020-04-03
Add problem matcher

ralder:


ralder:
opss

paskal:
@ralder will you create a PR on that? Looks promising.

PR #640, opened 2020-04-05
Site-wide email notifications for site owner 

### LINKED ISSUE #633: Site-wide email notifications for site owner
I see a scope selection dropdown for RSS, but not for email.  So it seems we'd need to subscribe separately for each page we create.

Is it possible to get an email notification for comments posted to all pages on our site, or set up a user as subscribed to everything automatically?

paskal:
Add admin email notifications on new comments. Resolves #633.

<details>
<summary>Screenshots</summary>

Emails on new top-level comment and reply to it, with emails to admin and to parent comment user:

image

Admin notification on top-level comment, style can be improved by frontenders (@akellbl4 ?):

image

Notification on reply comment (I've removed "for dev_user" from here as well, screenshot is incorrect only about it):

image

User notification on reply should be unaltered to the current state:

image
</details>

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -200,6 +200,7 @@ type NotifyGroup struct {
 	Email struct {
 		From                string `long:"fromAddress" env:"FROM" description:"from email address"`
 		VerificationSubject string `long:"verification_subj" env:"VERIFICATION_SUBJ" description:"verification message subject"`
+		AdminNotifications  bool   `long:"notify_admin" env:"ADMIN" description:"notify admin on new comments via ADMIN_SHARED_EMAIL"`
 	} `group:"email" namespace:"email" env-namespace:"EMAIL"`
 }
 
@@ -785,6 +786,10 @@ func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *
 					}
 					return tkn, nil
 				},
+			}
+			// enable admin notifications only if admin email is set
+			if s.Notify.Email.AdminNotifications && s.Admin.Shared.Email != "" {
+				emailParams.AdminEmail = s.Admin.Shared.Email
 			}
 			smtpParams := notify.SmtpParams{
 				Host:     s.SMTP.Host,
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -438,6 +438,11 @@ func (s *ServerCommand) newServerApp() (*serverApp, error) {
 		SimpleView:         s.SimpleView,
 	}
 
+	// enable admin notifications only if admin email is set
+	if s.Notify.Email.AdminNotifications && s.Admin.Shared.Email != "" {
+		srv.AdminEmail = s.Admin.Shared.Email
+	}
+
 	srv.ScoreThresholds.Low, srv.ScoreThresholds.Critical = s.LowScore, s.CriticalScore
 
 	var devAuth *provider.DevAuthServer
@@ -786,10 +791,6 @@ func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *
 					}
 					return tkn, nil
 				},
-			}
-			// enable admin notifications only if admin email is set
-			if s.Notify.Email.AdminNotifications && s.Admin.Shared.Email != "" {
-				emailParams.AdminEmail = s.Admin.Shared.Email
 			}
 			smtpParams := notify.SmtpParams{
 				Host:     s.SMTP.Host,
--- a/backend/app/notify/email.go
+++ b/backend/app/notify/email.go
@@ -14,6 +14,7 @@ import (
 
 	log "github.com/go-pkgz/lgr"
 	"github.com/go-pkgz/repeater"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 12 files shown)


backend/app/notify/email.go
```diff
@@ -227,47 +239,105 @@ func NewEmail(emailParams EmailParams, smtpParams SmtpParams) (*Email, error) {
...
-	}
 	select {
 	case <-ctx.Done():
 		return errors.Errorf("sending message to %q aborted due to canceled context", req.Email)
 	default:
 	}
+
+	var emails []emailMessage
+
+	email, err := e.createUserEmail(req)
```
umputun, line 254: I think i complained about it before, but every time I see "email" in either param of var it takes mental efforts to figure what this was not email address but email message. Calling it emailMsgs will help

backend/app/notify/email.go
```diff
@@ -227,47 +239,105 @@ func NewEmail(emailParams EmailParams, smtpParams SmtpParams) (*Email, error) {
...
+	if err != nil {
+		return err
+	}
+	if email != nil {
+		emails = append(emails, *email)
+	}
+
+	errs := new(multierror.Error)
+	for _, email := range emails {
+		errs = multierror.Append(errs, repeater.NewDefault(5, time.Millisecond*250).Do(
```
umputun, line 272: nit: mkaing it 2-liner will help to read it better

backend/app/notify/email.go
```diff
@@ -227,47 +239,105 @@ func NewEmail(emailParams EmailParams, smtpParams SmtpParams) (*Email, error) {
...
+			ctx,
+			func() error {
```
umputun, line 274: and this is a good candidate for one-liner

backend/app/notify/email.go
```diff
@@ -227,47 +239,105 @@ func NewEmail(emailParams EmailParams, smtpParams SmtpParams) (*Email, error) {
...
-		ctx,
-		func() error {
-			return e.sendMessage(emailMessage{from: e.From, to: req.Email, message: msg})
-		})
+	return &emailMessage{from: e.From, to: req.Email, message: msg}, nil
+}
+
+// construct email for admin if it's enabled
+func (e *Email) createAdminEmail(req Request) (*emailMessage, error) {
+	switch {
```
umputun, line 318: Usually I'd preffer case to 3 ifs, but here it feels strange and unusual as all it does an early exit. Ifs feel more apropriate

backend/app/cmd/server.go
```diff
@@ -198,8 +198,10 @@ type NotifyGroup struct {
 		API     string        `long:"api" env:"API" default:"https://api.telegram.org/bot" description:"telegram api prefix"`
 	} `group:"telegram" namespace:"telegram" env-namespace:"TELEGRAM"`
 	Email struct {
-		From                string `long:"fromAddress" env:"FROM" description:"from email address"`
-		VerificationSubject string `long:"verification_subj" env:"VERIFICATION_SUBJ" description:"verification message subject"`
+		From                 string `long:"fromAddress" env:"FROM" description:"from email address"`
+		VerificationSubject  string `long:"verification_subj" env:"VERIFICATION_SUBJ" description:"verification message subject"`
+		AdminNotifications   bool   `long:"notify_admin" env:"ADMIN" description:"notify admin on new comments via ADMIN_SHARED_EMAIL"`
+		AdminNotifyOnReplies bool   `long:"notify_admin_on_replies" env:"ADMIN_ON_REPLIES" description:"notify admin on replies to comments as well (default is only to top-level comments)"`
```
umputun, line 204: Is this one really needed? I mean all-or-none sounds simple and predictable, but adding an option to notify on replies in some conflict with the regular notification.  Fot instance if admin want to get notifications on all new comments and replies to his comments he need to follow two different paths. My suggestiong is to keep AdminNotifications only and send everyting

backend/app/notify/email.go
```diff
@@ -227,47 +239,105 @@ func NewEmail(emailParams EmailParams, smtpParams SmtpParams) (*Email, error) {
...
 }
 
-// Send email about reply to Request.Email if it's set, otherwise do nothing and return nil, thread safe
-// do not returns sending error, only following:
-// 1. (likely impossible) template execution error from email message creation from Request
-// 2. message dropped without sending in case of closed ctx
+// Send email about comment reply to Request.Email if it's set,
+// also sends email to site administrator if appropriate option is set.
+// Thread safe
 func (e *Email) Send(ctx context.Context, req Request) (err error) {
```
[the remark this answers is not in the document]
paskal: I've made everything non-blocking which solves this problem. I think two emails should be sent from email.Send, rather than notify.Submit knowing anything more about it's destinations.

backend/app/cmd/server.go
paskal: I've tried making this work for small and not-so-small sites. I think you are right and it's better to have a simple scheme which will work for small sites well without trying to predict potential users desires.

backend/app/notify/email.go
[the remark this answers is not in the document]
umputun: well, I meant something different. Whatever calls `notify.Submit` should do it twice and nothing inside the notify package should be that smart. Generally, the idea of Send sometimes sending one message (email) and sometimes two, feel odd and unexpected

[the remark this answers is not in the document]
paskal: We already have to look up user email before sending a notification, which is exposing a few inner details of `Email` to `notify.Submit`.

For admin notifications outside of `Email` module either `notify.Submit` or it's caller need to fetch the admin email and pass it on.
If we want different title \ content of the message for users and for admin, things complicate: we also would need to pass `ForAdmin` flag alongside the message, which would affect its formatting.

It would make some sense to have this complexity exposed to `notify.Submit` or its callers if we would have more notification destinations with similar requirements of sending sometimes user notifications of different kinds and sometimes admin notifications of different kinds.
I would argue that it's better to contain this logic inside `Email` for now as it's the only module that is aware of admin notifications existence and change it when we'll have at least one more notification destination with feature parity with `Email`.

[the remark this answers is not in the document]
paskal: tl;dr: moving sending logic up the call stack is a good feature for supporting N notification providers, but at the moment we have only 1 notification provider utilizing this logic and it would introduce complexity with no value paying for it.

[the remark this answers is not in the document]
umputun: I have no clue what complexity you meant. All we need is a check and call near this place https://github.com/umputun/remark42/blob/master/backend/app/rest/api/rest_private.go#L134. This place also seems to be the right one to know about admin's email. To satisfy different formatter we have, we could add `Request.SiteAdmin` or smth like this.

backend/app/rest/api/rest_private.go
```diff
@@ -131,9 +132,14 @@ func (s *private) createCommentCtrl(w http.ResponseWriter, r *http.Request) {
 	s.cache.Flush(cache.Flusher(comment.Locator.SiteID).
 		Scopes(comment.Locator.URL, lastCommentsScope, comment.User.ID, comment.Locator.SiteID))
 
+	// user notification
 	if s.notifyService != nil {
 		s.notifyService.Submit(notify.Request{Comment: finalComment})
 	}
+	// admin notification
+	if s.notifyService != nil && s.adminEmail != "" {
+		s.notifyService.Submit(notify.Request{Comment: finalComment, Email: s.adminEmail, ForAdmin: true})
```
paskal, line 141: I can hide `Email` from caller and instead fetch it inside` notify.Submit` if `ForAdmin` is true.

backend/app/rest/api/rest_private_test.go
```diff
@@ -596,11 +596,12 @@ func TestRest_EmailNotification(t *testing.T) {
...
-	time.Sleep(time.Millisecond * 5)
-	require.Equal(t, 1, len(mockDestination.Get()))
-	assert.Equal(t, "", mockDestination.Get()[0].Email)
+	time.Sleep(time.Millisecond * 30)
+	require.Equal(t, 2, len(mockDestination.Get()))
+	assert.Empty(t, mockDestination.Get()[0].Email)
+	assert.Equal(t, "admin@example.org", mockDestination.Get()[1].Email)
 
-	// create child comment from another user, no email notification expected
+	// create child comment from another user, email notification only to admin expected
```
paskal, line 604: I don't like this test complexity increase but don't know what to do with it.

PR #650, opened 2020-04-08
Update typescript to 3.8.3

akellbl4:
Update typescript and tools connected with it.
Also `tsconfig.typecheck.json` was added because an incremental build prohibited with `noEmit` option. For type check, we don't need to build our code, because I added a new config without incremental option.


frontend/tsconfig.typecheck.json
```diff
@@ -0,0 +1,26 @@
+{
```
Mavrin, line 1: Please use extends.
{
  "extends": "./tsconfig.json",
   ...
}

akellbl4: Agree

PR #655, opened 2020-04-10
add emailprovider

ttys3:
move smtp to emailprovider
notify and auth: add mailgun support, add sendgrid support
update README.md

> Mailgun (https://www.mailgun.com/) Free Plan provides 10,000 Emails per month

> SendGrid(https://sendgrid.com) Trial Plan provides 40,000 emails for 30 days, After your trial ends, you can send 100 emails/day for free

reason: 
nowadays SMTP seems a bit more deprecated internet things.
and if you are using a CDN to protect your server from source IP leaking problem,
using SMTP protocol will cause your server's source IP leaking thus the protection work of CDN service providers like cloudflare will be wasted.

from code side:
separate SMTP email sender out of the `notify` or `auth` logic, and abstracted the `sender`  into
an `emailprovider` interface.
this interface provide support for both  `notify` and  `auth`

and, if others like, they can add more usefull email service providers to the project

maybe `emailprovider` is a bit more long name ? any other suggestions ?


umputun:
> nowadays SMTP seems a bit more deprecated internet things.

Confused. SMTP is a thing used by all of these services to deliver emails. Btw, mailgun supports smtp relays just fine. The same for AWS SES. Not sure about sendgrid but looks like it supports SMTP as well https://sendgrid.com/docs/for-developers/sending-email/getting-started-smtp/

ttys3:
> > nowadays SMTP seems a bit more deprecated internet things.
> 
> Confused. SMTP is a thing used by all of these services to deliver emails. Btw, mailgun supports smtp relays just fine. The same for AWS SES. Not sure about sendgrid but looks like it supports SMTP as well https://sendgrid.com/docs/for-developers/sending-email/getting-started-smtp/

yes, SMTP do works, I'm not saying that SMTP not works good.
but like I said, it will cause source IP leaking.
and the `API ways` is what both sendgrid and mailgun recommends, `not the SMTP way`

ttys3:
some links:
[Actual IP outflow when sending mail](https://community.cloudflare.com/t/actual-ip-outflow-when-sending-mail/5680)

[Mailgun leaking server's real IP behind Cloudflare](https://meta.discourse.org/t/mailgun-leaking-servers-real-ip-behind-cloudflare/59128)

image

umputun:
Leaking of what IP exactly you worry about? Remark42 usually runs as a part of the public (web) server/instance, which has publicly available DNS/IP. In other words, IP already "leaked". So. I don't really understand what IP should be protected and why.

My major problem with this request - instead of standard and universal way to deliver emails it offers support of custom APIs of particular providers. Pls, note - this can be achieved with the current code by using some sort of smtp-mailgun bridge. I don't recall details but seen several containers doing exactly this.
.

ttys3:
> Leaking of what IP exactly you worry about? Remark42 usually runs as a part of the public (web) server/instance, which has publicly available DNS/IP. In other words, IP already "leaked". So. I don't really understand what IP should be protected and why.
> 
> My major problem with this request - instead of standard and universal way to deliver emails it offers support of custom APIs of particular providers. Pls, note - this can be achieved with the current code by using some sort of smtp-mailgun bridge. I don't recall details but seen several containers doing exactly this.
> .

you seems did not know CDN tech very well. no do DDoS.
server source IP protection is very import in security 

you think it's nothing because no one attack your server.

umputun:
I really don't care about "knowing CDN tech very well", whatever it means. But I do care about preventing code bloat for no good reason.

Let me suggest one more time - if you have a use case requiring API communication with mailgun and others, there is no need to support all of these providers in the code directly.  As long as remark42 supports the standard way to send emails (aka SMTP) adding a local container with SMTP->Mailgun bridge is a trivial task. And each customer could add whatever bridge they like.

ttys3:
yeah, that's a way. use a side container which provider email sending service.

but I prefer  `WEB API` not the `smtp` protocol.

maybe I can refactor the code into this if I have time:

remove any mailgun or sendgrid related code.
keep the provider. but only contains two provider:
`smtp` and `api`

then build another service which provide the `standard remark42 api` support for remark42.
when using `api` mail provider, remark42 send request to the specific side container (or native service if you don't like container) like:

```
POST /api/{email_sp_provider}

from: from@example.com
to: to@example.com
headers: [{"List-Unsubscribe-Post": "List-Unsubscribe=One-Click"}, {"List-Unsubscribe": "<http://example.co/unsubscrib>"}]
subject: xxxxx
body: xxxx
```

response may like:

```
{
code: 200
message: "message send"
}
```

```
{
code: 400
message: "message send failed, authentication failed ......"
}
```

umputun:
Making your own API to communicate between the core system (remark42) and side container has several disadvantages:

- SMTP is a standard, described in several RFCs. It means the user has a chance to find third-party containers doing this thing. In theory, even service providers may have a bridge ready.
- You will have to maintain all those implementations of side containers forever.

Probably there is a value of making such a thing, but this is outside of the scope of remark42. Maybe someday the protocol you suggesting will be a new standard and I would reconsider, but for now, I'm going to reject this PR.

@paskal - we could add some info into your email document describing email setup and mention a possibility of running smtp->providet-api bridge container.

paskal:
@ttys3 your code is of high quality as far as I can see and it's really unfortunate that it doesn't belong to this project.

However, I was looking for separate SMTP -> SendGrid / Mailgun API software to write instructions about it into the documentation and I found none.

If you have such an ability, I would suggest you publish the code you wrote as a separate project (something like "smtp2api") and would be happy to participate in testing and write documentation about its usage with remark42 in this repo. Source IP protection is indeed a very important concern and I see a lot of value in your proposal.

I would expect it to be a Docker container with the SMTP server with or without authorization, which will re-send all received messages using SendGrid\Mailgun API according to a configuration provided by the user. Having configuration in the environment to ease the use with docker-compose, as well as having example docker-compose would be extremely useful.

Feel free to ping me if you'll need any help, I would be happy to assist with any part of the project.

paskal:
I dig more and found the link I missed in the thread above as well as one other:
https://hub.docker.com/r/stevenolen/mailgun-smtp-server ([Ruby](https://github.com/stevenolen/mailgun-smtp-server))
https://hub.docker.com/r/fgribreau/smtp-to-sendgrid-gateway ([JS](https://github.com/FGRibreau/smtp-to-sendgrid-gateway))
I would still prefer having a separate Go project from the code of this PR as both projects above don't seem to be active.

ttys3:
@paskal thanks for the friendly reply
I stared write the SMTP bridge server today.

most of the part has been done,  currenttly left CI and container image building job.

as  umputun said, adding a local container with SMTP->Mailgun bridge is a trivial task.

but it is not that `trivial` if you want to implement a whole SMTP server.
implement a SMTP server which make remark42 done the job is no problem.

since it only sending simply HTML content.
no attachment. no pdf. no pictures. no sub message.  no multipart.

Issue #584, opened 2020-01-27
False-positve on image removal from the staging


umputun:
I have seen the missing images in several comments. This is a rare event but still

I think, there are two reasons for this:

1. sometimes cleanup activated on exact TTL and commit may be delayed for at least 10ms
1. the logic of submitting multiple ids is sequential and blocking. I didn't expect it to delay commit significantly and still don't understand how this even possible, but I see an example of a failing attempt to commit already removed staging image with a strangely long delay between cleanup (07:46:34.384) and commit (07:47:22.677)

Infinoid:
I had this problem while using the `umputun/remark42:latest` docker image, and then saw this ticket.  So I updated to `umputun/remark42:master`, and just saw the problem again.  So I think the bug is still present.

I've reproduced the problem 3 times in a row; do you have any ideas for things I can try?

paskal:
@Infinoid could you please post here the app version which is printed at the beginning of the log after the container starts?

Infinoid:
```
init container
set timezone America/Chicago (Wed Apr  1 09:26:06 CDT 2020)
custom APP_UID not defined, using default uid=1001
execute /srv/init.sh
prepare environment
execute "/srv/remark42 server"
remark42 master-10c1aa9-20200401-02:36:51
2020/04/01 09:26:07.238 [INFO]  {cmd/server.go:253 cmd.(*ServerCommand).Execute} start server on port 8080
```

paskal:
As this issue might be tied to settings values, could you please share a non-sensitive part of your docker-compose.yml file?
I think it would be better to upload it to Pastebin to prevent having a commend with 50 lines of code there.

I’ve worked on image code recently and will try to write a failing test case with this problem reproduction if @umputun was right about the root cause.
But it’s very unlikely because of the applied fix until the commit time is very small and the difference between its value and 3x its value is neglectable in the scale of CPU/io, so I assume anything higher than 100ms couldn’t be a problem.

Infinoid:
https://gist.github.com/Infinoid/5e290480d67a586870b071cb7bb00967

Infinoid:
I am doing the following to reproduce the issue:

1. Type a sentence of text into the comment field
2. Hit enter a couple of times
3. Find an image in google image search, copy image, paste it into the comment box
4. Click Send
5. Wait 30-40 minutes
6. See a log entry about the image being cleaned up (see below)
7. Refresh page and see my comment has a broken icon where the image was

The log entry looks like this:
[...]

PR #658, opened 2020-04-11
Application shutdown problem fix

paskal:
- fix a few things in Readme
- shutdown all services before stopping the application (previously shutting down services was a non-blocking operation and likely did not complete properly)
- fix `image.Service.Close()` as previously it was stuck for eternity in case `image.Service.Submit` was not called at least once during the application run

This PR likely will improve the situation with the problem described in #584 as previously `image.Service.Close()` likely was not even called on program shutdown.

### DIFF
--- a/backend/app/store/image/image.go
+++ b/backend/app/store/image/image.go
@@ -149,14 +149,20 @@ func (s *Service) Cleanup(ctx context.Context) {
 }
 
 // Close flushes all in-progress submits and enforces waiting commits
-func (s *Service) Close() {
+func (s *Service) Close(ctx context.Context) {
 	log.Printf("[INFO] close image service ")
 	// terminate commit goroutine using s.term only if it is started
 	if atomic.LoadInt32(&s.term) == 1 {
 		atomic.StoreInt32(&s.term, 0) // enforce non-delayed commits for all ids left in submitCh
+		var ctxCancel bool
 		for {
+			select {
+			case <-ctx.Done():
+				ctxCancel = true
+			default:
+			}
 			// set to 1 by commit goroutine after everything waited on TTL sent
-			if atomic.LoadInt32(&s.term) == 1 {
+			if atomic.LoadInt32(&s.term) == 1 || ctxCancel {
 				break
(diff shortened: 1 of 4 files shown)


backend/app/store/image/image.go
```diff
@@ -150,13 +151,16 @@ func (s *Service) Cleanup(ctx context.Context) {
...
 	log.Printf("[INFO] close image service ")
-	atomic.StoreInt32(&s.term, 1) // enforce non-delayed commits for all ids left in submitCh
-	for {
-		// set to 0 by Commit goroutine after everything waited on TTL sent
-		if atomic.LoadInt32(&s.term) == 0 {
-			break
+	// terminate commit goroutine using s.term only if it is started
+	if atomic.LoadInt32(&s.term) == 1 {
+		atomic.StoreInt32(&s.term, 0) // enforce non-delayed commits for all ids left in submitCh
+		for {
```
umputun, line 157: not related yo your changes but a general suggestion - Close is blocking operation which may take a long time. Maybe we want to pass ctx for early termination?

paskal: It's an interesting point. I would suggest merging this PR as it makes something which worked fast but unreliable working slower but 100% reliable.

---

Adding context cancelation which later would be just checked by Close() function makes things more complex but would speed them up.

As far as I see one of the concerns is that we need to wait for an HTTP server to stop first to prevent crippled answers to requests in progress, which would leave us with everything but HTTP server to shut down. That means that we can parallelize only things that terminate later in code.
Both with or without concern above I don't know how much time we really save on parallelizing these things: slowest operation is disk IO, so `image.Store.Close()` moving files on HDD from staging to permanent storage would take magnitudes higher amount of time than any other operation, driving effects of parallelization to zero. And un scenario `image.Store.Close()` doesn't need to do anything or moves files fast on SSD, we are saving previous calls unparalleled execution time which is unknown and we shouldn't care about until we really see a problem there.

In summary, I think it's premature optimization and we shouldn't care about it yet.

PR #659, opened 2020-04-11
Fix docker run error without attached /srv/var

paskal:
Before:

```
➜  docker run umputun/remark42:dev
init container
set timezone America/Chicago (Sat Apr 11 11:19:28 CDT 2020)
custom APP_UID not defined, using default uid=1001
execute /srv/init.sh
prepare environment
/srv/init.sh failed
```

After:

```
➜ docker run umputun/remark42:dev
init container
set timezone America/Chicago (Sat Apr 11 15:30:59 CDT 2020)
custom APP_UID not defined, using default uid=1001
execute /srv/init.sh
prepare environment
ERROR: /srv/var doesn't exist, which means that state of the application
ERROR: will be lost on container stop or restart.
ERROR: Please mount local directory to /srv/var in order for it to work.
/srv/init.sh failed
```

### DIFF
--- a/docker-init.sh
+++ b/docker-init.sh
@@ -5,4 +5,11 @@ sed -i "s|https://demo.remark42.com|${REMARK_URL}|g" /srv/web/*.js
 # remove devtools attach helper. TODO: move to webpack loader
 sed -i "/REMOVE-START/,/REMOVE-END/d" /srv/web/iframe.html
 
-chown -R app:app /srv/var 2>/dev/null
+if [ -d "/srv/var" ]; then
+  chown -R app:app /srv/var 2>/dev/null
+else
+  echo "ERROR: /srv/var doesn't exist, which means that state of the application"
+  echo "ERROR: will be lost on container stop or restart."
+  echo "ERROR: Please mount local directory to /srv/var in order for it to work."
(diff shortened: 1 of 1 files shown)


docker-init.sh
```diff
@@ -5,4 +5,5 @@ sed -i "s|https://demo.remark42.com|${REMARK_URL}|g" /srv/web/*.js
 # remove devtools attach helper. TODO: move to webpack loader
 sed -i "/REMOVE-START/,/REMOVE-END/d" /srv/web/iframe.html
 
+mkdir -p /srv/var
```
umputun, line 8: this change effectively makes remark42 ephemeral in cases it no mounting provided. The side effects can be surprising for the user, i.e. restarted/recreated containers will lose all data.

Failing, in this case, feels like a better approach. I suggest adding an `if` check to the script and show some friendly message.

paskal: Giving the commit history, I would argue that failing the start of the container without `/srv/var` was surprising for the user in the first place as it seems to be a semi-random side-effect of the change which broke it:)
I completely agree that it's better to document that behavior with a proper error message. I adjusted PR accordingly.

PR #660, opened 2020-04-11
Highlight comment when user went to it by link

### LINKED ISSUE #559: Highlight comment when user went to it by link
We can highlight a comment for 5sec when a user goes to it by link.
It can attract attention of user to right place of view.

lalentine96:
Resolves #559

Adds .comment_highlighting class after scrolling to comment.

### DIFF
--- a/frontend/app/components/comment/_theme/_dark/comment_theme_dark.scss
+++ b/frontend/app/components/comment/_theme/_dark/comment_theme_dark.scss
@@ -95,4 +95,8 @@
       }
     }
   }
+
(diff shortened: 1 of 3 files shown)


paskal:
Thanks for your contribution! LGTM

image

akellbl4:
@lalentine96 thanks! LGTM, but I can suggest adding `transition: background 0.3s ease-in-out`.

It could be better to add paddings to horizontal sides, but now it not available for comments on the first level. I think we can discuss it in a separate issue.

PR #662, opened 2020-04-11
Return proper content type instead of image/*

### LINKED ISSUE #598: Lack of proper image types
Part of #447 lost image types and replaced `image/gif`, `image/png` and so on with the generic `image/*` type.

This is the quote:

> in order to guaranty that image ID doesn't change (which happened in case of resizing and FS image store) I removed storing images extensions for FS store at all. content-type for cached images is image/* which seems to be accepted by browsers without any problems.

The problem with `image/*` is not how we show it on the page (this works just fine), but how to do "Open Image in new tab/window". Both Chrome and Safari refuse to open the image and treated it as a downloadable file which is very unexpected behavior.

@smaant - I think we should restore image type somehow. Maybe we could exclude extension from image ID to prevent the confusion after resizing, not sure. Let me know what you think

paskal:
Resolve #598.

### Current master

```
➜ curl -o /dev/null -D - -s https://127.0.0.1:8080/api/v1/picture/github_f1f725d06af013456a9b98d923f9b2ce5c7ba91a/bq906dihajfi6fd90ci0 | grep '^Content-Type:'
Content-Type: image/*
```

### After this PR

```
➜  remark42 git:(master) ✗ curl -o /dev/null -D - -s http://127.0.0.1:8080/api/v1/picture/dev_user/bq90856q9r8d70c29vpg | grep '^Content-Type:'
Content-Type: image/png
```

Also opening such an image in a browser now shows it instead of opening save prompt.

I've tested image save and load but didn't test proxy changes as it's not clear for me when it's involved. But I think it should work fine.


paskal:
I can't figure out why [this](https://github.com/umputun/remark42/pull/662/checks?check_run_id=579467855#step:6:18) fails, working on it.

Issue #444, opened 2019-10-10
"exec format error" when compiling on Raspberry Pi 4


quantonganh:
`armv7l` is 32-bit, so I tried to compile from source by running:

```
make OS=linux ARCH=arm32
```

and got the below error:

```
Step 28/34 : RUN     export WEB_ROOT=/go/src/github.com/umputun/remark/backend/web &&     sed -i "s|https://demo.remark42.com|http://127.0.0.1:8080|g" ${WEB_ROOT}/*.js &&     sed -i "/REMOVE-START/,/REMOVE-END/d" ${WEB_ROOT}/iframe.html &&     go get -v github.com/rakyll/statik &&     statik --src=${WEB_ROOT} --dest=/go/src/github.com/umputun/remark/backend/app/rest -p api -f &&     ls -la /go/src/github.com/umputun/remark/backend/app/rest/api/statik.go &&     ls -la /go/src/github.com/umputun/remark/backend/web/
 ---> Running in ac3923f42342
standard_init_linux.go:211: exec user process caused "exec format error"
```

The reason is at the step 20:

```
Step 20/34 : FROM umputun/baseimage:buildgo-latest as build-backend
 ---> cc19be47a26c
```

This image is not compiled for arm.

umputun:
A little bit confused - the title says "compiling on Raspberry Pi 4", but the make file you mentioned intended to help with cross-compile, i.e. compile binary for different platforms on something more ... traditional.

All binaries in remark42 releases were made with `make deploy`, so it can compile for different architectures just fine and it doesn't require differnet base images. However `ARCH=arm32` is not the architecture supported by go compiler. I think you want `arm` architecture (see https://golang.org/doc/install/source#environment). 

Pls note - Docker.artifact didn't make binary for `arm` and I just added it. With this change you should be able to run `make bin ARCH=arm` and it will build the binary you are looking for.

quantonganh:
> A little bit confused - the title says "compiling on Raspberry Pi 4", but the make file you mentioned intended to help with cross-compile, i.e. compile binary for different platforms on something more ... traditional.

No, I'm on RP4 and just follow your [README.md](https://github.com/umputun/remark#without-docker):

> alternatively compile from the sources - make OS=[linux|darwin|windows] ARCH=[amd64,386,arm64,arm32]

> All binaries in remark42 releases were made with `make deploy`, so it can compile for different architectures just fine and it doesn't require differnet base images.

[...]

PR #664, opened 2020-04-12
fix artifact filename on arm32

quantonganh:
https://github.com/umputun/remark42/issues/444


paskal:
@quantonganh I am pretty sure architecture name is [arm](https://github.com/golang/go/wiki/MinimumRequirements#arm) and not arm32, I never seen builds with arm32 in name anywhere and couldn't find them now. Why do you think this change is necessary?
//cc @umputun

paskal:
Here is [another link](https://build.golang.org/#) where you can see it goes under name `arm` and not `arm32`.

quantonganh:
@paskal You're completely right:

```
$ go tool dist list | grep arm
linux/arm
linux/arm64
```

I'm pretty sure I checked the correct architecture name before, but don't know why I just follow the README blindly:

> alternatively compile from the sources - make OS=[linux|darwin|windows] ARCH=[amd64,386,arm64,arm32]

It should be fixed to avoid confusion in the future.
Created: https://github.com/umputun/remark42/pull/684

PR #680, opened 2020-04-13
Revert "fix artifact filename on arm32"

paskal:
Reverts umputun/remark42#664 as I think introduced change should not be in place, see discussion in original PR.

//cc @quantonganh


paskal:
Close in favor of #684.

PR #684, opened 2020-04-13
the correct GOARCH value is arm, not arm32

quantonganh:


paskal:
Could you please add https://github.com/umputun/remark42/pull/680 change into this PR as well?

quantonganh:
@paskal Reverted. I'm sorry about that.

paskal:
No problem at all! I'm glad we figured out the source of confusion and will fix it.

PR #687, opened 2020-04-14
 serve proxified images disregarding if saving is enabled

### LINKED ISSUE #614: Disabling image proxy breaks cached images
Creating an separate issue from [this PR discussion](https://github.com/umputun/remark42/pull/447#discussion_r363167431). Problem statement:

https://github.com/umputun/remark42/blob/019fa9e5ac2fc79dba64f8231be9ef65c6471b81/backend/app/rest/proxy/image.go#L87-L91

Proposed solution by @umputun:
> Probably we can always serve cached images regardless of CacheExternal flag. Not 100% sure, just an idea. In this case, already converted images will be always reachable and if admin wants to get rid of them for real, he could remove them from the store.

paskal:
Resolves #614.

- serve `/api/v1/img` requests disregarding if saving of image proxy is enabled or not
  This does increase the load time for every proxified image with caching disabled, as we first try to read it from cache anyway.
- document behavior above in the readme
- fix reference to outdated  ENV configuration variable in Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -879,5 +879,6 @@ _all admin calls require auth and admin privilege_
 * User can edit comments in 5 mins (configurable) window after creation.
 * User ID hashed and prefixed by oauth provider name to avoid collisions and potential abuse.
 * All avatars resized and cached locally to prevent rate limiters from oauth providers, part of [go-pkgz/auth](https://github.com/go-pkgz/auth) functionality.
-* Images can be proxied (`IMG_PROXY=true`) to prevent mixed http/https.
+* Images can be proxied (`IMAGE_PROXY_HTTP2HTTPS=true`) to prevent mixed http/https.
+* All images can be proxied and saved (`IMAGE_PROXY_CACHE_EXTERNAL=true`) instead of serving from original location. Beware, images which are posted with this parameter enabled will be served from proxy even after it will be disabled.
 * Docker build uses [publicly available](https://github.com/umputun/baseimage) base images.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -879,5 +879,5 @@ _all admin calls require auth and admin privilege_
 * User can edit comments in 5 mins (configurable) window after creation.
 * User ID hashed and prefixed by oauth provider name to avoid collisions and potential abuse.
 * All avatars resized and cached locally to prevent rate limiters from oauth providers, part of [go-pkgz/auth](https://github.com/go-pkgz/auth) functionality.
-* Images can be proxied (`IMG_PROXY=true`) to prevent mixed http/https.
+* Images can be proxied (`IMAGE_PROXY_HTTP2HTTPS=true`) to prevent mixed http/https. Beware, images which are posted with this parameter enabled will be served from proxy even when it will be disabled.
```
umputun, line 882: I don't get it. http2https doesn't serve anything "from proxy", all it does it reverse proxy and serves from the original location unless CACHE_EXTERNAL enabled

backend/app/rest/proxy/image.go
```diff
@@ -84,15 +84,6 @@ func (p Image) replace(commentHTML string, imgs []string) string {
 
 // Handler returns http handler respond to proxied request
 func (p Image) Handler(w http.ResponseWriter, r *http.Request) {
-	if !p.HTTP2HTTPS && !p.CacheExternal {
```
umputun, line 87: not 100% sure but I think it effectively enforces http->https proxy regardless the settings

umputun:
the PR missing some test proving how it works in real life. I.e. serve with cached proxy, turn the cached proxy off and try to get the image

README.md
paskal: Adjusted readme to reflect that.

backend/app/rest/proxy/image.go
paskal: Converter won't alter text of new comments. But existing links will continue to work, which is exactly what we want.

paskal:
I've added test which returns `501 Not Implemented` before this PR and proper image with it.

PR #691, opened 2020-04-18
Add message of image upload try by anonymous user

### LINKED ISSUE #661: Lack of message in case of image upload try by anonymous user
Hello there!

If I drag-and-drop image onto the comment box as logged in user, it gets uploaded.

In case I do it as an anonymous user, nothing happens. I think this is wrong and we should show an error message instead, like "Please log in not as anonymous user to be able to attach images".

dellink:
fix #661 
Screen Shot 2020-04-18 at 17 29 24


frontend/app/components/comment-form/comment-form.tsx
```diff
@@ -273,10 +279,13 @@ export class CommentForm extends Component<Props, State> {
   }
 
   onDrop(e: DragEvent) {
-    if (!this.props.user) {
+    const isAnonymous = this.props.user && isUserAnonymous(this.props.user);
```
akellbl4, line 282: `this.props.user` can be `User` or `null`. `isUserAnonymous()` is already checking whether user is equal to `null` or not.

frontend/app/locales/de.json
```diff
@@ -53,7 +53,7 @@
   "comment.verified-user": "Bestätigter Benutzer",
   "comment.verify-user": "Möchtest du den Benutzer {userName} wirklich bestätigen?",
   "comment.vote-error": "Fehler beim Abstimmen: {voteErrorMessage}",
-  "commentForm.anonymous-uploading-disabled": "Der Upload von Bildern ist nur für registrierte Benutzer möglich. Bitte melde dich zuvor an.",
+  "commentForm.unauthorized-uploading-disabled": "Der Upload von Bildern ist nur für registrierte Benutzer möglich. Bitte melde dich zuvor an.",
```
akellbl4, line 56: Lang generation is broken, It's dropped message for `anonymous-uploading-disabled`.
I think it happened because the generator is not able to extract translation when you get it with trinary operator.

frontend/app/components/comment-form/comment-form.tsx
dellink: I suppose `isUserAnonymous` is incorrect. Anonymous is user who logged as anonymous, not null.

akellbl4: Oh, yes, sorry. You are right.

frontend/app/components/comment-form/comment-form.test.tsx
```diff
@@ -131,4 +131,24 @@ describe('<CommentForm />', () => {
...
   });
+
+  it('should show error message of image upload try by anonymous user', () => {
+    const props = { ...DEFAULT_PROPS, anonymousUser, intl };
+    const wrapper = shallow(<CommentForm {...props} />);
+    // @ts-ignore
+    const instance: CommentForm = wrapper.instance();
+
+    instance.onDrop(new Event('drag') as DragEvent);
+    expect(wrapper.exists('.comment-form__error')).toEqual(true);
```
akellbl4, line 142: Nice job. But I suggest adding that the right message is shown.

akellbl4:
Almost done. You fixed message getting and you should regenerate translations. After that, you will see a new translation line in JSON.

PR #693, opened 2020-04-19
Image commit on start

### LINKED ISSUE #584: False-positve on image removal from the staging

paskal:
Resolves #584. Relies on merging #692 first.

## Behavior before this PR
1. Set `EDIT_TIME=20s` in docker-compose.yml
2. Start dev instance using docker-compose, post comment with image
3. Restart docker-compose
4. Wait for 2 minutes
5. Reload the page: image from the comment is no longer available as it was stuck in Staging without commit and then was deleted after `EDIT_TIME*5`, `1m40s` in this case.

## Behavior after this PR
1. Set `EDIT_TIME=20s` in docker-compose.yml
2. Start dev instance using docker-compose, post comment with image
3. Restart docker-compose
4. Wait for 2 minutes
5. Reload the page: image is still in place

### DIFF
--- a/backend/app/store/image/fs_store.go
+++ b/backend/app/store/image/fs_store.go
@@ -115,6 +115,33 @@ func (f *FileSystem) Cleanup(_ context.Context, ttl time.Duration) error {
 	return errors.Wrap(err, "failed to cleanup images")
 }
 
+// Info returns meta information about storage
+func (f *FileSystem) Info() (StoreInfo, error) {
+	if _, err := os.Stat(f.Staging); os.IsNotExist(err) {
+		return StoreInfo{}, nil
+	}
+
+	var ts time.Time
+	err := filepath.Walk(f.Staging, func(fpath string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		created := info.ModTime()
+		if ts.IsZero() || created.Before(ts) {
+			ts = created
+		}
+		return nil
+	})
+	if err != nil {
+		return StoreInfo{}, errors.Wrapf(err, "problem retrieving first timestamp from staging images on fs")
+	}
+	return StoreInfo{FirstStagingImageTS: ts}, nil
+}
+
 // location gets full path for id by adding partition to the final path in order to keep files in different subdirectories
 // and avoid too many files in a single place.
 // the end result is a full path like this - /tmp/images/user1/92/xxx-yyy.png.
--- a/backend/app/store/image/image.go
+++ b/backend/app/store/image/image.go
@@ -52,6 +52,11 @@ type ServiceParams struct {
 	MaxWidth  int
 }
 
+// StoreInfo contains image store meta information
+type StoreInfo struct {
+	FirstStagingImageTS time.Time
+}
+
 // To regenerate mock run from this directory:
 // sh -c "mockery -inpkg -name Store -print > /tmp/image-mock.tmp && mv /tmp/image-mock.tmp image_mock.go"
 
@@ -60,6 +65,7 @@ type ServiceParams struct {
 // Two-stage commit scheme is used for not storing images which are uploaded but later never used in the comments,
 // e.g. when somebody uploaded a picture but did not sent the comment.
 type Store interface {
+	Info() (StoreInfo, error)         // get meta information about storage
 	Save(id string, img []byte) error // store image with passed id to staging
 	Load(id string) ([]byte, error)   // load image by ID
 
@@ -148,6 +154,11 @@ func (s *Service) Cleanup(ctx context.Context) {
 			}
 		}
 	}
+}
+
+// Info returns meta information about storage
+func (s *Service) Info() (StoreInfo, error) {
+	return s.store.Info()
 }
 
 // Close flushes all in-progress submits and enforces waiting commits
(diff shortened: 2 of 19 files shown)


backend/app/store/service/service.go
```diff
@@ -200,6 +200,30 @@ func (s *DataStore) DeleteUserDetail(siteID string, userID string, detail engine
...
+		return err
+	}
+	ts := info.FirstStagingImageTS
+	if ts.IsZero() {
+		return nil
+	}
+	for _, site := range sites {
+		locator := store.Locator{SiteID: site}
+		comments, err := s.FindSince(locator, "time", store.User{}, ts)
+		if err != nil {
```
umputun, line 217: as this operation runs for different sites it makes sense not to terminate on the first error but collect in multierror and continute.And return collected error(s) after this

backend/app/store/service/service_test.go
```diff
@@ -1300,6 +1302,88 @@ func TestService_submitImages(t *testing.T) {
...
+	_, err := b.Engine.Create(comment)
+	require.NoError(t, err)
+
+	// resubmit single comment with single image
+	mockStore.On("Info").Once().Return(image.StoreInfo{FirstStagingImageTS: time.Time{}.Add(time.Second)}, nil)
+	err = b.ResubmitStagingImages([]string{"radio-t"})
+	assert.NoError(t, err)
+
+	// wait for Submit goroutine to commit image
+	mockStore.On("Commit", "dev_user/bqf122eq9r8ad657n3ng").Once().Return(nil)
```
umputun, line 1336: the test will be more convincing if at least 2 images processed

backend/app/store/service/service_test.go
```diff
@@ -1300,6 +1302,88 @@ func TestService_submitImages(t *testing.T) {
...
+
+	// resubmit single comment with single image
+	mockStore.On("Info").Once().Return(image.StoreInfo{FirstStagingImageTS: time.Time{}}, nil)
+	err := b.ResubmitStagingImages([]string{"radio-t"})
+	assert.NoError(t, err)
+
+	mockStore.AssertNumberOfCalls(t, "Info", 1)
+}
+
+func TestService_ResubmitStagingImagesError(t *testing.T) {
```
umputun, line 1365: will be nice to cover the case with partial failure (one passed, one failed)

backend/app/store/image/image.go
```diff
@@ -76,6 +88,10 @@ type submitReq struct {
 
 // NewService returns new Service instance
 func NewService(s Store, p ServiceParams) *Service {
+	p.commitTTL = p.EditDuration * 15 / 10  // Commit call on every 1.5 * EditDuration
```
umputun, line 91: how 15/10 is easier to understand? even comment saying 1.5x

backend/app/store/image/image.go
```diff
@@ -76,6 +88,10 @@ type submitReq struct {
...
+	p.cleanupTTL = p.EditDuration * 25 / 10 // Cleanup call on every 2.5 * EditDuration
```
umputun, line 92: the same, this is 2.5x and easer to read as `*2.5`

backend/app/store/image/image.go
```diff
@@ -133,23 +149,28 @@ func (s *Service) ExtractPictures(commentHTML string) (ids []string, err error)
...
 
 	for {
 		select {
 		case <-ctx.Done():
 			log.Printf("[INFO] cleanup terminated, %v", ctx.Err())
 			return
-		case <-time.After(s.TTL / 2): // cleanup call on every 1/2 TTL
-			if err := s.store.Cleanup(ctx, s.TTL); err != nil {
+		case <-time.After(s.cleanupTTL):
+			if err := s.store.Cleanup(ctx, s.EditDuration); err != nil {
```
umputun, line 162: the knowlege about internalas of other parts with such a concete EditDuration name looks strange. Not sure what should be the right name.Probably this is cleanupTTL, and the thing you currently call cleanupTTL is some sort of clanupInterval or cleanupDuration.

backend/app/store/image/fs_store.go
```diff
@@ -115,6 +115,33 @@ func (f *FileSystem) Cleanup(_ context.Context, ttl time.Duration) error {
 	return errors.Wrap(err, "failed to cleanup images")
 }
 
+// Info returns meta information about storage
+func (f *FileSystem) Info() (StoreInfo, error) {
+	if _, err := os.Stat(f.Staging); os.IsNotExist(err) {
+		return StoreInfo{}, nil
```
umputun, line 121: this is top-level API func and returned errors better to be wrapped with some context

backend/app/store/image/image.go
paskal: this has to be integers and not float value, I can't put `* 1.5` there unfortunately.

backend/app/store/image/image.go
paskal: Changed to `cleanupTTL` which is more in-line with original value.

backend/app/store/image/image.go
umputun: ah, ok.

PR #701, opened 2020-04-24
Commit proxy images same (delayed) way as normally uploaded ones

### LINKED ISSUE #694: Proxified images commited disregarding the edit period
$subj, now proxyfied images stored permanently even if they edited out of the comment. Following function can be used to fix that:
https://github.com/umputun/remark42/blob/b5e31f30818eaf63c36e1b9fe10d1bcb0796ddf4/backend/app/store/service/service.go#L205

paskal:
Resolves #694.

Behavior before this PR

- Set `EDIT_TIME=20s` and `IMAGE_PROXY_CACHE_EXTERNAL=true` in docker-compose.yml
- Start dev instance using docker-compose, post comment with text ``
- Restart docker-compose
- Wait for 2 minutes
- Reload the page: image from the comment will be re-loaded as it was stuck in Staging without commit and then was deleted after `EDIT_TIME*5`, `1m40s` in this case. And with current logic, we re-load it in case it's not in storage.

Behavior after this PR

- Set `EDIT_TIME=20s` and `IMAGE_PROXY_CACHE_EXTERNAL=true` in docker-compose.yml
- Start dev instance using docker-compose, post comment with text ``
- Restart docker-compose
- Wait for 2 minutes
- Reload the page: the image is still in place and loaded from storage.
  Message in logs:
  ```
  Commit image cached_images/da39a3ee5e6b4b0d3255bfef95601890afd80709-da39a3ee5e6b4b0d3255bfef95601890afd80709
  ```

### DIFF
--- a/backend/app/store/service/service_test.go
+++ b/backend/app/store/service/service_test.go
@@ -19,7 +19,6 @@ import (
 	"github.com/go-pkgz/lgr"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	bolt "go.etcd.io/bbolt"
 
@@ -100,7 +99,6 @@ func TestService_CreateFromPartial(t *testing.T) {
 }
 
 func TestService_CreateFromPartialWithTitle(t *testing.T) {
-
 	ks := admin.NewStaticKeyStore("secret 123")
 	eng, teardown := prepStoreEngine(t)
 	defer teardown()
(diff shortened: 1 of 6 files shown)


backend/app/store/service/service_test.go
```diff
@@ -1279,7 +1277,9 @@ func TestService_submitImages(t *testing.T) {
 	lgr.Setup(lgr.Debug, lgr.CallerFile, lgr.CallerFunc)
 
 	mockStore := image.MockStore{}
-	mockStore.On("Commit", mock.Anything).Times(2).Return(nil)
+	mockStore.On("Commit", "dev/pic1.png").Once().Return(nil)
+	mockStore.On("Commit", "dev/pic2.png").Once().Return(nil)
+	mockStore.On("Commit", "cached_images/da39a3ee5e6b4b0d3255bfef95601890afd80709-da39a3ee5e6b4b0d3255bfef95601890afd80709").Times(2).Return(nil)
```
umputun, line 1282: worth to add mockStore.Assert to make sure all those Commits called

paskal: We have `mockStore.AssertNumberOfCalls(t, "Commit", 3)` few lines lower, and all three mock calls have specified params so the call with something else will panic.
What I missed here is `Times(2)` instead of `Once()` which breaks this arithmetic, I'll fix that.

paskal: Nvm, saw my error.

PR #703, opened 2020-04-25
Open hidden thread and jump to linked comment 

### LINKED ISSUE #594: Link to comment in collapsed thread doesn't jump to it
[Example link](https://radio-t.com/p/2020/02/04/prep-688/#remark42__comment-6e1a49f5-1724-48c5-b5e7-8fffc4fdaf01).

Result of going by the link:
image

It works until I collapse the parent thread:
image
->
image

After that opening the example link from above doesn't jump to the comment:
image

dellink:
Fix #594 

ezgif com-video-to-gif

### DIFF
--- a/frontend/app/components/root/root.tsx
+++ b/frontend/app/components/root/root.tsx
@@ -42,6 +42,13 @@ const mapStateToProps = (state: StoreState) => ({
   sort: state.comments.sort,
   isCommentsLoading: state.comments.isFetching,
   user: state.user,
+  childToParentComments: Object.entries(state.comments.childComments).reduce(
+    (accumulator: Record<string, string>, [key, children]) => {
+      children.forEach(child => (accumulator[child] = key));
+      return accumulator;
+    },
+    {}
+  ),
   topComments: state.comments.topComments,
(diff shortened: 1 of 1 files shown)


paskal:
@dellink is it possible to open hidden thread and jump to linked comment instead of jumping to it's parent?

frontend/app/components/root/root.tsx
```diff
@@ -85,6 +92,18 @@ const messages = defineMessages({
   },
 });
 
+const getVisibleParentComment = (hash: string, childToParentComments: Record<string, string>) => {
```
akellbl4, line 95: It looks like isn't the best way to find the parent of the hidden comment.
* because it trying to find parent every time even we don't have any collapsed comments (we can get a list of collapsed comments from redux)
* we can find a comment to which we trying to scroll without query to DOM, just from data representation of children comments

dellink:
> @dellink is it possible to open hidden thread and jump to linked comment instead of jumping to it's parent?

@paskal yes

frontend/app/components/root/root.tsx
dellink: We call this function only if we can't find comment in DOM. 
I don't understand how find uncollapsed parent for comment in a list of collapsed comments from redux, because we can collapse parent and save hash to grandchild.

PR #705, opened 2020-04-27
go-consistent proposed changes

paskal:
[go-consistent](https://github.com/quasilyte/go-consistent)-proposed changes.
I couldn't satisfy one of them (`./app/cmd/server_test.go:492:13: arg list parens: align `)` to a same line with last argument`) but all others are covered in this PR. I wonder if that makes any sense, maybe not fully but partially?

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -823,7 +823,7 @@ func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *
 		}
 	}
 
-	if len(destinations) != 0 {
+	if len(destinations) > 0 {
 		log.Printf("[INFO] make notify, types=%s", s.Notify.Type)
 		notifyService = notify.NewService(dataStore, s.Notify.QueueSize, destinations...)
 	}
--- a/backend/app/rest/api/rest_public_test.go
+++ b/backend/app/rest/api/rest_public_test.go
@@ -526,15 +526,15 @@ func TestRest_Config(t *testing.T) {
 	j := R.JSON{}
 	err := json.Unmarshal([]byte(body), &j)
 	assert.NoError(t, err)
-	assert.Equal(t, 300., j["edit_duration"])
+	assert.Equal(t, 300.0, j["edit_duration"])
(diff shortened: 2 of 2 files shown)


backend/app/cmd/server_test.go
```diff
@@ -489,9 +489,12 @@ func TestServerAuthHooks(t *testing.T) {
 	tkNoAud, err := tkService.Token(claimsNoAud)
 	require.NoError(t, err)
 	t.Logf("no-aud claims: %s", tkNoAud)
-	req, err = http.NewRequest("POST", fmt.Sprintf("http://localhost:%d/api/v1/comment", port),
+	req, err = http.NewRequest(
```
umputun, line 492: this makes very little sense, why split? the line is not that long

backend/app/migrator/mapper.go
```diff
@@ -37,7 +37,7 @@ func (u *URLMapper) loadRules(reader io.Reader) error {
 
 	rulesText := strings.TrimSpace(string(data))
 
-	u.rules = make(map[string]string)
+	u.rules = map[string]string{}
```
umputun, line 40: this one I don't really get either. make seems like a clear intention here and for a potentially dangerous thing like make I'd prefer make way

backend/app/store/comment.go
```diff
@@ -81,7 +81,7 @@ const snippetLen = 200
 func (c *Comment) PrepareUntrusted() {
 	c.ID = ""                 // don't allow user to define ID, force auto-gen
 	c.Timestamp = time.Time{} // reset time, force auto-gen
-	c.Votes = make(map[string]bool)
+	c.Votes = map[string]bool{}
```
umputun, line 84: same as above

backend/app/store/engine/bolt.go
```diff
@@ -422,7 +422,7 @@ func (b *BoltDB) Delete(req DeleteRequest) error {
 
 // Close boltdb store
 func (b *BoltDB) Close() error {
-	errs := new(multierror.Error)
+	errs := &multierror.Error{}
```
umputun, line 425: don't like it either, doesn't really make anything better, but up to you

PR #708, opened 2020-05-10
Switch from go-cache to lcw

paskal:
This gets rid of the direct dependency on go-cache. Leaking goroutines from it because of [go-pkgz/jrpc](https://github.com/go-pkgz/jrpc) imports [didip/tollbooth](https://github.com/didip/tollbooth/blob/master/limiter/limiter.go) which use go-cache.

I'll prepare PR to didip/tollbooth as well to get closer to eliminating go-cache leaking goroutines: for a start, I've created an issue https://github.com/didip/tollbooth/issues/84.

### DIFF
--- a/backend/app/store/service/service.go
+++ b/backend/app/store/service/service.go
@@ -9,10 +9,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-pkgz/lcw"
 	log "github.com/go-pkgz/lgr"
 	"github.com/google/uuid"
 	"github.com/hashicorp/go-multierror"
-	"github.com/patrickmn/go-cache"
 	"github.com/pkg/errors"
 
 	"github.com/umputun/remark/backend/app/store"
@@ -45,7 +45,7 @@ type DataStore struct {
 	}
 
 	repliesCache struct {
-		*cache.Cache
+		lcw.LoadingCache
 		once sync.Once
 	}
 }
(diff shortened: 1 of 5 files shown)


backend/app/store/service/service.go
```diff
@@ -493,7 +493,8 @@ func (s *DataStore) HasReplies(comment store.Comment) bool {
 	for _, c := range comments {
 		if c.ParentID != "" && !c.Deleted {
 			if c.ParentID == comment.ID {
-				s.repliesCache.Set(comment.ID, true, cache.DefaultExpiration)
+				// renew the cache
```
umputun, line 496: this is not exactly what it did before. I think you set nil to cause cache invalidation and make the next hit to get it? If so, this changes the logic and not the best way to do it. Probably calling Invalidate and get return true will be much closer.

_Generally speaking, this use case indicates the lack of somewhat useful Set method in lcw_

paskal: It's storing value in the cache, same as it did before. My change is to store `nil` instead of `true` as we don't care about the value here anyway.

umputun: well, adding a comment about it will be nice, because this logic is really not that obvious. Another thing - I fail to see how this code provides renew

paskal: Make sense, I rewrote the comment: please look into it once more.

backend/app/store/service/service.go
```diff
@@ -493,7 +493,10 @@ func (s *DataStore) HasReplies(comment store.Comment) bool {
 	for _, c := range comments {
 		if c.ParentID != "" && !c.Deleted {
 			if c.ParentID == comment.ID {
-				s.repliesCache.Set(comment.ID, true, cache.DefaultExpiration)
+				// When this code is reached, key "comment.ID" is not in cache.
+				// Calling cache.Get on it will put it in cache with 5 minutes TTL.
+				// We call it with nil value as we care about keys and not values.
+				_, _ = s.repliesCache.Get(comment.ID, func() (lcw.Value, error) { return nil, nil })
```
umputun, line 499: any particular reason why not to set `true` here? I understand we don't care about the value, but nil as a value feels odd. Btw, usually to indicate such "we don't care" `struct{}{}` used

paskal: Switched to `struct{}{}` as it's guaranteed to take 0 bytes. Both `nil` and `struct{}{}` are less confusing than `true` in my opinion.

Issue #722, opened 2020-05-25
Not SPA router friendly, even in history mode


patarapolw:
I have tried deploying on Nuxt. With a little tweaking, it works, but it throws Error on route change, except for the first route. 

Tweak --

```js
const remark_config = {
  host: 'https://remark.polv.cc', // hostname of remark server, same as REMARK_URL in backend config, e.g. "https://demo.remark42.com"
  site_id: 'polv',
  components: ['embed'], 
  locale: 'en', // set up locale and language, if it isn't defined default value ('en') will be used,
  init() {
    ;(function(c) {
      for (let i = 0; i < c.length; i++) {
        const d = document
        const s = d.createElement('script')
        s.className = 'remark42-script'
        s.src = remark_config.host + '/web/' + c[i] + '.js'
        s.defer = true
        ;(d.head || d.body).appendChild(s)
      }
    })(remark_config.components || ['embed'])
  },
  remove() {
    document.querySelectorAll('iframe[title=Remark42]').forEach((el) => {
      el.remove()
    })
    document.querySelectorAll('script.remark42-script').forEach((el) => {
      el.remove()
    })
  }
}
window.remark_config = remark_config
```

Error,

```
Uncaught TypeError: Cannot read property 'postMessage' of null
    at l (embed.js:1)
    at MutationObserver.<anonymous> (embed.js:1)
l @ embed.js:1
(anonymous) @ embed.js:1
childList (async)
updateTitle @ vue-meta.esm.browser.js?2fd6:1065
updateClientMetaInfo @ vue-meta.esm.browser.js?2fd6:1260
refresh @ vue-meta.esm.browser.js?2fd6:1437
refresh$1 @ vue-meta.esm.browser.js?2fd6:1504
eval @ vue-meta.esm.browser.js?2fd6:169
eval @ vue-meta.esm.browser.js?2fd6:191
setTimeout (async)
batchUpdate @ vue-meta.esm.browser.js?2fd6:190
triggerUpdate @ vue-meta.esm.browser.js?2fd6:168
eval @ vue-meta.esm.browser.js?2fd6:477
setInterval (async)
eval @ vue-meta.esm.browser.js?2fd6:470
eval @ vue.runtime.esm.js?2b0e:1980
flushCallbacks @ vue.runtime.esm.js?2b0e:1906
Promise.then (async)
timerFunc @ vue.runtime.esm.js?2b0e:1933
nextTick @ vue.runtime.esm.js?2b0e:1990
[...]

PR #723, opened 2020-05-25
Fix to work without errors in Nuxt Universal

patarapolw:
As a solution to https://github.com/umputun/remark42/issues/722 I cloned and edited the repo.

- In `embed.ts`,
  - Replace `!.` with `?.` so it will not throw error ... of undefined.
  - Add `destroy()` to `window.Remark42`, which actually removes global listeners (on `window` and `document`)
  - Prevent duplication of Remark42 Iframe with

```js
  const oldIframe = document.querySelector('iframe[title=Remark42]') as HTMLIFrameElement;
  if (oldIframe) {
    return oldIframe;
  }
```

Also, in production, instead of your init function <https://github.com/umputun/remark42#comments>, I moved it inside a function and add a typing for TypeScript. -- [remark42.ts](https://github.com/patarapolw/polv/blob/6ac9035a030b4ed4492e1613e60ab9183db7af53/assets/remark42.ts#L1) and it is implemented here -- [PostFull.vue](https://github.com/patarapolw/polv/blob/6ac9035a030b4ed4492e1613e60ab9183db7af53/components/PostFull.vue#L52)

In order to test in production, I deploy it to Google Compute Engine and Docker Hub with `patarapolw/remark42`, as can be seen at

- https://remark42.polv.cc/web -- Remark42 hosting
- https://www.polv.cc -- Nuxt with Remark42


umputun:
I'm not really sure what we supposed to see, but both links from the message above don't work. One returns "not found" (maybe you meant https://remark.polv.cc/web/ ?) and the second "Error: Server Error"

patarapolw:
@umputun Sorry, the website was down for some reasons (Google App Engine throws no log.) But it is back now.

paskal:
Gentle ping for @patarapolw to change the code after the review.

frontend/app/embed.ts
```diff
@@ -25,6 +25,11 @@ function createFrame({
   height?: string;
   __colors__?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
 }) {
+  const oldIframe = document.querySelector('iframe[title=Remark42]') as HTMLIFrameElement;
```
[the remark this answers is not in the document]
patarapolw: How do you recommend I check for previous existence of Iframe?

frontend/app/embed.ts
```diff
@@ -203,9 +214,9 @@ function init() {
       document.body.appendChild(this.node);
       document.addEventListener('keydown', this.onKeyDown);
       setTimeout(() => {
```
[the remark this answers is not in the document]
patarapolw: Ok, but where to save `timeoutId`?

frontend/app/embed.ts
```diff
@@ -203,9 +214,9 @@ function init() {
...
-        this.back!.setAttribute('data-animation', '');
-        this.node!.setAttribute('data-animation', '');
-        this.iframe!.focus();
+        this.back?.setAttribute('data-animation', '');
```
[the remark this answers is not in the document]
patarapolw: I see, so it would probably become,

```ts
      setTimeout(() => {
        this.back!.setAttribute('data-animation', '');
        this.node!.setAttribute('data-animation', '');
        this.iframe!.focus();
      }, 400);
```

patarapolw:
Do you need some kind of commitiquette, like commitizen or commitlint?

akellbl4:
LGTM, but I would ask @Mavrin to revise.

patarapolw:
BTW, does not work in GatsbyJS, but OK in NextJS (and of course NuxtJS).

A reproducible repo is here -- https://github.com/patarapolw/jpdiary/tree/gatsby

Mavrin:
@patarapolw. Thank you for reproducible example.  Did you read my comment?


> For SPA we could create api like:
> 
> ```js
> const instance = window.REMARK42.createInstance({
>   host: "host",
>   site_id: "siteId",
>   url: "url",
>   max_shown_comments: 50,
>   theme: "light",
>   page_title: "test",
>   node: DOMNode,
>   locale: "en",
> });
> // instance.changeTheme();
> // instance.destroy();
> ```

I think if you impent this all should work as expected. You need to call this code when page is mount and unmount.

patarapolw:
> @patarapolw. Thank you for reproducible example. Did you read my comment?
> 
> > For SPA we could create api like:
> > ```js
> > const instance = window.REMARK42.createInstance({
> >   host: "host",
> >   site_id: "siteId",
> >   url: "url",
> >   max_shown_comments: 50,
> >   theme: "light",
> >   page_title: "test",
> >   node: DOMNode,
> >   locale: "en",
> > });
> > // instance.changeTheme();
> > // instance.destroy();
> > ```
> 
> I think if you impent this all should work as expected. You need to call this code when page is mount and unmount.

So, I need to create a class call `Remark42Instance`? Do I need to create a new file?

Mavrin:
You can run all test by `npx run-p check lint`.
I will look PR today.

frontend/app/embed.ts
```diff
@@ -47,85 +48,98 @@ function createFrame({
...
+      window.REMARK42.destroy = remark42Instance.destroy;
+    }
   }
 
-  if (!window.remark_config.site_id) {
-    console.error('Remark42: Site ID is undefined.');
-    return;
-  }
+  function createInstance(remark_config: CommentsConfig) {
+    let initDataAnimationTimeout: NodeJS.Timeout | null = null;
```
Mavrin: should be `let initDataAnimationTimeout: number | null = null`; https://stackoverflow.com/a/50428377/13329651

frontend/app/embed.ts
```diff
@@ -174,136 +188,156 @@ function init() {
...
+          (document.querySelector('iframe[title=Remark42]') as HTMLIFrameElement) ||
+          createFrame({ host: BASE_URL, query: queryUserInfo, height: '100%' });
+        this.node.appendChild(iframe);
+        this.iframe = iframe;
+        this.node.appendChild(this.closeEl);
+        document.body.appendChild(this.style);
+        document.body.appendChild(this.back);
+        document.body.appendChild(this.node);
+        document.addEventListener('keydown', this.onKeyDown);
+        initDataAnimationTimeout = setTimeout(() => {
```
Mavrin: Please use window.setTimeout  https://stackoverflow.com/a/50428377/13329651

frontend/app/embed.ts
```diff
@@ -47,85 +48,98 @@ function createFrame({
...
-    });
-  }
+    const query = Object.keys(remark_config)
+      .filter(key => key !== '__colors__')
+      .map(key => {
+        return `${encodeURIComponent(key)}=${encodeURIComponent(remark_config[key as keyof typeof remark_config])}`;
+      })
+      .join('&');
+    const iframe =
+      (document.querySelector('iframe[title=Remark42]') as HTMLIFrameElement) ||
```
Mavrin: You don't need to check that frame already exist. You could do it in your integration code and don't create new instance with the same config.

frontend/app/embed.ts
patarapolw: Screen Shot 2563-06-24 at 18 30 29

frontend/app/embed.ts
```diff
@@ -246,7 +260,7 @@ function init() {
         clearTimeout(t.delay);
         t.delay = null;
       }
-      t.events.forEach(event => t.node!.removeEventListener(event, t.animationStop, false));
+      t.events.forEach(event => t.node?.removeEventListener(event, t.animationStop, false));
```
Mavrin, line 263: Pease revert `?`. After migrate `createInstance` api. t.node always exist and not equal `null`
l

patarapolw:
@Mavrin OK, but in theory, I am not sure when `iframe.contentWindow` will be `null`?

Mavrin:
> @Mavrin OK, but in theory, I am not sure when `iframe.contentWindow` will be `null`?

If it is possible. We should add this check. Could you describe case when it is null?

Mavrin:
Lgtm.
Did you try run current solution on your example?
Thank you.

akellbl4:
@patarapolw good job, would you be so kind to add documentation for this feature? And yes, as @Mavrin asked, it would be great to check working example.

patarapolw:
For the example, where I really used, see my website -- <https://www.polv.cc>.

I also made a much simpler example -- https://github.com/patarapolw/remark42-spa (will update in there.)

frontend/app/embed.ts
```diff
@@ -329,9 +326,22 @@ function createInstance(config: CommentsConfig) {
...
     document.removeEventListener('click', postClickOutsideToIframe);
+
+    if (titleObserver) {
+      titleObserver.disconnect();
+    }
+
     iframe.remove();
   }
 
+  // TODO: These do not appear in Chrome DevTools
```
patarapolw, line 337: I have this unresolved problem. Not really a problem to me, but may affect backward compatibility...

akellbl4: As I see. API appears  in DevTools
Screenshot 2020-07-11 at 01 14 37

gvlasov:
@patarapolw @umputun Iframe reusing introduced in this PR breaks the possibility of having multiple comment sections on the same page. In this PR, iframe reuse is said to be for "Preventing duplication of Remark42 Iframe", but that definitely shouldn't be remark42's responsibility.

With current `master`, all that is needed to fix remark42 for multiple comment sections is changing

```ts
  const iframe =
    (document.querySelector('iframe[title=Remark42]') as HTMLIFrameElement) ||
    createFrame({ host: BASE_URL, query, __colors__: config.__colors__ });

```
to
```ts
  const iframe =
    createFrame({ host: BASE_URL, query, __colors__: config.__colors__ });
```

gvlasov:
A small correction: that's definintely not the only thing preventing multiple comment sections from working on the same page, but changing that is required if multiple comment sections are ever to be implemented (I'm probably going to be working on it soon and I will provide a PR)

Issue #728, opened 2020-06-08
Same IP check prevents "changed my mind" voting 


umputun:
In case if same-ip voting check enabled the correctional voting (i.e. `+, -, -`) rejected with `the same ip already voted`. This is incorrect and without same-ip checks works fine.

/re: #400

PR #729, opened 2020-06-08
Addresses the bug preventing correctional voting if same-ip check turned on

umputun:
see #728


umputun:
@paskal pls take a look. The change should be innocent, new and old tests are passing, but just in case

umputun:
this is the original bug report: https://radio-t.com/p/2020/06/06/podcast-705/#remark42__comment-e73a959e-08d1-4f8b-8cd4-5fa2e23e6570

PR #736, opened 2020-06-14
add distributed cache support via redis

paskal:
Use new `eventbus.NewRedisPubSub` to enable distributed cache support in remark42.

Tests are written only for the failure of new cache creation. Testing its work would require setting up a redis and doing black-box cache testing which might be complex but I could do it if it's necessary.


umputun:
All of this makes sense but causes some confusion because cache supports redis for real as well. The type `redis` is not really redis cache but rather distributed local cache with redis pub-sub.

I'm not sure what the right terminology here but the current one is confusing for sure

Issue #731, opened 2020-06-10
Is it possible to enabled admin notifications (about new comments), but without email subscriptions for other users?


ajgon:
Hey!

So I have a problem, like described in the question.

I'm hosting my own email server, and I don't want it to send emails to (or from) anybody but me. 

After quick looking to the code, I noticed that enabling `NOTIFY_TYPE` will enable both admin and registered user notifications. Is it possible to keep only the former?

patarapolw:
I currently cannot let user subscribe for some reasons, and my settings are

```yaml
      - AUTH_EMAIL_FROM=polv@polv.cc
      - AUTH_EMAIL_ENABLE=true
      - NOTIFY_EMAIL_FROM=polv@polv.cc
      - NOTIFY_EMAIL_ADMIN=true
      - ADMIN_SHARED_EMAIL=polv@polv.cc
      - NOTIFY_TYPE=email
```

I still receive my needed emails as an admin, though.

paskal:
@patarapolw if admin notifications work, then following would work:

1. Subscribe as user A, leave a comment
2. Log in as user B, leave a reply to comment from user A

Then you'll receive a notification about new reply to your comment as user A. Replying to your own comment won't produce notification. Please let me know if that doesn't work as I described.

paskal:
@ajgon my apologies for the delay with the answer. Currently, email notifications enable site email subscription as well but I do see value in adding a separate option for that.

I'll add an option like "disable email notifications for users" which would disable the frontend part of the subscription. Disabling API might be harder but I'll try to do it as well to have a proper solution.

paskal:
@ajgon we decided to move this configuration option to frontend configuration which makes more sense and also gives the ability to customize this behavior per-site in a multi-site setup. Passing to @akellbl4 for implementation.

PR #738, opened 2020-06-14
Add ability to disable user email notification UI

paskal:
Fix for #731. As stated in the documentation, it adds a new option to disable UI for email notifications when notifications themselves are enabled. API left enabled because it's enabled unconditionally now, so there is no change in that part.

This is necessary for the enablement of admin email notifications only, without this PR there is no workaround for it.

Truly disabling notifications other than for admin would require more significant changes and I don't know if it's necessary, leaving it for your judgment.


umputun:
I think disabling UI is good enough, but having some issues with the proposed implementation:

1. I don't like "disable" kind of parameter, due to confusion it causes, i.e "disable.blah:true" is odd. I'd  suggest renaming the param to be "enabled-style", i.e. NOTIFY_EMAIL_ADMIN_ONLY or smth like this
1. All it does is setting a flag for UI. Maybe moving this flag directly to UI settings will be a better idea? In addition to "moving to the right place", it has one more benefit - ability do enable/disable per site (in multi-site configurations)

paskal:
Per off-GitHub talk, we decided to move this configuration option to frontend configuration which makes more sense and also give ability to customize this behavior per-site in multi-site setup. Closing this PR.

Issue #737, opened 2020-06-14
README on how I really did development


patarapolw:
Shouldn't it be told to do something like this?

```sh
# if you mainly work on backend
cp compose-dev-backend.yml compose-private-backend.yml
# if you mainly work on frontend
cp compose-dev-frontend.yml compose-private-frontend.yml

# And edit / debug `compose-private-*.yml` to your heart's content.
```

This is just because `compose-private-*.yml` are gitignored. As a matter of fact, my `compose-private-frontend.yml` contains this line.

```yaml
services:
    remark42:
        ...
        image: patarapolw/remark42:rev1
```

And, I pushed the image to Docker Hub, for deployment on VPS, to test that it really works.

paskal:
Could you please prepare a PR with the change? As far as I see it it would be the easiest way to ask a questions which is better, old or new version.

patarapolw:
I still have one PR pending. How do I manage another one? Just overwrite the old PR?

paskal:
1. Create a new branch in your fork and switch to it
2. In this new branch roll back all commits which are made on top of original master using command `git reset --hard HEAD^1` as many times as necessary
3. Make your changes and commit them, then send a PR from this new branch.

The second step would be unnecessary if you would next time use branch other than `master` for the PR in the first place, I would advise having `master` in sync with `upstream/master` for that purpose. [Here](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) is a documentation on how to do that.

patarapolw:
I see, but how do I change from

```
	patarapolw wants to merge 2 commits into umputun:master from patarapolw:master
```

to

```
	patarapolw wants to merge 2 commits into umputun:master from patarapolw:spa-and-nuxt
```

or I cannot?

patarapolw:
> I still have one PR pending. How do I manage another one? Just overwrite the old PR?

What I actually did,

```sh
git remote add upstream git@github.com:umputun/remark42.git
git fetch upstream
git switch -c readme-for-devs upstream/origin 
```

The actual steps are a little different as I used VSCode's GUI.

PR #739, opened 2020-06-14
update README according to #737

patarapolw:
According to https://github.com/umputun/remark42/issues/737, actually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -562,11 +562,14 @@ To bring it up run:
 
 ```bash
 # if you mainly work on backend
-docker-compose -f compose-dev-backend.yml build
-docker-compose -f compose-dev-backend.yml up
+cp compose-dev-backend.yml compose-private.yml
 # if you mainly work on frontend
-docker-compose -f compose-dev-frontend.yml build
-docker-compose -f compose-dev-frontend.yml up
+cp compose-dev-frontend.yml compose-private.yml
+# now, edit / debug `compose-private.yml` to your heart's content.
+
+# build and run
+docker-compose -f compose-private.yml build
+docker-compose -f compose-private.yml up
 ```
 
 It starts Remark42 on `127.0.0.1:8080` and adds local OAuth2 provider “Dev”.


README.md
```diff
@@ -562,11 +562,14 @@ To bring it up run:
...
-docker-compose -f compose-dev-backend.yml up
+cp compose-dev-backend.yml compose-private.yml
 # if you mainly work on frontend
-docker-compose -f compose-dev-frontend.yml build
-docker-compose -f compose-dev-frontend.yml up
+cp compose-dev-frontend.yml compose-private.yml
+# now, edit / debug `compose-private.yml` to your heart's content.
+
+# build and run
+docker-compose -f compose-private.yml build
```
paskal, line 571: I would propose to leave `docker-compose build` commands near `cp` instead of moving to separate section. Otherwise, make sense!

paskal:
I thought about it some more and now I see that there might be some misunderstanding which you might help to figure out using the documentation.

The section you are changing is supposed to run as-is, with generic compose files. If you want to customize docker-compose file and run your own development code on top of it, you should copy it to `compose-private.yml` and run `make rundev` in order to build new dev image and run remark42 locally from it. Details are in the makefile:

```make
backend:
	docker-compose -f compose-dev-backend.yml build

frontend:
	docker-compose -f compose-dev-frontend.yml build

rundev:
	docker pull umputun/baseimage:buildgo-latest
	SKIP_BACKEND_TEST=true SKIP_FRONTEND_TEST=true docker-compose -f compose-private.yml build
	docker-compose -f compose-private.yml up
```

Now I see that makefile is not reflected in the Readme at all, It would be awesome if you'll test it and update the instruction in a way you understand it. Thank you for raising it!

patarapolw:
I would want to somehow edit `docker-compose.yml`, without worrying about tampering with PR. That's why I prefer gitignored file.

I understand about `make rundev`. Seems to have things I need. However, I am not sure how to add to `README.md`...

paskal:
That's how frontend development described currently:

```shell
cp compose-dev-frontend.yml compose-private-frontend.yml
docker-compose -f compose-private-frontend.yml up --build
cd frontend
npm run dev
```

I think, similar section with `make rundev` might be added to backend development guide section. And the part which you changed now could potentially be left unaltered or not, based on how you see it fit the whole document and new section for the backend.

vdimir:
Also maybe adding `compose-private.yml` to `.gitignore` is an option?

umputun:
it is in .gitignore already, unless I missed your point https://github.com/umputun/remark42/blob/master/.gitignore#L25

vdimir:
> it is in .gitignore already, unless I missed your point https://github.com/umputun/remark42/blob/master/.gitignore#L25

Yes, sorry, I haven't check file before writing comment.

PR #744, opened 2020-07-05
Fix stuck in image submit while importing comments from backup

vdimir:
I tried to import batch of comments via native importer. I notices that importing frozen at 5000 comment, due to size of buffered channel in image service. It was continued after ttl (1.5  edit time) had expired, and stuck again.
So, this PR contains possible solution -- using not current time to check comment edit time expiration before indexing images, but actual comment timestamp.
I am not sure that it is the best solution and maybe some other needs to be considered. For example non-blocking version of submit, of forcing submition for importer.

### DIFF
--- a/backend/app/migrator/disqus.go
+++ b/backend/app/migrator/disqus.go
@@ -135,6 +135,7 @@ func (d *Disqus) convert(r io.Reader, siteID string) (ch chan store.Comment) {
 						Text:      d.cleanText(comment.Message),
 						Timestamp: comment.CreatedAt,
 						ParentID:  comment.Pid.Val,
+						Imported:  true,
 					}
 					if c.User.ID == "disqus_" { // empty comment.AuthorUserName from disqus
 						c.User.ID = "disqus_" + c.User.Name
--- a/backend/app/store/image/image.go
+++ b/backend/app/store/image/image.go
@@ -102,7 +102,7 @@ func NewService(s Store, p ServiceParams) *Service {
 }
 
 // Submit multiple ids via function for delayed commit
-func (s *Service) Submit(idsFn func() []string) {
+func (s *Service) Submit(idsFn func() []string, ts time.Time) {
 	if idsFn == nil || s == nil {
 		return
 	}
@@ -130,7 +130,12 @@ func (s *Service) Submit(idsFn func() []string) {
 	})
 
 	atomic.AddInt32(&s.submitCount, 1)
-	s.submitCh <- submitReq{idsFn: idsFn, TS: time.Now()}
+
+	now := time.Now()
+	if ts.IsZero() || ts.After(now) {
+		ts = now
+	}
+	s.submitCh <- submitReq{idsFn: idsFn, TS: ts}
 }
 
 // ExtractPictures gets list of images from the doc html and convert from urls to ids, i.e. user/pic.png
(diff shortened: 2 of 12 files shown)


vdimir:
I noticed that CI in my repo failed https://github.com/vdimir/remark/runs/839113674
Is it flapping test?

paskal:
@vdimir I think it is.

umputun:
> I am not sure that it is the best solution and maybe some other needs to be considered. For example non-blocking version of submit, of forcing submition for importer.

I think the forced image submission is the right solution here. In case of import, we don't need two-stage submit/commit logic. However, I'm not sure what the best way to do it. One of the solutions I was thinking abouy is to add an "Imported" field to the `Comment` struct and for such comments perform some `image.SubmitAndCommit`,  saving images right away. Generally speaking, having `comment.Imported` field can be useful in some other cases too.

paskal:
Just in case it is a blocker for someone, I think the workaround would be setting lower `--edit-time` which is `5m` by default, leading to commits every `5m * 1.5 = 7.5m`:

https://github.com/umputun/remark42/blob/9510962e2521ba77d4391190116c546c83f08db3/backend/app/store/image/image.go#L96-L102

backend/app/migrator/disqus.go
```diff
@@ -135,6 +135,7 @@ func (d *Disqus) convert(r io.Reader, siteID string) (ch chan store.Comment) {
 						Text:      d.cleanText(comment.Message),
 						Timestamp: comment.CreatedAt,
 						ParentID:  comment.Pid.Val,
+						Imported:  true,
```
umputun, line 138: will be nice to add a check for Imported field to test(s)

backend/app/store/image/image.go
```diff
@@ -101,6 +101,15 @@ func NewService(s Store, p ServiceParams) *Service {
 	return &Service{ServiceParams: p, store: s}
 }
 
+// SubmitAndCommit multiple ids immediately
+func (s *Service) SubmitAndCommit(idsFn func() []string) {
```
umputun, line 105: this code inside a function now and can return an error, instead of logging. And whoever calls it could decide what to do with that error

backend/app/store/service/service.go
```diff
@@ -101,7 +101,8 @@ func (s *DataStore) Create(comment store.Comment) (commentID string, err error)
 		comment.PostTitle = title
 	}()
 
-	s.submitImages(comment.Locator, comment.ID)
+	defer s.submitImages(comment) // submit images only after creation in Engine
```
umputun, line 104: I don't really like defer use for this. Probably just adding it directly after Create call will be easier to read and understand.

backend/app/store/image/image.go
vdimir: Should `Submit` return error too? Now `SubmitAndCommit` return error, but `Submit` not and it looks a bit inconsistent, especially here. https://github.com/umputun/remark42/pull/744/files#diff-0c9477a43c3f506684e6e23b4ce2e918R247-R256

But what error can return `Submit`, always nil for now?

umputun: Submit just put things to a channel. As long as we have no context here to interrupt it I see no reason to return any error because it has no path returning any error.

umputun:
hi. I see you pushed some commits. Let me know if the code is ready for review

vdimir:
> hi. I see you pushed some commits. Let me know if the code is ready for review

Yes, I resolved all your comments to code, thanks for it. Failed CI is flapping test I guess (in branch in my fork and locally it is ok)

PR #761, opened 2020-08-18
Run FE specs on CI

pustovalov:
Is there any reason to run FE specs on local ENV instead of run them on CI?


akellbl4:
@pustovalov agree with you

akellbl4 (approved):

Issue #756, opened 2020-07-31
Multiple admin emails


dkornilove:
Hello! I have a couple of managers that should be notified about all the new comments via email. But as I see I can set only one admin email at the ADMIN_SHARED_EMAIL variable. Is there any way to configure multiple admin emails? Thanks

umputun:
Currently, there is no way to define multiple admin emails. This is something I'd like to add, but for now you have two workarounds:

- Create a "group email" if your email service support it (google's gsuite can do it) 
- Use telegram notifications as it pushes updates to groups

umputun:
@paskal let me know if you want to deal with this one. It seems to be a trivial change:

1. Change AdminGroup.Email (both shared and RPC) to the slice / repeated
1. Change rest.AdminEmail to slice
1. Change notify.Request to support slice of emails  or send one-by-one from `createCommentCtrl`

paskal:
ETA ~1 week.

paskal:
@dkornilove my apologies, it took longer than I anticipated. After linked MR will be merged you'll have ability to set up multiple emails for notifications.

dkornilove:
@paskal , @umputun 
Thanks for the attention! But sadly I am not able to get it work with this env variable defined:
`- ADMIN_SHARED_EMAIL=exam.ple@mail1.com,example2@mail2.ru`
Emails are not sent. Only this working for me
`- ADMIN_SHARED_EMAIL=example@mail.com`
Im using last master version

paskal:
@dkornilove, the first variant should work. Would you be so kind as to post docker-compose log around the time the message supposed to be sent to these emails? There should be something from notification part of the program at least trying to send them.

PR #770, opened 2020-09-02
add ability to set multiple admin emails for notifications

paskal:
This is a fix for #756, after this change user would be able to set multiple admin emails and all of them will be used for admin notifications.

That's a simple approach to this problem, a more sophisticated one would be introducing admins per-site, who will receive notifications for their sites. It might be useful for single remark42 instance with multiple different sites, but on the other hand, I think that we might discourage that pattern and force users to run multiple instances of remark42 in separate docker containers if they want to host completely different sites together.


umputun (approved):
LGTM, thx

PR #772, opened 2020-09-06
Setup coverage for FE

pustovalov:
Example:
https://codecov.io/gh/pustovalov/remark42/branch/fe-cov-1

To do:
setup `CODECOV_TOKEN` to gh-actions ENV variables, link https://codecov.io/gh/umputun/remark42/settings

### DIFF
--- a/.github/workflows/ci-test-frontend.yml
+++ b/.github/workflows/ci-test-frontend.yml
@@ -71,5 +71,10 @@ jobs:
       - run: npm ci --loglevel warn
         working-directory: ./frontend
 
-      - run: npm run test
+      - run: npm run test:coverage
         working-directory: ./frontend
+
+      - name: submit coverage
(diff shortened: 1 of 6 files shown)


paskal (approved):

umputun:
This is nice. The only question I have - is there a reason to use a different service, i.e. codecov instead of coverals? The backend publishes reports to coverals and, if possible, it will be nice to have the frontend reports on the same service. I have no clue how good support for JS in coverals and how to publish the second report for the same repo

pustovalov:
a couple of reasons:
1. easier to integrate
2. do not change the coverage for BE

I first tried to integrate coveralls for FE, but the first attempts were unsuccessful:
https://coveralls.io/repos/311880/builds

If you prefer one service, I can try to integrate coveralls for FE again

umputun:
nm, this is fine

PR #795, opened 2020-10-18
Recursive email notifications

### LINKED ISSUE #758: Create recursive email notifications to parent comment about all children
A follow-up for #745 involved parties agreed upon, we want to implement email notifications to be sent not only about direct replies to the user's comment but about any replies to them as well as deep as it goes.

- 1
  - 2
    - 3
      - 4

In the scenario above, user 1 would receive notification for comment 2 (but should 2-4), user 2 would receive notification for comment 3 (should be 3-4), user 3 would receive notification for comment 4 (will not be changed).

paskal:
Send email notifications recursively up the chain of comments. Changes the default behavior.

Relies on merging #796 first. Will resolve #758.

### DIFF
--- a/backend/app/notify/notify.go
+++ b/backend/app/notify/notify.go
@@ -108,6 +108,11 @@ func (s *Service) getNotificationEmails(req Request, notifyComment store.Comment
 			result = append(result, email)
 		}
 	}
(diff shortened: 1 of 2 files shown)


umputun:
LGTM

PR #796, opened 2020-10-18
Clarify notify.Email implementation and notify logic

paskal:
This PR moves more email notification logic from `notify.Submit` callers and `notify.Submit` itself into `Email.Send`. In particular, it starts handling user and admin email notifications in a single `Email.Send` to hide this logic from `notify.Submit`. As a tradeoff, I had to move email user verification logic to `notify.getNotificationEmails`, as I'll need it there for recursive notifications.

The new logic flow is extensively tested with existing tests as well as the newly created one.

Also, this PR hides user email from error and logs messages about canceled context and adds a test for covering default template settings.

### DIFF
--- a/backend/app/notify/email.go
+++ b/backend/app/notify/email.go
@@ -177,43 +177,31 @@ func (e *Email) Send(ctx context.Context, req Request) error {
 
 	result := new(multierror.Error)
 
-	// send user notifications
 	for _, email := range req.Emails {
-		email := email
-		log.Printf("[DEBUG] send user notification via %s, comment id %s", e, req.Comment.ID)
-		msg, err := e.buildMessageFromRequest(req, email, false)
-		if err != nil {
-			return err
-		}
-
-		err = repeater.NewDefault(5, time.Millisecond*250).Do(
-			ctx,
-			func() error {
-				return e.sendMessage(emailMessage{from: e.From, to: email, message: msg})
-			})
-
+		err := e.buildAndSendMessage(ctx, req, email, false)
 		result = multierror.Append(errors.Wrapf(err, "problem sending user email notification to %q", email))
 	}
 
-	// send admin notifications
 	for _, email := range req.AdminEmails {
-		email := email
-		log.Printf("[DEBUG] send admin notification via %s, comment id %s", e, req.Comment.ID)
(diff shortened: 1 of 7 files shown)


backend/app/notify/email.go
```diff
@@ -164,35 +165,55 @@ func (e *Email) setTemplates() error {
...
 	}
 
-	if req.parent.User.ID == req.Comment.User.ID && !req.ForAdmin {
-		// don't send anything if if user replied to their own comment
-		return nil
+	result := new(multierror.Error)
+
+	// send user notifications
+	for _, email := range req.Emails {
+		email := email
```
umputun, line 182: unless I missed something I see no reason for this as the flow with `email` seems to be synced, not  captured for goroutine

backend/app/notify/email.go
```diff
@@ -164,35 +165,55 @@ func (e *Email) setTemplates() error {
...
+		result = multierror.Append(errors.Wrapf(err, "problem sending user email notification to %q", email))
 	}
-	log.Printf("[DEBUG] send notification via %s, comment id %s", e, req.Comment.ID)
-	msg, err := e.buildMessageFromRequest(req, req.ForAdmin)
-	if err != nil {
-		return err
+
+	// send admin notifications
+	for _, email := range req.AdminEmails {
+		email := email
```
umputun, line 200: same as above

backend/app/notify/notify.go
```diff
@@ -37,12 +37,12 @@ type Store interface {
 	GetUserEmail(siteID string, userID string) (string, error)
 }
 
-// Request notification either about comment or about particular user verification
+// Request notification either about comment
```
umputun, line 40: either seems to be leftover

backend/app/notify/email.go
```diff
@@ -164,35 +165,55 @@ func (e *Email) setTemplates() error {
...
+		log.Printf("[DEBUG] send admin notification via %s, comment id %s", e, req.Comment.ID)
```
umputun, line 201: two blocks seens to be identical, except the last param (forAdmin). Maybe extract to an inner func?

umputun (changes requested):
minor things

umputun (approved):
LGTM

PR #805, opened 2020-10-28
added translations to pl language

markiewiczjulian:
This is the translation for the Polish language. Below I attach the screens with a translation in action.

PS. This is my first pull request for open source, so if anything is not right please let me know, I will be more than happy to help.

Screenshot from 2020-10-28 09-04-31
Screenshot from 2020-10-28 09-05-12
Screenshot from 2020-10-28 09-06-44
Screenshot from 2020-10-28 09-07-36
Screenshot from 2020-10-28 09-07-46
Screenshot from 2020-10-28 09-12-59


akellbl4:
My comment isn't related to PR but what browser do you use? I can notice a white border inside the gray area and it looks like a bug.

mind-overflow:
> My comment isn't related to PR but what browser do you use? I can notice a white border inside the gray area and it looks like a bug.

It's either Chrome or Chromium. You can see it from the UI in one of the screenshots. Very weird...

markiewiczjulian:
Sorry for the late response. I was pretty busy.
About the white line. I've investigated a little today and this seems to happen on all resolutions on my pc. @mind-overflow guess is correct this is Chrome (which...btw is up to date).
image
 
I see that the top white line disapears when you lower the top margin for this selector (change from 4px to 3px) `#remark42 .root__input`. So this should look like that: 
```
#remark42 .root__input {
    margin-top: 3px;
} 
```

Also the lines on right and left dissapear when you change the border-width for this selector `#remark42 .comment-form` so it will look like that (I've changed the left right border-width from 12px to 13px and top one from 6px to 7px):
```
#remark42 .comment-form {
    position: relative;
    display: block;
    border-style: solid;
    border-width: 7px 13px 12px;
    border-radius: 2px;
}
```
After these changes it looks fine:
image

markiewiczjulian:
Conflicts resolved :+1:

Mavrin (approved):

umputun (approved):
Thx

Issue #567, opened 2020-01-18
Demo page won't work If custom SITE variable was used 


akellbl4:
`site_id` (it is named just `site` on backend) was hardcoded in the template `frontend/index.ejs` and when you change it in `docker-compose.yml`it won't work with backend because another `site` was defined.

Example: https://remark42.mineev.me

akellbl4:
@umputun we can pass env variable to the template but only on build because on a start we already have built static files.
Maybe we can change `site` in the template on start by `sed`.

paskal:
No sed, please. Env variable?

akellbl4:
> Env variable?
As I see only on the build. A static HTML can't able to get any var from any place.

@Mavrin maybe you have some fresh ideas connected with it?

paskal:
Fixed by 06c3406: `docker-init.sh` now rewrites the hardcoded `site_id` in the bundled HTML at container start, using the first ID from the comma-separated `SITE` list. Released in v1.7.0.

PR #806, opened 2020-11-02
Fixed build for usage without subdomain

sattellite:
I installed remark42 in the easiest way from the docker. Configured nginx to use app without subdomain. It works okay for comments but widgets not working as expected. Widgets use it own config with host as window.location.origin which returns the fqdn only.

This can be fixed on server side by multiple sed commands in container like this `docker exec -it remark42 sed -i "s|host: window.location.origin|host: 'https://b.sattellite.me/comments/'|g" /srv/web/last-comments.html`

I prepared fix to code for exclude this problem in future:

  - Replaced host value from window.location.origin to "REMARK_URL" || window.location.origin.
  - Changed init script for container to replace forced `site_id` by 'remark' to environment SITE_ID.

### DIFF
--- a/docker-init.sh
+++ b/docker-init.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 echo "prepare environment"
 # replace BASE_URL constant by REMARK_URL
-sed -i "s|https://demo.remark42.com|${REMARK_URL}|g" /srv/web/*.js
+sed -i "s|https://demo.remark42.com|${REMARK_URL}|g" /srv/web/*.{js,html}
 
 if [ -d "/srv/var" ]; then
   chown -R app:app /srv/var 2>/dev/null
--- a/docker-init.sh
+++ b/docker-init.sh
@@ -3,6 +3,11 @@ echo "prepare environment"
 # replace BASE_URL constant by REMARK_URL
 sed -i "s|https://demo.remark42.com|${REMARK_URL}|g" /srv/web/*.{js,html}
 
+if [ -n "${SITE_ID}" ]; then
+  #replace "site_id: 'remark'" by SITE_ID
+  se -i "s|'remark'|'${SITE_ID}'|g" /srv/web/*.html
+fi
+
 if [ -d "/srv/var" ]; then
   chown -R app:app /srv/var 2>/dev/null
 else
--- a/frontend/last-comments.ejs
+++ b/frontend/last-comments.ejs
@@ -25,7 +25,7 @@
     <script>
       var remark_config = {
         site_id: 'remark',
-        host: '<%= REMARK_URL %>',
+        host: '<%= htmlWebpackPlugin.options.remarkUrl %>',
         components: ['last-comments']
       };
 
--- a/frontend/last-comments.ejs
+++ b/frontend/last-comments.ejs
@@ -25,7 +25,7 @@
     <script>
       var remark_config = {
         site_id: 'remark',
-        host: window.location.origin,
+        host: '<%= REMARK_URL %>' || window.location.origin,
         components: ['last-comments']
       };
 
--- a/frontend/last-comments.ejs
+++ b/frontend/last-comments.ejs
@@ -25,7 +25,7 @@
     <script>
       var remark_config = {
         site_id: 'remark',
-        host: '<%= REMARK_URL %>' || window.location.origin,
+        host: '<%= REMARK_URL %>',
         components: ['last-comments']
       };
 
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -180,21 +180,33 @@ module.exports = () => ({
     new Html({
       template: path.resolve(__dirname, 'index.ejs'),
       inject: false,
+      templateParameters: {
+        REMARK_URL: remarkUrl,
+      },
     }),
     new Html({
       template: path.resolve(__dirname, 'counter.ejs'),
       filename: 'counter.html',
       inject: false,
+      templateParameters: {
+        REMARK_URL: remarkUrl,
+      },
     }),
     new Html({
       template: path.resolve(__dirname, 'last-comments.ejs'),
       filename: 'last-comments.html',
       inject: false,
+      templateParameters: {
+        REMARK_URL: remarkUrl,
+      },
     }),
     new Html({
       template: path.resolve(__dirname, 'comments.ejs'),
       filename: 'comments.html',
       inject: false,
+      templateParameters: {
+        REMARK_URL: remarkUrl,
+      },
     }),
     new MiniCssExtractPlugin({
       filename: '[name].css',
--- a/frontend/webpack.config.js
+++ b/frontend/webpack.config.js
@@ -180,33 +180,25 @@ module.exports = () => ({
     new Html({
       template: path.resolve(__dirname, 'index.ejs'),
       inject: false,
-      templateParameters: {
-        REMARK_URL: remarkUrl,
-      },
+      remarkUrl,
     }),
     new Html({
       template: path.resolve(__dirname, 'counter.ejs'),
       filename: 'counter.html',
       inject: false,
-      templateParameters: {
-        REMARK_URL: remarkUrl,
-      },
+      remarkUrl,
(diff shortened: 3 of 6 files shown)


docker-init.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/sh
 echo "prepare environment"
 # replace BASE_URL constant by REMARK_URL
-sed -i "s|https://demo.remark42.com|${REMARK_URL}|g" /srv/web/*.js
+sed -i "s|https://demo.remark42.com|${REMARK_URL}|g" /srv/web/*.{js,html}
```
akellbl4, line 4: We don't need to `sed` in HTML files because we already have the right URL throw env variable.
Correct me if I'm wrong.

docker-init.sh
```diff
@@ -1,7 +1,12 @@
...
+
+if [ -n "${SITE_ID}" ]; then
+  #replace "site_id: 'remark'" by SITE_ID
+  se -i "s|'remark'|'${SITE_ID}'|g" /srv/web/*.html
```
akellbl4, line 8: Also, I think `SITE_ID` could be passed via env vars on build time without replacing it after.

docker-init.sh
sattellite: After my changes final static html files in docker image contain constant value for `host` in `<script>` tag. And constant value `remarkUrl` in `webpack.config.js` is `https://demo.remark42.com`.

docker-init.sh
sattellite: Yes, if I will build my own docker image this pull request not needed. But I use public docker image`umputun/remark42` and this changes will fix it for usage without  subdomain.

docker-init.sh
sattellite: I changed templates and replaced `host: window.location.origin` to `host: <%= REMARK_URL %> || window.location.origin`.
`REMARK_URL` in templates wll be replaced by webpack. In webpack config this variable filled by `remarkUrl`.

umputun:
confused. How this even related to subdomain configuration vs without subdomain configuration? Either way  REMARK_URL has to be passed in or UI it won't be able to communicate with the server.

> Widgets use it own config with host as window.location.origin which returns the fqdn only.

I don't get it.  @akellbl4  correct if I wrong, but I think `window.location.origin` is not used to discover the server. I don't see how this can be the case and how `window.location.origin` can be used as a substitution for REMARK_URL

akellbl4:
@umputun, I forgot to say about I made review only from code point of view. And I want to reproduce this bug and check how this PR fix it.
`window.location.origin` takes current host name where document is served. It means this variable should represent current remark host.

sattellite:
This fix affects the pages of widgets, when directly accessing them.

`/srv/web` contains static html files for widgets: `comments.html`, `last-comments.html`, `counter.html` and demo page `index.html`.
This pages contain mostly identical inline js-config for remark42:
```js
var remark_config = {
    site_id: 'remark',
    host: window.location.origin,
    components: ['last-comments'] // or other 
};
```

And this config is used to connect widget scripts:

```js
(function(c) {
        for (var i = 0; i < c.length; i++) {
          var d = document,
            s = d.createElement('script');
          s.src = remark_config.host + '/web/' + c[i] + '.js';
          (d.head || d.body).appendChild(s);
        }
      })(remark_config.components || ['embed']);
```

`remark_config.host` contains origin (fqdn) instead of correct `REMARK_URL`. Generates incorrect url for component => script not loaded => widget don't working.

Current behavior for REMARK_URL=https://remark.example.com:
  - remark_config.host will be equal `https://remark.example.com`
  - page https://remark.example.com/web/last-comments.html will load https://remark.example.com/web/last-comments.js
  - all work as expected

Current behavior for REMARK_URL=https://example.com/remark:
  - remark_config.host will be equal `https://example.com`
  - page https://example.com/**remark/**web/last-comments.html will try load https://remark.example.com/web/last-comments.js
  - empty page

My patches force to set correct REMARK_URL and SITE_ID to relevant variables.

umputun:
thx for the detailed explanation.

sed we currently doing is not something I proud of. If possible it will be nice to resolve the issue without adding more seds.

sattellite:
`sed` is simple solution in my opinion. Other solutions make it more difficult:
- on backend side serve as templates, not static files. Fully generate all html and js with webpack to correct files for `html/template`. All `sed` will not be needed.
- on backend side add api route to get current host and bucket (no new `sed`, old `sed` will remain)
- on web server(nginx, caddy, etc.) side replace strings. Ugly and difficult, but `sed` will be not needed.

sattellite:
And if create serving templates instead static files it will be step to solve style customization issue #5

akellbl4:
@sattellite as I said in comments all of this things could be done only with passing env vars via webpack to templates and js.

docker-init.sh
akellbl4: yep, you can change this behavior and replace `remarkUrl` value to process.env.REMARK_URL

akellbl4:
[This bug](https://github.com/umputun/remark42/issues/567) relative to the PR.

sattellite:
> UPD: I've checked our building process again. And if we want to handle it via webpack we need to do it on the build stage and variables pass to container on run time. It means we could change these vars only by sed or be serving them and replace it by go.

Agree with you. Serving with go web-server also can help solve other problems, as I said upper.

akellbl4:
@umputun wdyt? Using `sed` means we have some bad practice but an easier way to serve HTML. Serve templates with transformation in `go` means cleaner code but more complexity in this place.

I prefer to stay with `sed` for now and think about how we can handle it in some other way because I don't want to bring more complexity.

umputun:
yeah, agree. will keep sed for now

frontend/webpack.config.js
```diff
@@ -180,21 +180,33 @@ module.exports = () => ({
     new Html({
       template: path.resolve(__dirname, 'index.ejs'),
       inject: false,
+      templateParameters: {
```
akellbl4, line 183: Sorry for mixing you up. This way is absolutely useless for us and we can remove these params passing to the template.

frontend/last-comments.ejs
```diff
@@ -25,7 +25,7 @@
     <script>
       var remark_config = {
         site_id: 'remark',
-        host: window.location.origin,
+        host: '<%= REMARK_URL %>' || window.location.origin,
```
akellbl4, line 28: `REMARK_URL` is required param. It means this expression will always return the first value and we don't need to keep `window.location.origin`

docker-init.sh
akellbl4: Replacing `remark` could cause side-effects like replacing this word in unexpected palaces.
I could suggest putting unique value or template like expression to places that we want to replace.

Something like `{% SITE_ID %}` and comment about how it works.
The same we should do for replacing `REMARK_URL`

frontend/webpack.config.js
sattellite: This option allows use `REMARK_URL` instead of `htmlWebpackPlugin.options.REMARK_URL`

https://github.com/jantimon/html-webpack-plugin#:~:text=templateParameters

akellbl4: Yep, I understand. But it's useless because we should to replace strings only on running time with passed vars to env and this template compilation will be done on building stage.

docker-init.sh
sattellite: Yes, current solution is dirty hack. And `site_id` is hard coded. But this script called only inside docker container. Would not this solution break the usage without docker on bare metal?

If I will add expressions like `{% SITE_ID %}` and `{% REMARK_URL %}` it will keep the same after `npm run build`. And replace expressions to correct values will require `postbuild` script with `sed` commands.

akellbl4: Yep, you are right. And now we have the issue in bin build. With this PR we will fix the issue only for docker.
Let's use your fix and I will think about how to fix it in all ways for all builds.

akellbl4 (approved):

umputun (approved):
thx

PR #808, opened 2020-11-04
Fix negative comments count

### LINKED ISSUE #776: -1 comment
Screenshot_20200914_074141_com yandex browser

mullakhmetov:
Comments count may be decreased twice (and even become negative) if you delete user with deleted comments.

closes #776

### DIFF
--- a/backend/app/store/engine/bolt.go
+++ b/backend/app/store/engine/bolt.go
@@ -797,6 +797,14 @@ func (b *BoltDB) deleteComment(bdb *bolt.DB, locator store.Locator, commentID st
 		if e = b.load(postBkt, commentID, &comment); e != nil {
 			return errors.Wrapf(e, "can't load key %s from bucket %s", commentID, locator.URL)
 		}
+
+		if !comment.Deleted {
+			// decrement comments count for post url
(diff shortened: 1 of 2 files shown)


paskal:
LGTM, one question: does the test checks count before and after deletion? By the changelog, it seems like the new behaviour is not tested, e.g. I see no test which would reproduce the error described in #776.

mullakhmetov:
>does the test checks count before and after deletion? 

Tests don't check count twice, but check count after repeated deletion.

>it seems like the new behaviour is not tested, e.g. I see no test which would reproduce the error described in #776

All three changed tests reproduce the bug without my changes in `deleteComment`. In `TestBolt_DeleteComment` after repeated deletion comments count become zero, [while 1 is expected.](https://github.com/mullakhmetov/remark/blob/ce7f727100540032154862fbb0a0034b3c205d6d/backend/app/store/engine/bolt_test.go#L710) In `TestBoltAdmin_DeleteUserHard` and `TestBoltAdmin_DeleteUserSoft` it becomes negative.

paskal (approved):

umputun (approved):
Thx for the PR, LGTM

Issue #819, opened 2020-11-22
Allow people to comment even if third party cookies not allowed


paul-nameless:
When Firefox with third party cookies enabled, here is what I get:
`Disable third-party cookies blocking to login or open comments in new page`

And it is expected. But other comment engine continue to work and allow to login though it is indeed will be logged out after page refresh.
I would expect to work this way and would like to know what you think about it?
Also, anonymous comments could work in this case.

umputun:
I think the backend is mostly ready to support headers-only auth. As long as [SendJWTHeader](https://github.com/go-pkgz/auth/blob/master/auth.go#L56) set, auth lib won't try to set cookies and will send auth header (with JWT) back.

In order to integrate it into remark42 we need 2 changes:

1. add a parameter to "auth" group setting  `SendJWTHeader`
1. change the frontend in 2 places:
   1. the login should store the auth header (default `X-JWT`) locally
   1. each request or at least each non-public request should send this header

The first part is trivial, but I'm not sure how hard to do UI side of this

umputun:
and one more thing - currently auth handled completely automatically for the UI, i.e. it doesn't have to worry about any aspect of it after the initial login. If we work without cookies and the browser doesn't help us with "cookies as a transport for JWT" the UI will have to handle token refreshes as well. Hopefully, this won't be too hard as all it has to do is an update for the stored JWT in case of `X-JWT` sent by the server.

paskal:
Backend part is done, frontend support need to be written.

PR #824, opened 2020-11-29
#819 add SendJWTHeader auth option, expose it on /config

paskal:
Backend code for #819 resolution, based on current auth library master (yet unpublished).


umputun:
you can switch it to v0.13.0 of auth

umputun (approved):
LGTM. Probably this part (after we have UI support) should be documented, but this will be a different issue and a different PR

PR #834, opened 2020-12-20
#819 Added jwt token to requests

madmages:
1. If X-JWT header received by any response - it will be stored in localstorage
2. If jwt is in localstorage - it will be send by each request
3. If 403 received and jwt in localstorage - it will be removed.

### DIFF
--- a/frontend/app/common/constants.ts
+++ b/frontend/app/common/constants.ts
@@ -33,9 +33,6 @@ export const LS_SORT_KEY = '__remarkSort';
 /** localstorage key for email of logged in user */
 export const LS_EMAIL_KEY = '__remarkEmail';
 
-/** localstorage key for JWT token */
-export const LS_JWT = '__JWT';
-
 /** Header name for jwt token */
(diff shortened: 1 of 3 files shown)


umputun:
thx for the PR. However, I'm not sure about storing JWT in the local storage. Isn't this open the system to a whole new class of XSS attacks? And, as far as I understand, #819 didn't propose this, and all it supposed to do is "allow to login though it is indeed will be logged out after page refresh.", i.e. keep the token in the memory only.

madmages:
Ok.
In #819 was not present restrictions for realization, so i made as i wish.
Usage of localstorage is present in couple places, i thought it`s not a problem.

Changed storage to local variable.

akellbl4 (dismissed):
LGTM

paskal:
If JWT token is stored in a variable which is not accessible outside of the script in question, I am OK with these changes. I am not good enough in JS to know for sure, I only know that local storage is not good for that for [multiple reasons](https://stackoverflow.com/a/44209185/961092).

Issue #836, opened 2020-12-30
Bug: Security 


uPagge:
Hello

I am facing the following problem. Completely outsider deployed my site (site was created as documentation), remark42 was configured on this site. And he was able to write quietly comments on his new site, which is a fork of mine.

This way anyone can use my instance. Is it possible to stop it somehow?

umputun:
So, someone copied your side, run it under a different url/domain and can post comments to his fork. If this is the case, I don't see it as a security bug. You have deployed an open system with virtually unlimited access in order to allow outsiders to post comments. To me it seems to be the same "security bug" as ability of someone on the internet to run your side in his iframe and he will "use your instance" in the similar way. And one more thing – outsiders (your normal users) already use your instance in the very same way, i.e. sending API requests to your instance. Btw, I think, I have seen such behavior in Disqus back in days, i.e. web forks post comments under the "original account". This is the side effect of including comments in an iframe.

The practical question is how to limit such unexpected and unwanted forks. @paskal, @akellbl4 (anybody else are welcome to suggest as well) feel free to suggest something clever. I can see several semi-working way to achieve it:

1. Setup CORS on your proxy to allow your domains only (AllowedOrigins). I think, nothing prevents you doing it in the current version. This will limit the ability of someone to clone the site, however nothing prevents some smarter "bad guy" from making such a calls from non-browser platform (mobile app for example).
2. You and only you have the admin control (admin user in your config only), unless your instance compromised. You may notify this outsider and ask him to stop it, and if he refuses - you may block his users and permanently delete all the comments they made. This is not an ideal solution in case if you share the same users, but anyway - you in control.

uPagge:
Thanks for the answer, I'll try to do CORS

umputun:
We just played with this a little bit. In case if you are running remark42 with nginx proxy, this is what you need on a server section hosting comments: `add_header Content-Security-Policy "frame-ancestors domain1 domain2;";`

i.e. `add_header Content-Security-Policy "frame-ancestors https://radio-t.com https://pirates.radio-t.com http://pirates.radio-t.com;";`

see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

PR #837, opened 2020-12-30
add optional frame-ancestors (Content-Security-Policy) for the list of hosts passed by user

umputun:
This PR adds `--allowed-hosts` (env:"ALLOWED_HOSTS") enforcing  frame-ancestors Content-Security-Policy to limit which sites can include remark42 in any kind of frame. Related to https://github.com/umputun/remark42/issues/836

For more details about frame-ancestors see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors


akellbl4 (approved):
LGTM

Issue #784, opened 2020-10-04
Not working on chrome due to lack of SameSite=None attribute


aca:
I think I have issue with all chromium based browser. (Tried brave, and mobile chrome)
With firefox everything is working just fine.

issue page: https://aca.github.io/2020_10_hello_world.html

If I try to login anonymously It prints
![image](https://user-images.githubusercontent.com/50316549/95025464-e200ec00-06c4-11eb-81cd-ed34cf67fd25.png)

If I try to log out, It prints some cookie error.
```
2020/10/04 19:32:07.882 [INFO]  logout not allowed - token cookie was not presented: http: named cookie not present - 403 - /auth/logout?site=aca.github.io [caused by auth/provider/oauth2.go:220 provider.Oauth2Handler.LogoutHandler]
```

GIthub login also doesn't work.

```
$ ./remark42 version
remark42 master-71dcb13-20200426T12:41:26
```

umputun:
it doesn't work in any browser for me, not just chromium-based. Probably due to mismatch in domains? I see your remark42 lives on https://acadx0.mooo.com but the site itself on https://aca.github.io

If you do https://acadx0.mooo.com/web/ you will see everything works on all browsers, including chromium-based.

aca:
@umputun 
Yes I hosted it on other domain as I cannot serve remark on github.io.
If I'm wrong or, if there is any workaround to integrate remark with github.io, please let me know.

aca: https://aca.github.io
remark: https://acadx0.mooo.com (with caddy reverse-proxy for https)

It seems it's working fine with ms edge(which is chromium as far as I know).
I've rechecked with firefox. It's working fine. What is going on?

umputun:
> What is going on?

Not sure. On FF I see cookies set as expected, however not on safari or chrome. 

@Mavrin something prevents browser to set cookies and I'm not sure what exactly. Can you take a look? 
@paskal in case you have some bright ideas pls share.

aca:
@umputun I found out what's going on.


Chrome set samesite default to Lax, and Firefox set it to None (which will change in the future).
I patched this line 
https://github.com/umputun/remark42/blob/ecb146bf20ba7f6f6ab2ec82e22e1b3de8dc9823/backend/vendor/github.com/go-pkgz/auth/token/jwt.go#L234

with
```
	jwtCookie := http.Cookie{
		Name: j.JWTCookieName, Value: tokenString, HttpOnly: true, Path: "/",
[...]

PR #841, opened 2021-01-05
Allow cross site requests

janikvonrotz:
We want to allow cross site requests if comments are hosted on different domain/subdomain.

Details on the setting: https://golang.org/pkg/net/http/#SameSite

Change as proposed in #784


umputun (changes requested):
This is vednored directory. We don't want to change it because the next update of go-pkgz/auth will wipe it out. The change should be submitted to https://github.com/go-pkgz/auth

janikvonrotz:
I opened an issue https://github.com/go-pkgz/auth/issues/73

I might give it try, so it does not end up in the backlog.

umputun:
this was done by #850

Issue #677, opened 2020-04-13
New look for auth form


akellbl4:
I made simple sketches (I am not a designer and I did it just on iPad with Adobe Comp), It shows the way to combine all auth providers in a compact and convenient form.
For social auth providers may use logos.

![telegram-cloud-photo-size-4-5996946266153660949-y](https://user-images.githubusercontent.com/2330682/79113980-c3134400-7d8a-11ea-881f-2ca657ca0d6f.jpg)

umputun:
looks good. And logos for social auth should make it even better

Issue #563, opened 2020-01-18
Modal dialogs for email subscription and notifications 


umputun:
Currently, both windows can be closed by accident. Practically this is not that bad, but annoying.

It would be nice to figure a way to make those dialog windows modal or immune to an accidental close in some other way.

umputun:
@Mavrin & @akellbl4  - I'm not sure what the best way to do it. If we make those window modal we will also need a way to close them somehow. Pls, share your brilliant ideas about how to achieve it without much work and redesign.

akellbl4:
It can be "permanent" dropdowns which can be closed only by clicking on cross icon.

PR #844, opened 2021-01-08
Update auth block

akellbl4:
- API placed inside auth component
- Convert dropdown to hook (`setInterval` changed to `MutationObserver` for changing the height of the window)
- Removed `sleep` from auth block (other dropdowns should be changed in the same way)
- Auth has its own button (the current button should be replaced by it because the new one uses css-modules and rely on the global theme defined by global class name)

Tasks:
- [x]  New design looks like prototype #677 
- [x] Permanent token step (in case of miss click out of dropdown) #563
- [x] Merge separated code
- [x] Write tests
- [x] Fix translations
- [x] Finalize changes


paskal:
I would propose doing "Magic login links" in a separate PR, it doesn't have to be tied to any other changes you want to introduce to the auth module. Blending it together only will make the review harder.

frontend/package.json
```diff
@@ -26,91 +26,93 @@
...
+    "react-intl": "^5.12.5",
     "react-redux": "^7.2.2",
     "redux": "^4.0.5",
     "redux-thunk": "^2.3.0"
   },
   "devDependencies": {
-    "@babel/preset-env": "^7.12.11",
-    "@babel/preset-react": "^7.12.10",
-    "@formatjs/cli": "^3.0.1",
+    "@babel/preset-env": "^7.12.17",
```
paskal, line 45: Let's bump dependencies in a separate PR? It doesn't seem related to auth changes.

akellbl4: Sure, I've already made a separate brand. PR will be in a min.

akellbl4:
@Mavrin hi there, I've finished writing tests. Could you please take a shallow look at it? Maybe you have some suggestions because I've done it the first time with `testing library`

frontend/app/custom-properties.css
```diff
@@ -1,52 +1,50 @@
 :root {
-  --color0: #000;
-  --color1: #aaa;
-  --color3: #e2efef;
-  --color4: #edf6f7;
-  --color5: #eee;
-  --color6: #fff;
+  --color0: #0f172a;
```
akellbl4, line 2: Variables were just sorted by color.
CleanShot 2021-02-25 at 21 41 40@2x

frontend/app/components/auth/components/oauth-button.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import type { OAuthProvider } from 'common/types';
+import { siteId } from 'common/settings';
+import capitalizeFirstLetter from 'utils/capitalize-first-letter';
+import messages from 'components/auth/auth.messsages';
+
+import styles from './oauth-button.module.css';
+
+const OAUTH_ICONS = {
+  facebook: require('./assets/facebook.svg').default as string,
```
Mavrin, line 14: Why do you use require instead of import?

frontend/app/components/auth/components/assets/microsoft.svg
```diff
@@ -0,0 +1 @@
+<svg width="16" height="16" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill="#EC4C23" d="M0 0h7.467v7.467H0z"/><path fill="#7DB300" d="M8.533 0H16v7.467H8.533z"/><path fill="#F7B302" d="M8.533 8.533H16V16H8.533z"/><path fill="#019FE8" d="M0 8.533h7.467V16H0z"/></svg>
```
Mavrin, line 1: end of line

Mavrin (changes requested):

frontend/app/components/auth/components/oauth-button.tsx
akellbl4: In my opinion, it looks better in comparison with imports. I don't like the fence of the imports.
The problem with it is we should get `.default`. It could be solved with disabling esmodule for `.svg`. I haven't done it from the first attempt and have left as it is for a better time.

Both approaches are acceptable. It's all only about style in this case. For JS chunks it works in a slightly different way.

```js
const OAUTH_ICONS = {
  facebook: require('./assets/facebook.svg').default as string,
}
```

```js
import facebookIcon from './assets/facebook.svg'

const OAUTH_ICONS = {
  facebook: facebookIcon
}
```

frontend/app/components/auth/components/oauth-button.tsx
```diff
@@ -0,0 +1,79 @@
...
+  variant: 'full' | 'name' | 'icon';
+  onClick: JSX.GenericEventHandler<HTMLAnchorElement>;
+  provider: OAuthProvider;
+};
+
+const OAuthButton: FunctionComponent<OAuthIconProps> = ({ provider, onClick, variant }) => {
+  const intl = useIntl();
+
+  const [icon, setIcon] = useState('');
+  const providerName = capitalizeFirstLetter(provider);
```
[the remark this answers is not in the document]
akellbl4: Yep, it looks reasonable. I tried to reduce hardcode but it doesn't work by capitalizing for this case.

Mavrin (approved):

Issue #400, opened 2019-07-31
Reject multiple votes from the same IP


umputun:
The idea is to add an optional restriction treating vote request from the same IP as a vote from the same user. Technically,  this is not per IP as we don't store it, but the hash of IP. Either way, this info can be extracted, saved, and analyzed. 

_This change may not be friendly for users behind some proxy (sharing ip) and will be off by default_

umputun:
One more thing - if we extend voting info anyway, we can also include timestamp which may help to fine-tune same-ip rejection by limiting this check to some predefined time window.

The biggest problem with this task is how to keep it back-compatible with the current `Votes       map[string]bool` stored in bolt for existing comments.

PR #857, opened 2021-01-18
strip voted_ip from response

umputun:
`voted_ip` map was added to comments as a part of #400, however I don't think there is a reason to send it back to UI/API clients. We already strip `votes` map of user on both send and input validation levels and I believe the same should be done for `voted_ip` as well.

Looks like this data not in any use by the official clients, so this PR strips it completely. Pls note - presence of this map in the response doesn't expose anything bad by itself (just a bunch of hashes and time stamps) but may allow some cross-post analysis of votes to figure how the same user voted without exposing the user ID. So, not terrible but still something we may want to prevent.

### DIFF
--- a/backend/app/rest/api/rest_private_test.go
+++ b/backend/app/rest/api/rest_private_test.go
@@ -409,7 +409,8 @@ func TestRest_Vote(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, 1, cr.Score)
 	assert.Equal(t, 1, cr.Vote)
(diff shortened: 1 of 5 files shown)


paskal (approved):
From diff I am not 100% sure tests are correct, LGTM otherwise.

umputun:
for tests the only thing I had was a check of the rest endpoint making sure `voted_ips` cleaned. Just added more checks in all places we check `votes`

umputun:
@akellbl4 pls confirm we good from UI side and the lack of `voted_ips` in responses won't break the world

paskal (approved):

PR #858, opened 2021-01-19
Fix admin names

umputun:
There was a bug checking admins match for anonymous names based on IDs. Obviously, this match never worked right because used id (i.e. github_123465abcdef) is not the name.

Added the new (optional) list of admin names and changed match logic.

see #605 and #612


umputun:
@paskal pls take a quick look as you have a chance, because I have updated your exsmple/mem implementation as well

README.md
```diff
@@ -110,7 +110,8 @@ _this is the recommended way to run remark42_
 | store.type              | STORE_TYPE              | `bolt`                   | type of storage, `bolt` or `rpc`                |
 | store.bolt.path         | STORE_BOLT_PATH         | `./var`                  | path to data directory                          |
 | store.bolt.timeout      | STORE_BOLT_TIMEOUT      | `30s`                    | boltdb access timeout                           |
-| admin.shared.id         | ADMIN_SHARED_ID         |                          | admin names (list of user ids), _multi_         |
+| admin.shared.id         | ADMIN_SHARED_ID         |                          | admin ids (list of user ids), _multi_           |
+| admin.shared.name       | ADMIN_SHARED_NAME       |                          | admin names, _multi_                            |
```
paskal, line 114: LGTM except for this line.

It's even `admin.shared.name`, it's "list of names restricted to use by anonymous users and maybe by email users in the future if we'll change our mind". We should add a proper description to that new option and (maybe) move it somewhere else, as it has nothing to do with admins other than it's recommended use for that functionality.

paskal (changes requested):

README.md
umputun: yeah, bothered me as well. The thoughts behind that choice were - such names may be used not just to limit auth but for something else, like get the list of admin names. However I tend to agree with you because there is no any relation between names and ids, so if we even add the list of names it won't be associated with ids.

Reverting this part and moving it to restricted-names param

backend/app/cmd/server.go
```diff
@@ -72,6 +72,7 @@ type ServerCommand struct {
 	WebRoot          string        `long:"web-root" env:"REMARK_WEB_ROOT" default:"./web" description:"web root directory"`
 	UpdateLimit      float64       `long:"update-limit" env:"UPDATE_LIMIT" default:"0.5" description:"updates/sec limit"`
 	RestrictedWords  []string      `long:"restricted-words" env:"RESTRICTED_WORDS" description:"words prohibited to use in comments" env-delim:","`
+	RestrictedNames  []string      `long:"restricted-names" env:"RESTRICTED_NAMES" description:"names prohibited to use by user" env-delim:","`
```
paskal, line 75: Should we mention "prohibited to use by the anonymous users"? AFAIK we fetch whatever user has in GitHub and in other places, and it's relevant only for anonymous and email users but latter are out of the scope. Otherwise, looks good (aside from many unrelated changes like go.sum).

umputun: i think we better keep it more universal. Currently we restrict anon and email only, however it may be a good idea to restrict any auth method with the similar logic. I'm not brave enough to do it right away but this is smth to consider

paskal (approved):

PR #866, opened 2021-01-28
Use parens for arrow funcs always

akellbl4:
I'm working on #844 and already changed this rule there. I want to separate this change to its own PR and make that PR smaller.
This PR contains only change of `.prettierrc.js` and changes of `nom run prettier` command.
Also, I would like to explain my decision but docs of [Prettier](https://prettier.io/docs/en/options.html#arrow-function-parentheses) already explained it.

> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option

### DIFF
--- a/frontend/.babelrc.js
+++ b/frontend/.babelrc.js
@@ -1,4 +1,4 @@
-const getPresetEnv = options => ['@babel/preset-env', options];
+const getPresetEnv = (options) => ['@babel/preset-env', options];
 const preactPreset = [
   '@babel/preset-react',
   {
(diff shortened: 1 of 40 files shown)


akellbl4:
Looks like code coverage report are lying because I've made only formatting but by its opinion coverage has changed :(

akellbl4:
I've found out how it works.
if part of a line is covered it will be count as fully cover line.
Then part of the line could be moved to the next line and by this change we also get coverage change.

Mavrin (approved):
👍

PR #868, opened 2021-01-31
Update fetcher and add good coverage to it

akellbl4:
I like this version because `stringifyUrl` is used inside of fetcher, but when we don't have query params we should pass it.


frontend/app/common/fetcher.ts
```diff
@@ -13,38 +14,60 @@ export const XSRF_HEADER = 'X-XSRF-TOKEN';
...
-const fetcher = METHODS.reduce((acc, method) => {
-  acc[method] = async (uri: string, params: FetchInit = {}) => {
-    const { headers = {}, json, ...fetchParams } = params;
+const fetcher = METHODS.reduce<FetcherMethods>((acc, method) => {
+  acc[method] = async (uri: string, query: QueryParams = {}, body?: Payload) => {
     // add api base if it's not auth request
     // we use `indexOf` instead of `startsWidth` because we don't want to have another one polyfill for no reason
-    const baseUrl = uri.indexOf('/auth') === 0 ? BASE_URL : `${BASE_URL}${API_BASE}`;
-    const url = `${baseUrl}${uri}`;
+    const url = stringifyUrl(uri, query);
```
paskal, line 49: Inconsistent use of BASE_URL across the changed code looks weird, as well as appending it here.

akellbl4: It was the same but determined by passed param.
We know we don't need to add API_BASE for auth requests and determine it by start of URI

akellbl4:
@paskal if you have any idea hot to make this abstraction better I appreciate any suggestions

frontend/app/common/api.ts
```diff
@@ -44,14 +20,14 @@ const __loginViaEmail = (token: string): Promise<User | null> =>
...
-  fetcher.get(stringifyUrl('/auth/email/login', { id: siteId, user: username, address }));
+  fetcher.get('/auth/email/login', { id: siteId, user: username, address });
 
 export const logIn = (provider: AuthProvider): Promise<User | null> => {
   if (provider.name === 'anonymous') return __loginAnonymously(provider.username);
   if (provider.name === 'email') return __loginViaEmail(provider.token);
 
   return new Promise<User | null>((resolve, reject) => {
-    const url = stringifyUrl(`${BASE_URL}/auth/${provider.name}/login`, {
+    const url = stringifyUrl(`/auth/${provider.name}/login`, {
```
paskal, line 30: That's the only use of exported `stringifyUrl` I could find, doesn't feel right to expose this internally used function for that single use case. Maybe would be better to create a wrapper `constructLoginURL` and export it? Then it will stay weird, but it will be explicitly weird instead of the current implicit way.

frontend/app/common/fetcher.ts
```diff
@@ -13,38 +14,60 @@ export const XSRF_HEADER = 'X-XSRF-TOKEN';
...
-type FetcherObject = Record<Methods[number], <T>(url: string, params?: FetchInit) => Promise<T>>;
+type QueryParams = Record<string, string | number | undefined>;
+type Payload = BodyInit | Record<string, unknown> | null;
+type FetcherMethods = Record<'get' | 'delete', <T>(url: string, query?: QueryParams) => Promise<T>> &
+  Record<'put' | 'post', <T>(url: string, query?: QueryParams, body?: Payload) => Promise<T>>;
+
+export function stringifyUrl(uri: string, query: QueryParams) {
+  const queryEntries = Object.entries(query);
+  const siteIdParam = `site=${encodeURIComponent(siteId)}`;
+  const baseUrl = uri.indexOf('/auth') === 0 ? BASE_URL : `${BASE_URL}${API_BASE}`;
```
paskal, line 25: I propose to extract that "either BASE_URL or API_BASE as well" to some higher level, so the base URL would be already clear by the time you call that function.

`stringifyUrl` currently does more than the name implies and also it should know some internals about the URL which is passed to it, that logic should be moved up the stack by wrapping this functionality into separate function or object, like `fetcher` vs `authFetcher`, or `fetcher` vs `apiFetcher` which would be distinct only by different baseURL, which would be determined by the time of object creation and not per URL passed to it.

frontend/app/common/fetcher.ts
paskal: Added a comment in the more relevant place of the code.

paskal (changes requested):

frontend/app/common/api.ts
```diff
@@ -157,54 +129,46 @@ export const blockUser = (
...
-      name: image.name,
-      size: image.size,
-      type: image.type,
-      url: `${BASE_URL + API_BASE}/picture/${resp.id}`,
-    }));
+  return fetcher.post<{ id: string }>('/picture', {}, data).then((resp) => ({
+    name: image.name,
+    size: image.size,
+    type: image.type,
+    url: `${BASE_URL + API_BASE}/picture/${resp.id}`,
```
paskal, line 156: Seems to be the only use of BASE_URL and API_BASE here, doesn't seem right as well. Should we make a helper function `constructPictureURL` in fetcher instead maybe?

frontend/app/common/fetcher.ts
```diff
@@ -13,38 +14,67 @@ export const XSRF_HEADER = 'X-XSRF-TOKEN';
...
-  json?: unknown;
-  query?: Record<string, string | number | undefined>;
-};
-type FetcherObject = Record<Methods[number], <T>(url: string, params?: FetchInit) => Promise<T>>;
+type QueryParams = Record<string, string | number | undefined>;
+type Payload = BodyInit | Record<string, unknown> | null;
+type FetcherMethods = Record<'get' | 'delete', <T>(url: string, query?: QueryParams) => Promise<T>> &
+  Record<'put' | 'post', <T>(url: string, query?: QueryParams, body?: Payload) => Promise<T>>;
+
+export function stringifyUrl(uri: string, query: QueryParams) {
```
paskal, line 22: When I'm thinking about it, maybe `stringifyUrl` shouldn't be in this module even. If I wrote that code, I would try to extract it to the separate module to clearly see the dependencies and make it work with as few dependencies as possible so that it wouldn't know too much about the app. Now it's not the case, and because it's in the fetcher module, it's not as apparent as it could be in the separate one.

frontend/app/common/fetcher.ts
akellbl4: Yep. You are right. I thought about making some sort of fabric and pass different base URLs.

akellbl4:
@paskal appreciate your opinion. Have changed the code in a better way.
@Mavrin I've removed the function for stringifying query params and started using `URLSearchParams` for it.
Also removed `size-limit` from legacy chunks because they are could grow when we use modern but more reliable browser API. The percentage of people with old browsers is small and I think we should think about the larger group of users.

akellbl4:
`api.test.ts` was dropped because this test not actual anymore in terms of use `URLSearchParams `. It covers encoding of params to save string for url.

frontend/app/common/fetcher.ts
```diff
@@ -1,100 +1,73 @@
...
-      credentials: withCredentials ? 'include' : 'omit',
-    };
-
-    if (body) {
-      if (contentType === 'multipart/form-data') {
-        parameters.body = body as FormData;
-      } else if (typeof body === 'object' && !(body instanceof Blob) && !(body instanceof ArrayBuffer)) {
-        parameters.body = JSON.stringify(body);
+const createFetcher = (baseUrl: string = '') =>
+  METHODS.reduce<FetcherMethods>((acc, method) => {
```
Mavrin, line 26: It is looks little messy. I think will be better to define each method separate.

Mavrin (approved):

frontend/app/common/fetcher.ts
```diff
@@ -104,36 +77,33 @@ const fetcher = methods.reduce<Partial<FetcherObject>>((acc, method) => {
...
-  };
-  return acc;
-}, {}) as FetcherObject;
+      }
+    };
+    return acc;
+  }, {} as FetcherMethods);
 
-export default fetcher;
+export const apiFetcher = createFetcher(`${BASE_URL}${API_BASE}`);
```
paskal, line 108: For consistency, if you are creating fetcher a level higher in one place, let's create apiFetcher a level higher as well?

akellbl4: `apiFetcher` now is used in two places. I would like to create it in other place when I'll find out how to tree shake all unused functions from file if one of our bundles used only one method from it.

paskal: Let's move `authFetcher` and `adminFetcher` in and don't export `createFetcher` then? It doesn't seem right to have two approaches both in the code at the same time.

paskal (approved):

paskal (dismissed):
LGTM except for failing build. Better than the point we started at.

frontend/app/common/fetcher.ts
```diff
@@ -118,3 +118,4 @@ const createFetcher = (baseUrl: string = ''): Methods => {
 export const apiFetcher = createFetcher(`${BASE_URL}${API_BASE}`);
 export const authFetcher = createFetcher(`${BASE_URL}/auth`);
 export const adminFetcher = createFetcher(`${BASE_URL}${API_BASE}/admin`);
+export default createFetcher;
```
paskal, line 121: I thought the whole idea was to keep `createFetcher` private? I propose to stop using it in the test and start using one of the pre-created fetchers instead.

paskal (changes requested):

frontend/app/common/fetcher.ts
akellbl4: okay, got it

paskal (approved):

Mavrin (approved):

akellbl4:
🎉

PR #872, opened 2021-02-04
send cookie with request

akellbl4:


akellbl4:
The bug appears only on OAuth authorization, don't know what difference in cookie settings between email and OAuth method.

PR #874, opened 2021-02-04
Fix oauth providers

akellbl4:


paskal:
@akellbl4 would you be so kind as to provide some content around that PR?

PR #879, opened 2021-02-10
Fix empty username check for Disqus migrator

enescakir:
I found a bug while importing comments from Disqus.

It generates `comment.User.ID` from author's username. 
https://github.com/umputun/remark42/blob/3628467367277be356fd56d4590ca1740a121cd7/backend/app/migrator/disqus.go#L130-L134

If author's username is empty (`c.User.ID == "disqus_"`), it uses author's name for user ID. It assumes `store.EncodeID` returns empty string for empty username.
https://github.com/umputun/remark42/blob/3628467367277be356fd56d4590ca1740a121cd7/backend/app/migrator/disqus.go#L140-L142

But `store.EncodeID` returns `da39a3ee5e6b4b0d3255bfef95601890afd80709` for empty string.
https://github.com/umputun/remark42/blob/3628467367277be356fd56d4590ca1740a121cd7/backend/app/store/user.go#L52-L54

`reValidSha.MatchString(val)` doesn't return `true` for empty values. Example: https://play.golang.org/p/b8CZt6gR4X7

As I understand from the `// already hashed or empty` comment, it expects to return for empty values too.

As a result, all authors from Disqus -who don't have username- have user id `disqus_da39a3ee5e6b4b0d3255bfef95601890afd80709`.
So all comments link to one author.

### DIFF
--- a/backend/app/migrator/disqus.go
+++ b/backend/app/migrator/disqus.go
@@ -137,7 +137,7 @@ func (d *Disqus) convert(r io.Reader, siteID string) (ch chan store.Comment) {
 						ParentID:  comment.Pid.Val,
 						Imported:  true,
 					}
-					if c.User.ID == "disqus_" { // empty comment.AuthorUserName from disqus
+					if len(comment.AuthorUserName) == 0 { // empty comment.AuthorUserName from disqus
 						c.User.ID = "disqus_" + c.User.Name
 					}
 					if c.ID == "" { // no comment.UID
--- a/backend/app/migrator/disqus.go
+++ b/backend/app/migrator/disqus.go
@@ -137,7 +137,7 @@ func (d *Disqus) convert(r io.Reader, siteID string) (ch chan store.Comment) {
 						ParentID:  comment.Pid.Val,
 						Imported:  true,
 					}
-					if len(comment.AuthorUserName) == 0 { // empty comment.AuthorUserName from disqus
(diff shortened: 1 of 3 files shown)


enescakir:
I see empty string test case now. I am confused. Should it return `""` or `da39a3ee5e6b4b0d3255bfef95601890afd80709` for empty string? I think test case doesn't match with  disqus importer's expectation and `EncodeID`s comment.
Does disqus importer have wrong expectation or Is testcase is wrong?

https://github.com/umputun/remark42/blob/3628467367277be356fd56d4590ca1740a121cd7/backend/app/store/user_test.go#L17

umputun:
>  Should it return "" or da39a3ee5e6b4b0d3255bfef95601890afd80709 for empty string? 

Good question. The comment " // already hashed or empty" implies the behavior you expected and fixed, however, tests ensure hashed values even for an empty string. I kind of remember making hashes for an empty string for some reason, but can't recall what the reason was.

I'd suggest leaving it as is and adjust the check in disqus importer. Also will be nice to have a test for such cases.

enescakir:
I revert my first commit, and move the fix to disqus importer. Also, I updated test of disqus importer to test empty username.

umputun:
Looking good. The only thing I don't like is the "naked" user name in the id. It doesn't expose any additional info (the name is kind of public anyway) but still inconsistent. 

I'll merge your PR and will change it latter on

PR #884, opened 2021-02-13
Ignore autogenerated loadLocale.ts in coverage report

akellbl4:


paskal:
Is it possible to test them instead to verify that text in the translated language is not the same as in English for some text from the list? Would verify that language added both to code and list of languages.

akellbl4:
@paskal If we would like to do so It won't cover this file. Also, code to this file adds when we generate a new dictionary, and the only possible way not to add it just to remove code from the generated file.
By the way, the idea is good we should add some tests for checking if locales work.

paskal (approved):

PR #888, opened 2021-02-18
Encode e-mail subject to support non-ASCII chars

VeselovAlex:
Add base64-encoding to email `Subject:` header value to allow non-ASCII characters, according to RFC2047. Current implementation produces invalid messages when e.g. page title contains Cyrillic letters.

### DIFF
--- a/backend/app/notify/email_test.go
+++ b/backend/app/notify/email_test.go
@@ -287,7 +287,7 @@ Date: `)
 	// send email to both user and admin, without parent set
 	email.AdminEmails = []string{"admin@example.org"}
 	req = Request{
-		Comment: store.Comment{ID: "999", User: store.User{ID: "1", Name: "test_user"}, PostTitle: "test_title"},
+		Comment: store.Comment{ID: "999", User: store.User{ID: "1", Name: "test_user"}, PostTitle: "Привет"},
 		Emails:  []string{"test@example.org"},
 	}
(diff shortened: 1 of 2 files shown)


paskal:
Would you be so kind as to add tests for this?

backend/app/notify/email_test.go
```diff
@@ -249,6 +249,63 @@ Content-Type: text/html; charset="UTF-8"
...
+		Comment: store.Comment{ID: "999", User: store.User{ID: "1", Name: "test_user"}, ParentID: "1", PostTitle: "Привет"},
+		parent:  store.Comment{ID: "1", User: store.User{ID: "999", Name: "parent_user"}},
+		Emails:  []string{"test@example.org"},
+	}
+	assert.NoError(t, email.Send(context.TODO(), req))
+	assert.Equal(t, "from@example.org", fakeSMTP.readMail())
+	assert.Equal(t, 1, fakeSMTP.readQuitCount())
+	assert.Equal(t, "test@example.org", fakeSMTP.readRcpt())
+	// test buildMessageFromRequest separately for message text
+	res, err := email.buildMessageFromRequest(req, req.Emails[0], false)
```
paskal, line 274: Could you please remove everything from that test function other than lines 274-285? Everything else is not needed there, these would be enough.

paskal (approved):

umputun (approved):
LGTM, thx

PR #889, opened 2021-02-19
Fix NPE when sending message to detached frame

VeselovAlex:
This patch fixes NPE from #722, tested in Firefox 85 an Edge 88 (Chromium).

The idea is to manually set `titleObserver = null` in `destroy()` and allow browser's GC to drop the observer, it's callback and captured iframe. Function `postTitleToIframe` is not called after destroy call, but I've also added null checks for functions which send messages to iframe window.

### DIFF
--- a/frontend/app/embed.ts
+++ b/frontend/app/embed.ts
@@ -300,17 +300,21 @@ function createInstance(config: typeof window.remark_config) {
   }
 
   function postTitleToIframe(title: string) {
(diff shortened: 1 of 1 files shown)


akellbl4 (approved):
LGTM, @Mavrin share your opinion on this PR before the merge.

Mavrin (approved):

PR #891, opened 2021-02-21
Make textarea-autosize shared

akellbl4:
- converted to a functional component
- moved from `comment-form` to `components` because it is used in other places not only in  `comment-form`
- methods for work with selections was moved to `comment-form` because it is used only in that component
- removed compatibility with IE9 for upload files

Separated from #844


frontend/app/components/auth/__email-login-form/auth__email-login-form.tsx
```diff
@@ -97,9 +97,8 @@ export class EmailLoginForm extends Component<Props, State> {
     try {
       await this.props.sendEmailVerification(this.state.usernameValue, this.state.addressValue);
       this.setState({ verificationSent: true });
-      setTimeout(() => {
-        this.tokenRef.current && this.tokenRef.current.focus();
-      }, 100);
+      await sleep(100);
```
paskal, line 100: Hardcoded sleep instead of a timeout doesn't seem right.

akellbl4: it's the same way as was used in this file and it is `setTimeout` wrapped in func

Mavrin (approved):

PR #899, opened 2021-02-24
Replaced parseQuery with URLSearchParams.

Paras0-7:


akellbl4:
@Paras0-7 in the PR I meant we should use browser API `URLSearchParams` ([doc](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)) instead of the function that we have.

Paras0-7:
> @Paras0-7 in the PR I meant we should use browser API `URLSearchParams` ([doc](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)) instead of the function that we have.

Issue #897, opened 2021-02-22
Post data via post message without serialization


akellbl4:
Sending data via post messages [works fine](https://caniuse.com/?search=post%20message) without any processing as a simple data structure.
It's completely unnecessary to stringify and parse JSON.

Now it's making logic more complex and unpredictable.

paskal:
Would you be so kind as to post links to the code you are referring to?

akellbl4:
https://github.com/umputun/remark42/blob/df92500bb41742dffb36f9d06fa26b506c72af1c/frontend/app/utils/postMessage.ts#L20
https://github.com/umputun/remark42/blob/5ec871fd372494318151d02ca45f3d677eabb7bb/frontend/app/embed.ts#L298 and below in this file
https://github.com/umputun/remark42/blob/df92500bb41742dffb36f9d06fa26b506c72af1c/frontend/templates/iframe.ejs#L165

I'm not sure If I found all of the places.

Elephmoon:
Hello. I belive I can help with it.

PR #903, opened 2021-03-04
PostMessage without serialization

Elephmoon:
Resolve https://github.com/umputun/remark42/issues/897
I searched through the files in the frontend, in all places where I found postMessage removed unnecessary serialization.
Also, I think if we haven't stringify, then we don't need to check for string type here.
https://github.com/umputun/remark42/blob/553bcf2fd59e400fa2f0ee4ddfa70eb9044f9f2f/frontend/app/components/root/root.tsx#L187

### DIFF
--- a/frontend/app/components/root/root.tsx
+++ b/frontend/app/components/root/root.tsx
@@ -4,7 +4,7 @@ import b from 'bem-react-helper';
 import { IntlShape, useIntl, FormattedMessage, defineMessages } from 'react-intl';
 import clsx from 'clsx';
 
-import type { Sorting } from 'common/types';
+import type { Sorting, Theme } from 'common/types';
 import type { StoreState } from 'store';
 import {
   COMMENT_NODE_CLASSNAME_PREFIX,
@@ -39,7 +39,7 @@ import { ConnectedComment as Comment } from 'components/comment/connected-commen
 import { uploadImage, getPreview } from 'common/api';
 import { isUserAnonymous } from 'utils/isUserAnonymous';
 import { bindActions } from 'utils/actionBinder';
-import { postMessage } from 'utils/postMessage';
+import { postMessage, parseMessage } from 'utils/postMessage';
 import { useActions } from 'hooks/useAction';
 import { setCollapse } from 'store/thread/actions';
 import { logout } from 'components/auth/auth.api';
@@ -177,13 +177,10 @@ export class Root extends Component<Props, State> {
     }
   };
 
-  onMessage(event: { data: string | object }) {
-    if (!event.data) {
-      return;
-    }
-
+  onMessage(event: MessageEvent<{ theme?: Theme }>) {
     try {
-      const data = typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
+      const data = parseMessage(event);
+
       if (data.theme && THEMES.includes(data.theme)) {
         this.props.setTheme(data.theme);
       }
--- a/frontend/templates/iframe.ejs
+++ b/frontend/templates/iframe.ejs
@@ -134,7 +134,7 @@
       setInterval(function () {
         if (document.body.offsetHeight !== lastHeight && document.body.offsetHeight > 22) {
           lastHeight = document.body.offsetHeight;
-          window.parent.postMessage(JSON.stringify({ remarkIframeHeight: lastHeight }), '*');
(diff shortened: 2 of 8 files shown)


akellbl4:
> Also, I think if we haven't stringify, then we don't need to check for string type here.

Yes, it's a good point, but for example, in dev, webpack could send string inside post message data.
In my opinion, we should drop `JSON.parse` for data from postMessage and process message if it's an object.
Good to add these changes in current PR.

Elephmoon:
> we should drop JSON.parse for data from postMessage

I'm not very good on typescript and I can't catch how to make it.
https://github.com/umputun/remark42/blob/553bcf2fd59e400fa2f0ee4ddfa70eb9044f9f2f/frontend/app/components/root/root.tsx#L187
In this line JSON.parse implicitly changing type to any and in next line we can get data.theme.
If I remove JSON.parse data type will not change and we will not get data.theme.
I need to define type with Theme type or what is the best way in this situation?

akellbl4:
@Elephmoon we already removed `JSON.stringify` on sending. It means we will receive JS object.
All we need to check if data is an object and work with it. Just like that:
```js
onMessage(event: MessageEvent<{ theme?: string }>) {
  if () {
     const data = event.data as Record<string, unknown>
  }
```
I don't like the way how it's should be checked but it's js. @Mavrin maybe you know cool methods to do that?

frontend/app/components/root/root.tsx
```diff
@@ -177,15 +177,14 @@ export class Root extends Component<Props, State> {
     }
   };
 
-  onMessage(event: { data: string | object }) {
-    if (!event.data) {
-      return;
-    }
-
+  onMessage(event: MessageEvent<{ theme?: string }>) {
```
akellbl4, line 180:
```suggestion
  onMessage(event: MessageEvent<{ theme?: Theme }>) {
```

frontend/app/components/root/root.tsx
```diff
@@ -177,15 +177,14 @@ export class Root extends Component<Props, State> {
...
     try {
-      const data = typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
-      if (data.theme && THEMES.includes(data.theme)) {
-        this.props.setTheme(data.theme);
+      const isObj = typeof event.data === 'object' && event.data !== null && !Array.isArray(event.data);
+      if (!isObj) {
+        return;
+      }
+      if (event.data.theme && THEMES.includes(event.data.theme as Theme)) {
```
akellbl4, line 186:
```suggestion
      if (event.data.theme && THEMES.includes(event.data.theme)) {
```

akellbl4 (approved):
Sorry for the long response.
Thank you for your contribution!

frontend/templates/iframe.ejs
```diff
@@ -130,15 +130,17 @@
...
         }
       }, 200);
 
       window.addEventListener('message', receiveMessages);
 
       function receiveMessages(event) {
         try {
-          const data = typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
+          const data = event.data;
+          const isObj = typeof data === 'object' && data !== null && !Array.isArray(data);
```
Yuriy-Karpov, line 142: sorry to interfere, but this code is repeated, I think it should be put in a separate function in utils

akellbl4: fixed ;)

akellbl4:
@Mavrin could you please review the PR? I've added some additional code, could be cool to have second opinion on it :)

Mavrin (approved):
lgtm

umputun:
@akellbl4 can you rebase all of this intto a single commit so I could merge it nicely?

akellbl4:
@Elephmoon hi, do you need help with it?

Elephmoon:
> @Elephmoon hi, do you need help with it?

Sorry. I’m moving to other city right now, I believe I have time today evening and I will solve it

Elephmoon:
Ugh. I made incorrect rebase and lost all changes (;

akellbl4:
@Elephmoon I'll try to revert it ;)

akellbl4:
@Elephmoon I've fixed it, Is it okay for you to merge it with squash?

Elephmoon:
Yep. I squashed it

akellbl4:
@umputun good to merge ;)

PR #911, opened 2021-03-11
Fix saving comment to LS

akellbl4:
**The problem**
When local storage is empty and we try to upload an image and send a form shows an error but the comment saves.

**Why**
Comment saves to local storage on edit.
When a user uploads an image comment doesn't save in local storage.
On send, we are trying to remove a saved comment from local storage but it's empty.
`delete data[....]` when data is `null`

**Solve**
* save comment after image upload
* do not affect error messages by saving to LS mechanism
* don't try to clear empty local storage

### DIFF
--- a/frontend/app/common/local-storage.ts
+++ b/frontend/app/common/local-storage.ts
@@ -46,9 +46,9 @@ export function setJsonItem<T = unknown>(key: string, data: T) {
   }
 }
 
-export function updateJsonItem<T extends {}>(key: string, value: (data: T) => T): void;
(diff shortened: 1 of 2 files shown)


umputun:
I think the issue was on saving image-only comments, i.e., without any text. At least this was my experience

akellbl4:
@umputun yep, If you upload only images and add no text you can lose your comment after refreshing the page.

akellbl4:
@Mavrin could you please leave a review this PR?

Mavrin (approved):

PR #919, opened 2021-03-18
corrected path in docker command to import comments from disqus

mipimipi:
When importing a file with comments in Disqus format, the correct path of that file for is `/srv/var/{disqus-export-name}.xml`. This should be made clear in the command template.

Signed-off-by: Michael Picht <mipi@fsfe.org>


umputun:
thx

PR #934, opened 2021-03-29
Support for Slack notifications

ronanj:
Implementation for the feature https://github.com/umputun/remark42/issues/924


backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (t *Slack) findChannelIDByName(name string) (string, error) {
+
+	params := slack.GetConversationsParameters{}
+	for {
+
+		chans, next, err := t.client.GetConversations(&params)
+		if err != nil {
+			log.Printf("[WARN] can't get the channel list: %v", err)
```
umputun, line 99: is WARN logging here even needed? if error returned handling the error usually done by the caller

backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		for _, channel := range chans {
+			fmt.Printf("ID: %s, Name: %s\n", channel.ID, channel.Name)
+			if channel.Name == name {
+				return channel.ID, nil
+			}
+		}
+
+		if next != "" {
```
umputun, line 110: by flipping the condition we can simplify it and get rid of else part

backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+	channelID string
+	client    *slack.Client
+}
+
+// NewSlack makes Slack bot for notifications
+func NewSlack(token, channelName string) (*Slack, error) {
+	return newSlack(token, channelName, 0, "")
+}
+
+func newSlack(token, channelName string, _ time.Duration, apiURL string) (*Slack, error) {
```
umputun, line 24: is there are a good reason for a separate method here? The only one I can think of is the ability to test it, but I'd rather see apiURL as a private struct filed and setting the value directly. Also, if time.Duration not used why it is even here? If this duration defines the timeout we definitely need it as we don't want to wait forever on the submission

umputun (changes requested):
Thx for the PR. A few suggestions

backend/app/cmd/server.go
```diff
@@ -807,6 +811,12 @@ func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *
 	var destinations []notify.Destination
 	for _, t := range s.Notify.Type {
 		switch t {
+		case "slack":
+			tg, err := notify.NewSlack(s.Notify.Slack.Token, s.Notify.Slack.Channel)
```
paskal, line 815: "tg" below is "telegram", so I propose to use different variable name there.

backend/app/cmd/server.go
```diff
@@ -211,6 +211,10 @@ type NotifyGroup struct {
 		VerificationSubject string `long:"verification_subj" env:"VERIFICATION_SUBJ" description:"verification message subject"`
 		AdminNotifications  bool   `long:"notify_admin" env:"ADMIN" description:"notify admin on new comments via ADMIN_SHARED_EMAIL"`
 	} `group:"email" namespace:"email" env-namespace:"EMAIL"`
+	Slack  struct {
+		Token   string        `long:"token" env:"TOKEN" description:"slack token"`
+		Channel string        `long:"chan" env:"CHAN" description:"slack channel"`
```
paskal, line 216: If I am not mistaken, a channel is not a channel name but some unique ID, if it's the case I would propose to add a note both here and in the Readme

backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+	if apiURL != "" {
+		opts = append(opts, slack.OptionAPIURL(apiURL))
+	}
+
+	client := slack.New(token, opts...)
+	res := &Slack{client: client}
+
+	channelID, err := res.findChannelIDByName(channelName)
+	if err != nil {
+		return res, err
```
paskal, line 40: Here we shouldn't return res and should wrap error with some text about where the error comes from.

backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+	log.Printf("[DEBUG] create new slack notifier for chan %s", channelID)
+
+	return res, nil
+}
+
+// Send to Slack channel
+func (t *Slack) Send(ctx context.Context, req Request) error {
+
+	if t.channelID == "" {
+		return errors.New("client is not connected")
```
paskal, line 53: It's a private field and must be set by that point in time. I think there is no point in having an extra check for that field in Send.

backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+
+	}
+	log.Printf("[DEBUG] send telegram notification, comment id %s", req.Comment.ID)
```
paskal, line 56: should be "slack notification"

backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+
+	author := req.Comment.User.Name
+	if req.Comment.ParentID != "" {
+		author += " → " + req.parent.User.Name
```
paskal, line 60: It's not an author actually, isn't it? I think that variable should be named differently.

backend/app/notify/slack.go
```diff
@@ -0,0 +1,117 @@
...
+
+}
+
+// SendVerification is not implemented for Slack
+func (t *Slack) SendVerification(_ context.Context, _ VerificationRequest) error {
+	return nil
+}
+
+func (t *Slack) String() string {
+	return "Slack@" + t.channelID
```
paskal, line 89: I think it might be useful to store (and print) the original channel name as well.

README.md
```diff
@@ -17,7 +17,7 @@ Remark42 is a self-hosted, lightweight, and simple (yet functional) comment engi
 * Images upload with drag-and-drop
 * Extractor for recent comments, cross-post
 * RSS for all comments and each post
-* Telegram and email notifications
+* Telegram, Slack and email notifications
```
paskal, line 20: I think by now we should better add clarification that Telegram and Slack and email are available for admins, and email is the only one available to users at the moment. Would be nice if you would do that documentation change as well.

paskal (changes requested):
Thanks a lot! Looks good except for few minor things here and there.

backend/app/notify/slack.go
```diff
@@ -12,32 +11,24 @@ import (
...
-	}
-
 	client := slack.New(token, opts...)
-	res := &Slack{client: client}
+	res := &Slack{client: client, channelName: channelName}
 
 	channelID, err := res.findChannelIDByName(channelName)
 	if err != nil {
-		return res, err
+		return res, errors.Wrap(err, "can not find slack channel '"+channelName+"'")
```
paskal, line 31: Shouldn't return `res` in case of error.

backend/app/notify/slack.go
```diff
@@ -86,7 +73,7 @@ func (t *Slack) SendVerification(_ context.Context, _ VerificationRequest) error
 }
 
 func (t *Slack) String() string {
-	return "Slack@" + t.channelID
+	return "slack: " + t.channelName
```
paskal, line 76: I think there is some benefit in knowing the channel ID as well, I propose to have them both returned here.

paskal (approved):

paskal:
Code looks good to me, however, I haven't run it and haven't tested actual work with slack.

backend/app/notify/slack.go
ronanj: I have refactored accordingly. The `NewSlack` function now takes as optional arguments the slack configuration.

umputun:
I wanted to test it and even created a slack account for this. However, I have no clue where to get a token. Can you add a file with some how-to about setup connection to slack? Just drop it here https://github.com/umputun/remark42/tree/master/docs/latest, thx

docs/latest/slack.md
```diff
@@ -4,14 +4,37 @@ title: Slack
 
 ## Slack notifications
 
-In order to integrate notifications from remark42 with the [slack](https://slack.com), you should creare [a bot](https://slack.com/intl/en-cn/help/articles/115005265703-Create-a-bot-for-your-workspace) and obtain a token. This token should be used as `NOTIFY_SLACK_TOKEN`. 
+In order to integrate notifications from remark42 with the [slack](https://slack.com), you should creare [a bot](https://slack.com/intl/en-cn/help/articles/115005265703-Create-a-bot-for-your-workspace) and obtain a token.
```
paskal, line 7:
```suggestion
In order to integrate notifications from remark42 with the [slack](https://slack.com), you should create [a bot](https://slack.com/intl/en-cn/help/articles/115005265703-Create-a-bot-for-your-workspace) and obtain a token. 
```

ronanj: Thanks. updated

umputun:
I was about to merge it but noticed it is a lot of commits and looks like some merge inside from the master, which is confusing. Could you pls rebase and reduce it to just a couple of commits? One with everything except dependencies and another one with dependencies

paskal (approved):

PR #936, opened 2021-03-30
#add auth source title international language support

johnwiichang:
with zh.json and en.json.


akellbl4 (approved):
@johnpuddephatt thanks for your contribution. I forgot about that text :)

Issue #898, opened 2021-02-23
Replace `parseQuery` with `URL`


akellbl4:
Remove `parseQuery` from the project and use browser API such as  `URL` or `URLSearchParams`. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
It is [placed here](https://github.com/umputun/remark42/blob/master/frontend/app/utils/parseQuery.ts). We already started to use it [here](https://github.com/umputun/remark42/blob/c4832ff47a7a7929c30330e7ef0ba6613b1b4014/frontend/app/common/fetcher.ts#L39)

Paras0-7:
Hii @akellbl4  May I  work on it.

PR #943, opened 2021-04-05
#898 Replace parseQuery with URLSearchParams

aepetelin:
Changes parseQuery implementation to URLSearchParams. 
I'm using .forEach() because not all browsers support URLSearchParams.entries()

### DIFF
--- a/frontend/app/utils/parseQuery.ts
+++ b/frontend/app/utils/parseQuery.ts
@@ -1,18 +1,9 @@
 /** converts widnow.location.search into object */
+
 export default function parseQuery<T extends {}>(search: string = window.location.search): T {
-  if (search.length < 2) {
-    return {} as T;
-  }
-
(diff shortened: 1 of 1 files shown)


Guria:
@umputun do we have to support ie11 in remark42?

umputun:
> @umputun do we have to support ie11 in remark42?

I don't really see why we should care about the officially deprecated browser. So, I'd say - no, we don't have to support ie11

Guria (approved):

Guria (approved):
Lgtm since we don't support ie11

akellbl4 (approved):

akellbl4:
@Guria we have two types of bundles for modern and old browsers. The modern bundle will be fresh and small and the bundle for the old browsers will be bigger with polly but that's fine. We expected it :)

PR #968, opened 2021-05-03
Update translations

dmitry-konchalenkov:
Update translations for 
- German 
- Spanish
- French
- Russian
- Turkish
- Chinese (Simplified)


paskal:
Do you possess these languages or used machine translations to update them?

Also, would you be so kind as to move new languages into separate MR and generate the code to enable them properly in it? For new languages, the mere presence is better than the lack of the language, so we'll merge them quickly. For existing languages changes, we'll need to check if the new translations are better than the existing ones, which will require more effort from reviewers and will take more time.

dmitry-konchalenkov:
Hi @paskal,
>Do you possess these languages or used machine translations to update them?

These translations were carried out by professional translators (native speakers) within [JetBrains](https://github.com/JetBrains/). :) 

>Also, would you be so kind as to move new languages into separate MR and generate the code to enable them properly in it?

Sure, I've removed new translations from this PR and created #969 (Add translations for Japanese, Korean, Brazilian Portuguese). Unfortunately I don't have Docker/npm on my machine, perhaps someone from the community will be able to use #969 and add new translations at some point. Thanks!

akellbl4 (approved):
@dmitry-konchalenkov thank you for such contribution!

umputun (approved):
Impressive work, thx

PR #976, opened 2021-05-05
add type module while load modern js bundle

akellbl4:
Now we have two requests for the js bundle because we don't add `type="module"` when requesting it.
This PR is fixing it.


Mavrin (approved):

umputun:
conflict pls?

Mavrin (approved):

PR #981, opened 2021-05-05
make form available after failed commit sending

### LINKED ISSUE #965: Invalid comments cannot be corrected and send
I have RESTRICTED_WORDS in my configuration:

```
RESTRICTED_WORDS=covfefe
```

Then if I try to send a comment with this word, I get the **Invalid comment data.** as expected.

But the textarea and the Send button are disabled and I cannot edit my comment, I'm stuck and need to refresh the page.

https://user-images.githubusercontent.com/225704/116722653-f6009800-a9de-11eb-8420-2b2c8ea841e9.mp4

I think that in this case we should be able to edit and send.

Thanks!

akellbl4:
Resolves #965

- add a new error code on the server for proper indication on the frontend
- simplify errors map on the frontend in favor of simple and small code


umputun (approved):
backend change makes sense

Mavrin (approved):

PR #894, opened 2021-02-22
bump deps

akellbl4:
- ~~update npm to v7~~ will do it in 1.8 release
- remove additional prettier presets because they were included inside the default preset

PR #982, opened 2021-05-05
use named exports

akellbl4:
I noticed an inconsistency when modules are using default naming. Even I sometimes rename imported module and then it's hard to read or refactor code. [Here are](https://basarat.gitbook.io/typescript/main-1/defaultisbad) other reasons why it's good.
I tried to use named imports and check if it makes an impact on bundle size and it's not.


Mavrin (dismissed):
👍

umputun:
another conflict

akellbl4:
We need to merge #894 because it could increase the number of the conflicts and fix them after it.

PR #977, opened 2021-05-05
minify css bundle

akellbl4:
Add minification for the CSS files.

public/remark.css                 9.42 KB (-6.47% 🔽)
public/last-comments.css   5.05 KB (-11.74% 🔽)

PR #986, opened 2021-05-06
minify css with csso

akellbl4:
Created for comparison with #977


akellbl4:
CleanShot 2021-05-06 at 23 56 57@2x

akellbl4:
csso lose at least on default settings :(

PR #991, opened 2021-05-07
replace deprecated golangci-lint check, fix goleak reports

paskal:
Before that fix running individual tests I changed in that PR produced goleak reports, they are not visible on a full test run likely because enough time passes for the client to die on its own.


umputun:
maybe we want to change this as well? https://github.com/umputun/remark42/blob/master/backend/.golangci.yml#L78

paskal:
Done.

PR #992, opened 2021-05-07
Renew image time to cleanup on comment Preview and Submit

### LINKED ISSUE #909: Image get cleaned up while waiting for CommitTTL to expire
## How to reproduce
Copy [image](https://github.com/dropbox/kglb/blob/master/doc/kglb_open_arch.jpg) to clipboard and paste it In the comment post. Image renders properly on preview and after "Post" action. But after page reload post will contain a placeholder "can't load" image. 

## Details and examples

Post example
<https://radio-t.com/p/2021/03/09/prep-745/#comments>

image

Problematic image xpath  `//*[@id="remark42__comment-6d119ea8-5d60-4302-8e45-53413f5c4b61"]/div[2]/div[1]/p[3]`: 
```html
<img src="https://remark42.radio-t.com/api/v1/picture/github_3d624f3e7d0d60cfcfa99b506fe245f1b678794f/c14hklqhajfpucec9kdg" alt="image.png" loading="lazy">
```

Response <https://remark42.radio-t.com/api/v1/picture/github_3d624f3e7d0d60cfcfa99b506fe245f1b678794f/c14hklqhajfpucec9kdg>:
```json
{
    "code": 18,
    "details": "can't get image github_3d624f3e7d0d60cfcfa99b506fe245f1b678794f/c14hklqhajfpucec9kdg",
    "error": "can't get image file for github_3d624f3e7d0d60cfcfa99b506fe245f1b678794f/c14hklqhajfpucec9kdg: can't get image stats for github_3d624f3e7d0d60cfcfa99b506fe245f1b678794f/c14hklqhajfpucec9kdg: stat var/pictures.staging/github_3d624f3e7d0d60cfcfa99b506fe245f1b678794f/86/c14hklqhajfpucec9kdg: no such file or directory"
}
```

paskal:
That PR will delay the image clean up after it's loaded by preview, so the commenter would have more time to post it before cleanup removes it.

Also, reset the cleanup time on images in the submitted comment to prevent the situation when they are deleted from temporary storage before being move to the permanent one, waiting for comment EditDuration to expire. That resolves #909.

### DIFF
--- a/backend/_example/memory_store/accessor/image.go
+++ b/backend/_example/memory_store/accessor/image.go
@@ -45,6 +45,17 @@ func (m *MemImage) Save(id string, img []byte) error {
 	return nil
 }
 
+// ResetCleanupTimer resets cleanup timer for the image
+func (m *MemImage) ResetCleanupTimer(id string) error {
+	m.Lock()
+	defer m.Unlock()
+	if _, ok := m.insertTime[id]; ok {
+		m.insertTime[id] = time.Now()
+		return nil
+	}
+	return errors.Errorf("image %s not found", id)
+}
+
 // Load image by ID
 func (m *MemImage) Load(id string) ([]byte, error) {
 	m.RLock()
--- a/backend/_example/memory_store/server/image.go
+++ b/backend/_example/memory_store/server/image.go
@@ -28,6 +28,16 @@ func (s *RPC) imgSaveWithIDHndl(id uint64, params json.RawMessage) (rr jrpc.Resp
 	return jrpc.EncodeResponse(id, nil, err)
 }
 
+func (s *RPC) imgResetClnTimerHndl(id uint64, params json.RawMessage) (rr jrpc.Response) {
+	var fileID string
+	if err := json.Unmarshal(params, &fileID); err != nil {
+		return jrpc.Response{Error: err.Error()}
+	}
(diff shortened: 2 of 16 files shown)


umputun:
I see what you did and this solution can probably work but there are a couple of things I don't like:

- the idea to make this countdown reset a part of storage feels odd. This is some kind of temporary state and keeping it in the real long-term store gives us only one benefit - the ability to survive reboots and, as a nice side effect, distributed mode (if the store supports it). I don't think we really care about the first one, and for the second - some sort of cache seems to be a closer match
- The implementation not insanely complex but still too much for what we are trying to achieve here.

_ as I opened the ticket my idea, was to isolate all of it on the image and service levels. i.e. some sort of event bus from rest to image communicating change event. I'm not sure if this even possible as I have not tried to write the actual code, just a feeling. Maybe moving this countdown store closer to the image will do it_

paskal:
If the staging storage is part of the image.Store interface as it is now, we can't alter its commit and cleanup behaviour other than the way I did it above. The only way around is I can think of is implementing separate staging storage for images in which will be independent of the normal one, then we'll have commit and cleanup and cleanup timer reset functionality in that separate package and not in three of them as of now.

What do you think?

umputun (approved):
LGTM

PR #998, opened 2021-05-13
Validate image existence before post or preview

paskal:
That should cover the case when an image is posted after it cleaned up or the URL is altered by mistake: in case image belongs to this site, it should be working at the time of preview or comment post.

### DIFF
--- a/backend/app/store/image/image.go
+++ b/backend/app/store/image/image.go
@@ -148,6 +148,8 @@ func (s *Service) Submit(idsFn func() []string) {
 }
 
 // ExtractPictures gets list of images from the doc html and convert from urls to ids, i.e. user/pic.png
+// It doesn't return possible errors parsing the cached images, and will try to return as many parsed ids
+// as possible instead.
 func (s *Service) ExtractPictures(commentHTML string) (ids []string, err error) {
 
 	doc, err := goquery.NewDocumentFromReader(strings.NewReader(commentHTML))
--- a/backend/app/rest/api/rest_private.go
+++ b/backend/app/rest/api/rest_private.go
@@ -88,6 +88,20 @@ func (s *private) createCommentCtrl(w http.ResponseWriter, r *http.Request) {
 	}
 	comment = s.commentFormatter.Format(comment)
 
+	// check if images are valid
+	imgIds, err := s.imageService.ExtractPictures(comment.Text)
+	if err != nil {
+		rest.SendErrorJSON(w, r, http.StatusBadRequest, err, "can't extract pictures from comment text", rest.ErrCommentValidation)
+		return
+	}
+	for _, id := range imgIds {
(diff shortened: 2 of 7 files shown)


umputun (approved):
LGTM. Won't provide much details to the user (like what picture caused a problem in case of multiple) but won't hurt for sure

umputun (changes requested):
oops, noticed image related tested failed

paskal:
@umputun now it's properly working both in reality and tests, please review and merge if everything is OK.

backend/app/store/image/image.go
```diff
@@ -164,22 +164,28 @@ func (s *Service) ExtractPictures(commentHTML string) (ids []string, err error)
...
+				imgID, e := CachedImgID(string(imgURL))
+				if e != nil {
+					err = errors.Wrapf(e, "problem retrieving cached URL for %s", im)
 					return
 				}
 				ids = append(ids, imgID)
 			}
 		}
 	})
+	if err != nil {
```
umputun, line 186: I'm not sure if the error check here safe/correct. I mean, extraction before this change worked even for partially valid document and returned some images anyway. I think it was intentional, and now we may have all or nothing

paskal: Now it's all or nothing, but there is no way you get an error on `submitImages` (the only function using that function before the PR) after comment sending now because post sending calls the same function and won't allow posting incorrect images.

I don't recall an intention of ignoring the error when I wrote cache image extraction, when I was preparing that PR I was surprised by the bare return inside that function returning null instead of `err`

umputun: hmm, the function ExtractPictures returned an error only in one case - goquery.NewDocumentFromReader failed. Everything else was forgiving and func tried to extract as many pictures as possible from the comment. After this change instead of valid pictures, submitImages won't get anything if one of the images failed, and idsFn will return an empty slice. Are you sure this behavior valid?

Is this change even relevant to the PR? I don't see how changing the logic of ExtractPictures is necessary for the proper validation.

paskal: Without that change, we do not validate cached images. But they are available in the comment only on edit and not on the post and we don't validate edits in this PR.

I think you are right and it would be better to revert that function change and document it better instead.

paskal: Now, as we talk about it, I started wondering if we should return the error at all: it's confusing to have an error of HTML parsing returned to the caller but cached image parsing errors hidden in the code. Making it errorless will clarify the intention of returning ids at any cost, what do you think?

umputun (approved):

backend/app/store/image/image.go
umputun: the error it returns now is kind of "panic case, should never happen". Logging this thing as we do is ok strategy, just in case.

PR #1008, opened 2021-05-17
Multiple improvements to image service code and tests

paskal:
These are changes from #992 which are not relevant to the purpose of that PR, cut into separate PR to simplify 992 code review.

- rename variables according to golangci-lint recommendation
- make consistent returns in bolt_store
- less magic consonants in tests
- make commitTTL equal to EditDuration (opposed to 1.5 previously), so that image is committed to permanent storage after the comment can no longer be edited.
- move cleanupTTL to the Cleanup function, as it's not used elsewhere in the code.

I advise merging this with rebase.


backend/app/store/image/image.go
```diff
@@ -186,17 +175,18 @@ func (s *Service) ExtractPictures(commentHTML string) (ids []string, err error)
 	return ids, nil
 }
 
-// Cleanup runs periodic cleanup with cleanupTTL. Blocking loop, should be called inside of goroutine by consumer
+// Cleanup runs periodic cleanup with 2.5*ServiceParams.EditDuration. Blocking loop, should be called inside of goroutine by consumer
 func (s *Service) Cleanup(ctx context.Context) {
-	log.Printf("[INFO] start pictures cleanup, staging ttl=%v", s.cleanupTTL)
+	cleanupTTL := s.EditDuration * 25 / 10 // cleanup images older than 1.5 * EditDuration
```
umputun, line 180: well, 25/10 is 2.5 not 1.5. Maybe just do direct *1.5 (or *2.5)?

umputun (changes requested):
looks ok, except the cleanupTTL confusion

backend/app/store/image/image.go
paskal: Fixed. There is type confusion otherwise, that is a way to keep numbers straight without types conversions.

umputun (approved):
LGTM

PR #1009, opened 2021-05-17
Remove error return from ExtractPictures

paskal:
That function returns an error in a never expected condition, and that error would be logged message on the caller side: none of the callers handles it.

That change hides that error from the caller so that function would have a signature that better fit what it does and how it behaves.


umputun (approved):
makes sense

Issue #707, opened 2020-05-08
[Feature Request] Telegram Login


MasterGroosha:
It would be nice to have "Login via Telegram" support:
https://core.telegram.org/widgets/login

umputun:
I have checked if this login can be implemented as a part of our [auth library](https://github.com/go-pkgz/auth) and the way telegram does it seems very odd and problematic. From what I can see their script renders "Log in with Telegram" button inside the iframe with another extra script that loads some Telegram entities to work with.

I think, this API  is not something useful to open authorization windows and do redirects to specified url, as any sane oauth provider does.

However, if anyone can figure a way to integrate this flow, pls provide PR to https://github.com/go-pkgz/auth

umputun:
TG login is a part of go-pkgz/auth now (see above), so the only thing left is the UI part and a couple of params on the backend side.

uPagge:
@paskal I understand correctly that there is no way to implement telegram-based authorization yet. I just sort of did everything according to the documentation, but the telegram button did not appear

PR #1021, opened 2021-05-24
Move telegram token and timeout to a separate CLI section

paskal:
This simplifies token and timeout reuse for `notify` module (used now) and for `auth` module later (not yet in the code). SMTP credentials are already set up that way.

Part of the work on #707.


umputun (approved):
makes sense

PR #1027, opened 2021-05-25
Separate user and admin notifications

paskal:
The current state is a mess of user and admin notifications, which will become worse after implementing the new user notification methods like a telegram.

This change makes things simpler for the remark42 users.

Can be merged after #1026 will be merged.


umputun:
I found the naming to be extremely confusing. As I read UserTypes my brain immediately translates it to "types of users". Not sure what will be the better name here

PR #1029, opened 2021-05-26
Telegram notifications for users

### LINKED ISSUE #830: Feature: Telegram Notify
The way the notifications are now made in telegrams is quite convenient for site administrators, but not at all convenient for users.

It would be cool to do, like on vas3k.club.

There you can subscribe for notifications to a separate post, as well as receive notifications to your comments.

paskal:
Would add backend support #830.

Verification message:
image
Reply notification:
image

For frontend (@akellbl4): when config option `telegram_bot_username` is not empty, telegram notifications are enabled. We need to ask the user to write a bot with that username (https://tg.me/$username) and then use his ID for subscription: send `/getid` to https://tg.me/myidbot in Telegram messenger.


backend/app/cmd/server.go
```diff
@@ -494,28 +500,29 @@ func (s *ServerCommand) newServerApp() (*serverApp, error) {
...
+		ReadOnlyAge:         s.ReadOnlyAge,
+		SharedSecret:        s.SharedSecret,
+		Authenticator:       authenticator,
+		Cache:               loadingCache,
+		NotifyService:       notifyService,
+		SSLConfig:           sslConfig,
+		UpdateLimiter:       s.UpdateLimit,
+		ImageService:        imageService,
+		EmailNotifications:  emailNotifications,
+		TelegramBotUsername: telegramNotificationsBotUsername,
```
umputun, line 519: this looks like java ;) tgNotifBotName maybe?

backend/app/cmd/server.go
```diff
@@ -868,44 +875,55 @@ func (s *ServerCommand) loadEmailTemplate() (string, error) {
 	return string(file), nil
 }
 
-func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *auth.Service) (*notify.Service, error) {
+// aside from notify.Service and error, returns telegram bot name which will be passed to the frontend
+func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *auth.Service) (*notify.Service, string, error) {
```
umputun, line 879: this one getting too big func, maybe split ? if you can figure a nice way to have makeNotifyAdmin and makeNotifyUsers it could be easier to digest

backend/app/rest/api/rest_private.go
```diff
@@ -314,6 +317,55 @@ func (s *private) sendEmailConfirmationCtrl(w http.ResponseWriter, r *http.Reque
...
+	address := r.URL.Query().Get("address")
+	siteID := r.URL.Query().Get("site")
+	if address == "" {
+		rest.SendErrorJSON(w, r, http.StatusBadRequest,
+			errors.New("missing parameter"), "address parameter is required", rest.ErrInternal)
+		return
+	}
+	existingAddress, err := s.dataService.GetUserTelegram(siteID, user.ID)
+	if err != nil {
+		log.Printf("[WARN] can't read telegram for %s, %v", user.ID, err)
```
umputun, line 334: not following the loginc here. is it even make sense to continue afte  failed  GetUserTelegram? If it is, we need some comment/explanation

backend/app/rest/api/rest_private.go
```diff
@@ -314,6 +317,55 @@ func (s *private) sendEmailConfirmationCtrl(w http.ResponseWriter, r *http.Reque
...
+func (s *private) sendTelegramConfirmationCtrl(w http.ResponseWriter, r *http.Request) {
+	user := rest.MustGetUserInfo(r)
+	address := r.URL.Query().Get("address")
+	siteID := r.URL.Query().Get("site")
+	if address == "" {
+		rest.SendErrorJSON(w, r, http.StatusBadRequest,
+			errors.New("missing parameter"), "address parameter is required", rest.ErrInternal)
+		return
+	}
+	existingAddress, err := s.dataService.GetUserTelegram(siteID, user.ID)
```
umputun, line 332: btw, if i understant "tricky to use SendErrorJSON" part this is about incorrect frame report? this can be solved with an extra param "skipFrames" of some sort. Not saying we have to refactor this, just a note

backend/app/rest/api/rest_private.go
```diff
@@ -366,6 +418,46 @@ func (s *private) setConfirmedEmailCtrl(w http.ResponseWriter, r *http.Request)
...
+		rest.SendErrorJSON(w, r, http.StatusForbidden, err, "failed to verify confirmation token", rest.ErrInternal)
+		return
+	}
+
+	if s.authenticator.TokenService().IsExpired(confClaims) {
+		rest.SendErrorJSON(w, r, http.StatusForbidden, errors.New("expired"), "failed to verify confirmation token", rest.ErrInternal)
+		return
+	}
+
+	elems := strings.Split(confClaims.Handshake.ID, "::")
```
umputun, line 443: some comment about how this ID formed will help to figure the reason for the split. The orig setConfirmation doesn't have it either

backend/app/rest/api/rest_private_test.go
```diff
@@ -645,7 +656,7 @@ func TestRest_Email(t *testing.T) {
 	}
 }
 
-func TestRest_EmailNotification(t *testing.T) {
+func TestRest_EmailAndTelegramNotification(t *testing.T) {
```
umputun, line 659: I'm not sure if i like the idea of the mixed tests here. I'd rather prefer setup code extracted or even duplicated and moving tg tests to a separate funcs

README.md
```diff
@@ -156,7 +156,7 @@ _this is the recommended way to run remark42_
 | auth.email.subj         | AUTH_EMAIL_SUBJ         | `remark42 confirmation`  | email subject                                   |
 | auth.email.content-type | AUTH_EMAIL_CONTENT_TYPE | `text/html`              | email content type                              |
 | auth.email.template     | AUTH_EMAIL_TEMPLATE     | none (predefined)        | custom email message template file              |
-| notify.users            | NOTIFY_USERS            | none                     | type of user notifications (email)              |
+| notify.users            | NOTIFY_USERS            | none                     | type of user notifications (telegram, email)    |
```
umputun, line 159: probably we need some detailed info about user's tg notif somewhere

umputun (changes requested):
just a quick view, plannig to go deeper later on

backend/app/rest/api/rest_private.go
paskal: Nope, just we'll need to have a nested function which also could potentially send an answer to the user (SendErrorJSON) so it would need request and response passed to it, and some boolean flag to check if we can carry on with our logic or that function took care about it for us and we need to just return immediately. It will hurt readability more than code duplication in my opinion.

README.md
paskal: Please elaborate on this one, documentation note on what is the trickiness of telegram notify?

backend/app/rest/api/rest_private_test.go
paskal: Addressed in the second commit.

backend/app/rest/api/rest_private.go
paskal: Addressed in the second commit.

backend/app/rest/api/rest_private.go
paskal: Addressed in the second commit.

backend/app/cmd/server.go
paskal: I didn't figure a good way to split it, however I was able to chop around 20 lines off it in a second commit I made, I hope now it's better than before.

backend/app/cmd/server.go
paskal: Got rid of this helper variable in the second commit for good.

paskal:
I've fixed most of the review comments in the second commit for the reviewer convenience, looking forward to the next review round.

umputun (approved):
LGTM

PR #1038, opened 2021-06-07
Add UserDetailTelegram support

paskal:
For telegram notifications we'll need to get a telegram user ID from the user and store and retrieve it later, this PR adds support for that user details and improves a bunch of tests in affected modules.

All newly added code is tested.

### DIFF
--- a/backend/app/notify/notify.go
+++ b/backend/app/notify/notify.go
@@ -35,13 +35,18 @@ type Destination interface {
 type Store interface {
 	Get(locator store.Locator, id string, user store.User) (store.Comment, error)
 	GetUserEmail(siteID string, userID string) (string, error)
+	GetUserTelegram(siteID string, userID string) (string, error)
 }
+
+// used for email and telegram retrieval from user details
+type getUserDetail func(string, string) (string, error)
 
 // Request notification for a Comment
 type Request struct {
-	Comment     store.Comment
-	parent      store.Comment
-	Emails      []string
+	Comment   store.Comment
+	parent    store.Comment
+	Emails    []string
+	Telegrams []string
 }
 
 // VerificationRequest notification for user
@@ -84,7 +89,8 @@ func (s *Service) Submit(req Request) {
 	if s.dataService != nil && req.Comment.ParentID != "" {
 		if p, err := s.dataService.Get(req.Comment.Locator, req.Comment.ParentID, store.User{}); err == nil {
 			req.parent = p
-			req.Emails = deduplicateStrings(s.getNotificationEmails(req, p))
+			req.Emails = s.getNotificationTargets(req, p, s.dataService.GetUserEmail)
+			req.Telegrams = s.getNotificationTargets(req, p, s.dataService.GetUserTelegram)
 		}
 	}
 	select {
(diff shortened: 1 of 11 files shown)


paskal:
@akellbl4 do you know why codecov is running on that MR which doesn't change anything in the frontend and is reporting `frontend/app/components/root/root.tsx` problems?

backend/app/notify/notify.go
```diff
@@ -94,22 +97,27 @@ func (s *Service) Submit(req Request) {
...
-// Emails is not added to the returned list in case original message is from the same user as the notification receiver.
-func (s *Service) getNotificationEmails(req Request, notifyComment store.Comment) (result []string) {
+// getNotificationDetails returns list of details (like email or telegram username) for users
+// interested in notifications for provided comment.
+// Details are not added to the returned list in case the original message is from the same user as the notification receiver.
+func (s *Service) getNotificationDetails(
+	req Request,
+	notifyComment store.Comment,
+	getUserDetail func(string, string) (string, error),
+) (result []string) {
```
umputun, line 107: this func is a little bit confusing. Ptobably definig a type for this getUserDetail will make it better looking. Also naming - it doesn't return much of the details but rather a list of destinations/targets

backend/app/notify/notify.go
```diff
@@ -35,13 +35,15 @@ type Destination interface {
...
 
 // Request notification for a Comment
 type Request struct {
-	Comment     store.Comment
-	parent      store.Comment
-	Emails      []string
+	Comment       store.Comment
+	parent        store.Comment
+	Emails        []string
+	TelegramUsers []string
```
umputun, line 46: is it really users? i.e. can this be a group/channel id? Maybe call it "Telergams"?

umputun (changes requested):
I have not checked much of the logic as it seems to me mostly an extension of the existing user email part of the managment. Just a couple of minor things

backend/app/notify/notify.go
paskal: Renamed, moved `getUserDetail` to a separate type with a comment.

backend/app/notify/notify.go
paskal: Renamed to the Telegrams, I agree with your point. I think it's technically possible to add user notifications to the channel (however, it would be quite a strange idea).

backend/app/notify/notify.go
```diff
@@ -94,25 +100,32 @@ func (s *Service) Submit(req Request) {
 	}
 }
 
-// getNotificationEmails returns list of emails for notifications for provided comment.
-// Emails is not added to the returned list in case original message is from the same user as the notification receiver.
-func (s *Service) getNotificationEmails(req Request, notifyComment store.Comment) (result []string) {
+// getNotificationTargets returns list of notification targets (like email or telegram username) for users
```
paskal, line 103: Updated the comment to reflect the new name, made results deduplicated to prevent the caller from doing it.

backend/app/store/engine/engine.go
```diff
@@ -109,8 +111,9 @@ type UserDetail string
 
 // UserDetailEntry contains single user details entry
 type UserDetailEntry struct {
-	UserID string `json:"user_id"`         // duplicate user's id to use this structure not only embedded but separately
-	Email  string `json:"email,omitempty"` // UserEmail
+	UserID   string `json:"user_id"`            // duplicate user's id to use this structure not only embedded but separately
+	Email    string `json:"email,omitempty"`    // UserEmail
+	Telegram string `json:"telegram,omitempty"` // UserEmail
```
umputun, line 116: bad copy-paste, should be UserTelegram

umputun (changes requested):
just a comment to fix and good to go

backend/app/store/engine/engine.go
paskal: Fixed.

PR #1042, opened 2021-06-12
Remove ability to set telegram API, clarify telegram notify params

paskal:
As discussed off-GitHub, remove the ability to set Telegram API (as it's used only in tests) and notify parameters passed as structure in the same way we do with Email notifications now.

### DIFF
--- a/backend/app/notify/telegram.go
+++ b/backend/app/notify/telegram.go
@@ -16,38 +16,44 @@ import (
 	"github.com/pkg/errors"
 )
 
+// TelegramParams contain settings for telegram notifications
+type TelegramParams struct {
+	AdminChannelID string        // unique identifier for the target chat or username of the target channel (in the format @channelusername)
+	Token          string        // token for telegram bot API interactions
+	Timeout        time.Duration // http client timeout
+
+	apiPrefix string // changed only in tests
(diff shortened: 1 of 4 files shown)


backend/app/notify/telegram.go
```diff
@@ -16,38 +16,44 @@ import (
 	"github.com/pkg/errors"
 )
 
+// TelegramParams contain settings for telegram notifications
+type TelegramParams struct {
```
umputun, line 20: what the reason for adding this? 3-4 params directly in the constuctor looks fine to me

umputun (changes requested):
just a minror question

backend/app/notify/telegram.go
paskal: Hiding API, which is used only for tests, and adding UserNotifications param in PR for a telegram user notifications, which will raise the count of params by one: I felt silly adding the new param to all the tests which even don't require it, five params to function felt like too many.

umputun: I didn't mean passing that apiPrefix via constructor but rather setting it directly in tests. anyway, it doesn't bother me enough, so fine with me ;)

PR #1043, opened 2021-06-13
Don't set X-XSRF-TOKEN when the user isn't logged in

forkbomb9:
An HTTP header cannot be empty, and although some webservers allow this (nginx, Apache), others answer 400 Bad Request (lighttpd), preventing the widget from loading.

I'm not sure this is the most elegant way to do it, that's why this is marked as a draft.

### DIFF
--- a/frontend/app/common/fetcher.ts
+++ b/frontend/app/common/fetcher.ts
@@ -45,7 +45,13 @@ const createFetcher = (baseUrl: string = ''): Methods => {
     if (activeJwtToken) {
       headers[JWT_HEADER] = activeJwtToken;
     }
-    headers[XSRF_HEADER] = getCookie(XSRF_COOKIE) || '';
+
+    // An HTTP header cannot be empty.
+    // Although some webservers allow this (nginx, Apache), others answer 400 Bad Request (lighttpd).
+    const xsrfToken = getCookie(XSRF_COOKIE);
+    if (xsrfToken !== undefined) {
+      headers[XSRF_HEADER] = xsrfToken;
+    }
 
     if (body instanceof FormData) {
       // Shouldn't add any kind of `Content-Type` if we send `FormData`
--- a/frontend/app/common/fetcher.test.ts
+++ b/frontend/app/common/fetcher.test.ts
@@ -4,7 +4,7 @@ jest.mock('./settings', () => ({
 
 import { RequestError } from 'utils/errorUtils';
 import { API_BASE, BASE_URL } from './constants.config';
-import { apiFetcher, authFetcher, adminFetcher, JWT_HEADER, XSRF_HEADER } from './fetcher';
+import { apiFetcher, authFetcher, adminFetcher, JWT_HEADER } from './fetcher';
 
 type FetchImplementaitonProps = {
   status?: number;
@@ -31,7 +31,7 @@ function mockFetch({ headers = {}, data = {}, ...props }: FetchImplementaitonPro
 }
 
 describe('fetcher', () => {
-  const headers = { [XSRF_HEADER]: '' };
+  const headers = {};
   const apiUri = '/anything';
   const apiUrl = `${BASE_URL}${API_BASE}/anything?site=remark`;
 


gstrauss:
Cross-reference: https://github.com/opnsense/core/issues/4917

The next version of lighttpd (lighttpd 1.4.60, not yet released) will relax this unintentional restriction which surfaced in the lighttpd HTTP/2 implementation.

(It is still poor behavior for a client to send an HTTP request header with an empty value.)

forkbomb9:
Oh, nice! I hadn't found anything about that.

This bug had me banging my head to the desk for two days 😆 
Shall we still remove the X-XSRF-TOKEN header when it's empty?

gstrauss:
I agree that it is a good idea to skip sending an HTTP request header with an empty value.

paskal (dismissed):
Does make sense to me.

frontend/app/common/fetcher.ts
```diff
@@ -46,6 +46,9 @@ const createFetcher = (baseUrl: string = ''): Methods => {
       headers[JWT_HEADER] = activeJwtToken;
     }
     headers[XSRF_HEADER] = getCookie(XSRF_COOKIE) || '';
+    // An HTTP header cannot be empty.
+    // Although some webservers allow this (nginx, Apache), others answer 400 Bad Request (lighttpd).
+    if (headers[XSRF_HEADER] === '') delete headers[XSRF_HEADER];
```
gstrauss, line 51: @forkbomb9 why not store `getCookie(XSRF_COOKIE)` into a temporary value and test that it exists and is not empty before setting `headers[XSRF_HEADER]`.  That way, you can skip the `delete headers[XSRF_HEADER]`

umputun:
Generally speaking, this is not a bug on remark42 side, but rather some deficiency of the proxy (Lighttpd). I don't have anything against not sending a no-value header but either way I see no reason why such a header may break anything. 

Take a look at another project [reproxy](https://github.com/umputun/reproxy) by the same authors, this one won't do insane things like Lighttpd does ;)

akellbl4:
@gstrauss I don't think it's poor because if it's supported by browsers it should be supported by http servers.

But I agree with the PR. If we don't need to send some headers we shouldn't.

frontend/app/common/fetcher.ts
forkbomb9: Yep, will do that. I'll also fix the test suite

forkbomb9:
@akellbl4: Actually, with the previous behavior (empty header), if I use Firefox to re-send the request to the server (the one Lighttpd answered 400), Firefox removes the empty header.

@umputun: reproxy looks like a nice piece of software. I'll look into it!
But as @gstrauss said, this behavior will be fixed in Lighttpd 1.4.60.

akellbl4:
>  if I use Firefox to re-send the request to the server (the one Lighttpd answered 400), Firefox removes the empty header.

It's problem of Lighttpd :) As you can see Firefox accepts sending empty header.

forkbomb9:
> > if I use Firefox to re-send the request to the server (the one Lighttpd answered 400), Firefox removes the empty header.
> 
> It's problem of Lighttpd :) As you can see Firefox accepts sending empty header.

Because it's being sent from a JS file. But if it's _Firefox_ making the request, it removes it.
Another example:
```sh
$ curl -H "X-XSRF-TOKEN: " https://remark42.example.org/api/v1/config\?site=example
```
Curl doesn't send the header.
But as gstrauss said above, Lighttpd 1.4.60 will match Apache and nginx's behavior: ignore the header.

frontend/app/common/fetcher.ts
```diff
@@ -45,7 +45,13 @@ const createFetcher = (baseUrl: string = ''): Methods => {
     if (activeJwtToken) {
       headers[JWT_HEADER] = activeJwtToken;
     }
-    headers[XSRF_HEADER] = getCookie(XSRF_COOKIE) || '';
+
+    // An HTTP header cannot be empty.
+    // Although some webservers allow this (nginx, Apache), others answer 400 Bad Request (lighttpd).
+    const xsrfToken = getCookie(XSRF_COOKIE);
+    if (xsrfToken !== undefined) {
+        headers[XSRF_COOKIE] = xsrfToken;
```
akellbl4, line 53:
```suggestion
        headers[XSRF_HEADER] = xsrfToken;
```

forkbomb9: Oops, good catch! I didn't remember they were two different constants.

akellbl4 (dismissed):

paskal (changes requested):
Should we remove XSRF_HEADER from the import where the linter detected unused variable?

forkbomb9:
@paskal, fixed. Also fixed prettier's complain about the indentation.

paskal (approved):

akellbl4:
> Should we remove XSRF_HEADER from the import where the linter detected unused variable (frontend/app/common/fetcher.test.ts)?

We shouldn't

akellbl4 (approved):

umputun (approved):
thx

PR #1048, opened 2021-06-13
Multiple notify module improvements

paskal:
A couple of changes to telegram notifications and error handling code, email text improvements, ignore IntelliJ Idea rest requests variables file.

Reply notification text, now more consistent with the email notification, and the same which will be used for user notifications. Before:
image
After:
image

Part of #1029 is not strictly tied to the telegram user notifications, and I wanted to merge separately.

### DIFF
--- a/backend/app/notify/telegram.go
+++ b/backend/app/notify/telegram.go
@@ -30,15 +30,18 @@ type Telegram struct {
 	TelegramParams
 }
 
+// telegramMsg is used to send message trough Telegram bot API
+type telegramMsg struct {
+	Text      string `json:"text"`
+	ParseMode string `json:"parse_mode,omitempty"`
+}
+
 const telegramTimeOut = 5000 * time.Millisecond
 const telegramAPIPrefix = "https://api.telegram.org/bot"
 
 // NewTelegram makes telegram bot for notifications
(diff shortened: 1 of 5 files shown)


backend/app/notify/telegram.go
```diff
@@ -155,30 +162,42 @@ func (t *Telegram) sendMessage(ctx context.Context, b []byte, chatID string) err
...
 	b, err := json.Marshal(body)
 	if err != nil {
 		return nil, err
 	}
 	return b, nil
 }
 
-func escapeTitle(title string) string {
-	escSymbols := []string{"[", "]", "(", ")"}
+func escapeText(title string) string {
```
umputun, line 199: I'm not sure about this one. The original ticket for all this escaping indicated no improvement for the initial problem https://github.com/umputun/remark42/issues/839

unless you see something really helpful due to this, I suggest dropping it completely

umputun (commented):
not sure about such a rich escaping

backend/app/notify/telegram.go
paskal: I saw messages not sending without escaping, status code 400. We can't drop it according to my tests with the verify messages.

PR #1053, opened 2021-06-20
Move existing documentation to the new site

paskal:
@akellbl4 I can't run the site using docker-compose (styles are not generated), please check and verify that it's what you had in mind. Also, I see that it's easy to break the site by using something like "\`\`\`sh" instead of "\`\`\`shell", we should have a CI job verifying that site is still valid if we don't have one now.


site/src/docs/backup/migration/index.md
```diff
@@ -3,3 +3,30 @@ title: Migration
...
 order: 300
 ---
+
+Here is an example on how to move your comments after your posts are moved, from `https://example.org/blog/<slug>` to `https://example.org/post/<slug>` in that example.
+
+### Rules file
+
+First you have to create a `rules` file in remark's `/var` folder.
+
+[Here is the test](https://github.com/akosourov/remark/blob/4dc123dbe84f4f248864bcdbbd6cc2b3a4dafe11/backend/app/migrator/mapper_test.go#L13-L17) with an example of rules file syntax, format is simply `old_url new_url` like following:
```
akellbl4, line 13: I don't think it's a good idea to mention code. I think all of the information should be shown on the doc page.

site/src/docs/backup/migration/index.md
```diff
@@ -3,3 +3,30 @@ title: Migration
 parent: Backup
 order: 300
 ---
+
+Here is an example on how to move your comments after your posts are moved, from `https://example.org/blog/<slug>` to `https://example.org/post/<slug>` in that example.
```
akellbl4, line 7: Is this info duplicates `url-migration/index.md`?

site/src/docs/configuration/reproxy/index.md
```diff
@@ -1,10 +1,13 @@
 ---
 title: Reproxy
+menuTitle: Reproxy
```
akellbl4, line 3: it's unnecessary to add `menuTitle` if you don't want to redefine `title`. By default `title` is used as the menu title.

site/src/docs/getting-started/installation/index.md
```diff
@@ -69,4 +69,4 @@ Put the next code snippet on a page of your site where you want to have comments
 
 After that widget will be rendered inside this node.
 
-If you want to set this up on a Single Page App, see [appropriate doc page](https://github.com/umputun/remark42/blob/master/docs/spa.md).
+If you want to set this up on a Single Page App, see [appropriate doc page](https://github.com/umputun/remark42/blob/master/site/src/docs/getting-started/installation/index.md).
```
akellbl4, line 72: As I can see docs on SPA usage is placed in `site/src/docs/configuration/frontend/index.md`

akellbl4 (approved):

PR #1057, opened 2021-07-19
Improve README

rakleed:
* Update links
* Update table of contents
* Improve the design
* Improve formatting
* Fix abbreviations and proper names
* Fix tables
* Fix indentation in code snippets
* Fix typos

PR #1059, opened 2021-07-25
Don't run CI pipelines on frontend Readme file change

paskal:
On an example of #1057, I found that we run CI pipelines when Readme file changes in the frontend directory, which doesn't make sense.


akellbl4:
I don't think we need to run it on any `.md` file in frontend folder and in backend as well.

PR #1082, opened 2021-08-08
move subscribe ui elements

akellbl4:
- use icons from subscribing buttons
- show the RSS subscription button for unauthorized users
- show markdown support as an icon with tooltip
- tooltips for descriptions

CleanShot 2021-08-08 at 19 04 31@2x
CleanShot 2021-08-08 at 19 41 17@2x


umputun:
The "m1" button/label looks odd. Maybe move it to the right side ?

akellbl4:
@umputun I'll try it ;)

PR #1094, opened 2021-08-09
fix: constants

akellbl4:
- move once used const to the chunk
- remove unused const


Mavrin (approved):

akellbl4:
@umputun it's good to merge

PR #1107, opened 2021-08-17
Add telegram auth to frontend

### LINKED ISSUE #707: [Feature Request] Telegram Login

Ksinia:
Screenshot 2021-11-07 at 15 32 33
Screenshot 2021-11-07 at 15 31 32

Fixes #707.


akellbl4:
@Ksinia thank you for your contribution!
I need a bit more time to review it. I would test it and maybe propose a design for this feature.
I think Telegram should be shown as other third-party providers but we need to add special flow.

akellbl4 (changes requested):
Hey @Ksinia,
I'd propose a visual [design like this](https://www.figma.com/file/lIWEmnnVs919aB16Om1vgR/Remark42-%7C-Telegram-auth-flow?node-id=0%3A1)
Telegram is a third-party auth provider and I'd expect it in the line with others. Another reason we can't put the full name of the provider on the tabs.

CleanShot 2021-08-29 at 17 01 14@2x

paskal:
@akellbl4 default mock render wrapper for tests don't set up state and fail because of that, would you be so kind as to look into it? The wrapper needs to be rewritten to inject state as well as other places will benefit from it as well.

frontend/app/components/auth/auth.hooks.ts
```diff
@@ -33,7 +33,13 @@ export function useDropdown(disableClosing?: boolean) {
     }
 
     function handleClickOutside(evt: MouseEvent) {
-      if (disableClosing || dropdownElement?.contains(evt.target as HTMLDivElement)) {
+      if (
+        disableClosing ||
+        dropdownElement?.contains(evt.target as HTMLDivElement) ||
+        (evt.target as Element).classList?.contains('telegram-auth')
```
akellbl4, line 39: Seems hacky. I'll check if there is a better workaround for it

frontend/app/components/auth/components/oauth.tsx
```diff
@@ -1,29 +1,33 @@
...
 type Props = {
   providers: OAuthProvider[];
+  toggleTelegram?: (showTelegram: boolean) => void;
 };
 
-export function OAuth({ providers }: Props) {
+export function OAuth({ providers, toggleTelegram }: Props) {
   const intl = useIntl();
   const dispath = useDispatch();
+  const telegramParams = useSelector((s: StoreState) => s.telegramParams);
```
akellbl4, line 30: I don't see any reason for redux here. We handle all of the logic inside `Auth` component.

I would suggest adding `onClickProvider`  callback for `OAuth` and handling this case in the parent component.

frontend/app/components/auth/components/oauth.tsx
```diff
@@ -39,24 +43,47 @@ export function OAuth({ providers }: Props) {
...
-              rel="noopener noreferrer"
-              href={`${BASE_URL}/auth/${p}/login?from=${location}&site=${siteId}`}
-              onClick={handleOathClick}
-              className={clsx('oauth-button', styles.button, styles[buttonVariant], styles[p])}
-              data-provider-name={name}
-              title={intl.formatMessage(messages.oauthTitle, { provider: name })}
-            >
-              <img className="oauth-icon" src={icon} width="20" height="20" alt="" aria-hidden={true} />
-            </a>
+            {name === 'Telegram' ? (
```
akellbl4, line 65: Maybe it would be better to treat telegram as any other provider and open a link to the bot immediately.
At the same time, we can change the view in the dropdown for telegram confirmation.

frontend/package.json
```diff
@@ -36,6 +36,7 @@
     "lodash-es": "^4.17.21",
     "node-emoji": "^1.10.0",
     "preact": "^10.5.13",
+    "qrcode.react": "^1.0.1",
```
akellbl4, line 39: We connect lib for QR code into bundle even if it's not enabled and it adds weight to it.
I would suggest finding the smallest lib for that or even better generate code on the backend.

like so:
``

cc @paskal

akellbl4 (changes requested):

frontend/app/components/auth/components/oauth.tsx
Ksinia: What is the reason? We already have redux used in this file prior to my changes.

frontend/app/components/auth/auth.tsx
```diff
@@ -16,19 +17,22 @@ import { OAuth } from './components/oauth';
...
   const intl = useIntl();
   const dispatch = useDispatch();
+  const telegramParams = useSelector((s: StoreState) => s.telegramParams);
   const [oauthProviders, formProviders] = getProviders();
 
   // UI State
   const [isLoading, setLoading] = useState(false);
   const [view, setView] = useState<typeof formProviders[number] | 'token'>(formProviders[0]);
-  const [ref, isDropdownShown, toggleDropdownState] = useDropdown(view === 'token');
+  const [isTelegramShown, toggleTelegram] = useState(false);
```
akellbl4, line 34: This state seems unnecessary. We can `view` as an indicator of the current view. Just add 'telegram' option to it.

elmaester:
image
There's a typo and the checks are failing. Or are there other more difficult to fix reasons? I would really like this feature to work as soon as possible ☺️ Let me know if I can help

frontend/app/components/auth/components/oauth.tsx
akellbl4: The reason is we shouldn't put local state into global store. It's component specific logic and it doesn't affect application. Another reason is to keep it simple and don't write more code than we need. Now logic is scattered between store and this component.

For `user` we put it to the global store and then use in other components that placed in other places.

elmaester:
@Ksinia can I pray to you so that you may please finish this feature? ☺️ Where is the altar? 😁 I want it to work so badly 😄

umputun:
> @Ksinia can I pray to you so that you may please finish this feature? ☺️ Where is the altar? 😁 I want it to work so badly 😄

I wonder, what is your use case and why is one more auth method that important to you? As you probably know, there are many other auth methods currently, and to me TG auth seems like an exotic way to authenticate into remark42.

elmaester:
Telegram is popular among my target audience, and personally I feel more comfortable giving my or users' data to Telegram as opposed to Google. GitHub is not used by non-developers. Twitter is silly with its account policies. Facebook I don't use, so can't create the integration. And it's one of the worst privacy offenders anyway. So Telegram is the best option really. And I have Google and GitHub active already.

umputun:
>  I feel more comfortable giving my or users' data to Telegram as opposed to Google. 

I'm not really sure what data you giving to google/fb/github in this particular case. Auth works in the opposite way and does not communicate with auth providers outside of the basic oauth(2) flow. In fact, this is google giving users' data (very basic data) to remark42. I mean, remark42 expects users with existing accounts on one of the popular places, and all those places know about the user is known regardless of remark42 usage. 

_btw, if such "anti-social login" is important, you may consider email auth_

elmaester:
> btw, if such "anti-social login" is important, you may consider email auth

yeah, but configuring an email server is kinda hard. it's on my to-do list, but feel like a big deal :)

> giving my or users' data

I guess I misspoke. more accurately I should say I'm reluctant to impose a requirement on my users to have an account with such a service. I appreciate the freedom not to. and also it's question of which services I depend on (with my personal account they control) to keep my site functioning, comment attribution consistent, etc. I have more trust in the sanity of Telegram than alternatives. and besides there's extra utility: users will be able to subscribe to comment threads with THEIR Telegram accounts, right?

Also since we're having a conversation, I was wondering if there's any admin dashboard in Remark42 which I somehow haven't discovered yet? Can all comments be exported as JSON? how?

umputun:
> I was wondering if there's any admin dashboard in Remark42 which I somehow haven't discovered yet?

see https://github.com/umputun/remark42/wiki/Admin-UI

> Can all comments be exported as JSON? how?

see API docs: https://remark42.com/docs/contributing/api/

frontend/app/components/auth/auth.tsx
Ksinia: Fixed in 589659d

frontend/package.json
Ksinia: Fixed in e6b5355

akellbl4 (approved):

umputun (approved):
Thx a lot!

elmaester:
Cool! Thanks a lot! When do those changes make it to the docker image? They haven't yet, as far as I can tell.

elmaester:
Ok, then please advise. I did rocker-compose pull, and :master image is specified in the .yml, but for me the old image has then loaded with docker-compose up. What should I do differently?

umputun:
well, something is not right with your compose I guess. I just did this:

```
docker pull  umputun/remark42:master
docker run -it --rm umputun/remark42:master /srv/remark42 server -h

....
remark42 master-6fd7306-20220209T11:07:04
```

as you can see this is a very fresh version built on 2022/02/09

elmaester:
Okay, yeah, I got the same output, running latest version then. But here's what the UI looks like:

image


Is it supposed to look like that? And it doesn't work either.
In my yml I've set `AUTH_TELEGRAM=true`
Is a different configuration necessary?

umputun:
> Is a different configuration necessary?

check the docks please, [here](https://remark42.com/docs/configuration/authorization/), see telegram section

elmaester:
I have that. I'm receiving notifications to my Telegram channel, that part works. Auth does not

umputun:
yeah, the same here. Just added a bot and allowed tg auth on https://remark42.com/demo/ and nothing good happened

@Ksinia @akellbl4 @paskal  - something is not right I think

paskal:
The current master image has build info `master-6fd7306-20220209T11:07:23`, which is one commit behind the master. We'll fix the build and Telegram auth will become available. ETA 12h.

umputun:
I think I spoke too soon. I was able to authenticate once, but after logout in all my attempts to auth I'm getting mysterious "error:0
frw5x-202202-10113850-qiuzk
"

umputun:
actually, it does work, but telegram ui is confusing as hell. @paskal - I think we need to add to docs "press `Start` button in telegram" 

Because I pressed it before I have missed the button the second time

elmaester:
Another issue:
doesn't fit
The popup overflows below the bottom of the webpage, and anything below is not reachable for viewing or clicking.
Maybe depends on page width. Or it appeared after I did
```
document.getElementsByClassName("auth-submit")[0].click()
```
image
Either way, clicking on `Check` got me the `error.0` an nothing else.

umputun:
have you followed the instruction and clicked the link above first?

elmaester:
Ok, yeah, I was missing the step in the tg client. Authenticated now. There was no notification about any necessary interaction from the tg client if it is minimized. A user would never know. I wonder if anything can be done to make the process more intuitive.
Also, the overflow is a serious issue. One must not assume that the page will have sufficient extra space below for the entire popup to fit. In my case it doesn't, until I manage to click the button with a console command. There was no other way to click the button as it was outside the view and not reachable. Seems like a styling issue.
And another styling thing:
image
When the window is narrow, this place could use a margin.

umputun:
> And another styling thing:

let's keep this PR discussion about tg auth issues

elmaester:
Come to think of it, the `error.0` text could be made more descriptive too.

paskal:
image
Now both issues (with hidden window and error.0) are fixed, here is what the Check button says if you haven't clicked Start in Telegram after clicking the link or scanning the QR.

PR #1144, opened 2021-10-09
patreon auth

### LINKED ISSUE #1125: Add support of Patreon auth
With https://github.com/go-pkgz/auth/pull/95 in place, we can add Patreon as auth provider. This provider also sets a new claim allowing us to check if the user paid a subscriber and based on this info, we can do interesting things. Probably we could add optional flags, like `subscribers-only`.

Another thing we may consider is `highlight-subscribers`. This one is mostlyty for UI to make such users visually different, for example, adding some colorful frame around the avatar. We need to extend the user's info and add this flag to the config response to make it happen.

romanilchyshyn:
Partially closes #1125 

Backend:
- Env vars, docs
-  `store.User` new field `PaidSub`
- `subscribers-only` env var & appropriate middleware

Frontend:
- Add new auth type
- Icon asset from [Patreon branding page](https://www.patreon.com/brand)

Need to implement on frontend:
- Subscriber highlighting
- Disable `Send` button for non-subs

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Remark42 is a self-hosted, lightweight and simple (yet functional) comment engine, which doesn't spy on users. It can be embedded into blogs, articles, or any other place where readers add comments.
 
-* Social login via Google, Twitter, Facebook, Microsoft, GitHub, Yandex and Telegram
+* Social login via Google, Twitter, Facebook, Microsoft, GitHub, Yandex, Patreon and Telegram
 * Login via email
 * Optional anonymous access
 * Multi-level nested comments with both tree and plain presentations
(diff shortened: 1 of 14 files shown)


romanilchyshyn:
Looks like `TestMain_WithWebhook` is flaky.

akellbl4:
> Subscriber highlighting
Disable Send button for non-subs.

1. We can add the Patreon logo right next to the username for subscribers or add them verified checkmark automatically (maybe we can color verified checkmark into the Patreon brand color)
2. That's  should be easy

umputun:
> We can add the Patreon logo right next to the username for subscribers or add them verified checkmark automatically (maybe we can color verified checkmark into the Patreon brand color)

Adding Patreon's logo right next to the username for subscribers sounds good

umputun:
@paskal - can you pls take a look? the PR looks fine to me, but another pair of eyes will be nice to have

PR #1160, opened 2021-10-29
Enable telegram notify trough writing bot a message

paskal:
Previously it was done through writing bot first, clicking a button, copying the token, and pasting it into the web interface.

The new flow is way simpler: click the link to write bot a message, then click the "Check" button in the web UI and you got notifications enabled.

Unsolved problems:
- [ ] auth is not listening to the same Telegram updates

### DIFF
--- a/backend/app/notify/telegram.go
+++ b/backend/app/notify/telegram.go
@@ -7,8 +7,11 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	neturl "net/url"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/go-multierror"
@@ -21,10 +24,11 @@ import (
 
 // TelegramParams contain settings for telegram notifications
 type TelegramParams struct {
-	AdminChannelID    string        // unique identifier for the target chat or username of the target channel (in the format @channelusername)
-	Token             string        // token for telegram bot API interactions
-	Timeout           time.Duration // http client timeout
-	UserNotifications bool          // flag which enables user notifications
+	AdminChannelID       string        // unique identifier for the target chat or username of the target channel (in the format @channelusername)
+	Token                string        // token for telegram bot API interactions
+	Timeout              time.Duration // http client timeout
+	UserNotifications    bool          // flag which enables user notifications
+	ErrorMsg, SuccessMsg string        // messages for successful and unsuccessful subscription requests to bot
 
 	apiPrefix string // changed only in tests
 }
@@ -33,13 +37,32 @@ type TelegramParams struct {
 type Telegram struct {
 	TelegramParams
 
-	username string // bot username
+	// Identifier of the first update to be requested.
+	// Should be equal to LastSeenUpdateID + 1
+	// See https://core.telegram.org/bots/api#getupdates
+	updateOffset           int
+	apiPollInterval        time.Duration // interval to check updates from Telegram API and answer to users
+	expiredCleanupInterval time.Duration // interval to check and clean up expired notification requests
+	username               string        // bot username
+	run                    int32         // non-zero if Run goroutine has started
+	requests               struct {
+		sync.RWMutex
+		data map[string]tgAuthRequest
+	}
 }
 
 // telegramMsg is used to send message trough Telegram bot API
 type telegramMsg struct {
 	Text      string `json:"text"`
 	ParseMode string `json:"parse_mode,omitempty"`
+}
+
+type tgAuthRequest struct {
+	confirmed  bool // whether login request has been confirmed and user info set
+	expires    time.Time
+	telegramID string
+	user       string
+	site       string
 }
 
 // TelegramBotInfo structure contains information about telegram bot, which is used from whole telegram API response
@@ -49,6 +72,8 @@ type TelegramBotInfo struct {
 
 const telegramTimeOut = 5000 * time.Millisecond
 const telegramAPIPrefix = "https://api.telegram.org/bot"
+const tgPollInterval = time.Second * 5
+const tgCleanupInterval = time.Minute * 5
 
 // NewTelegram makes telegram bot for notifications
 func NewTelegram(params TelegramParams) (*Telegram, error) {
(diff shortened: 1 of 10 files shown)


backend/app/notify/telegram.go
```diff
@@ -159,44 +187,129 @@ func escapeTelegramText(text string) string {
...
 	}
-	return b, nil
 }
 
-// GetBotUsername returns bot username
-func (t *Telegram) GetBotUsername() string {
-	return t.username
+// ProcessUpdate is alternative to Run, it processes provided plain text update from Telegram
+// so that caller could get updates and send it not only there but to multiple sources
+func (t *Telegram) ProcessUpdate(ctx context.Context, textUpdate string) error {
```
umputun, line 294: I think this t.run=true may cause some races, as Run probably will be run in a separate goroutine. At the same time public ProcessUpdate checks this flag to avoid mutually exclusive runs
maybe make it atomic?

backend/app/rest/api/rest_private.go
```diff
@@ -39,11 +41,19 @@ type private struct {
...
 	notifyService    *notify.Service
 	authenticator    *auth.Service
+	telegramService  TelegramService
 	remarkURL        string
 	anonVote         bool
 	templates        templates.FileReader
 }
 
+// TelegramService is a subset of Telegram service used for setting up user telegram notifications
+type TelegramService interface {
```
umputun, line 51: is there a reason for the public interface here? this thing constructed privately anyway

backend/app/rest/api/rest_private.go
```diff
@@ -349,55 +359,64 @@ func (s *private) sendEmailConfirmationCtrl(w http.ResponseWriter, r *http.Reque
...
+// telegramSubscribeCtrl generates and verifies telegram notification request
+// GET /telegram/subscribe?site=siteID<&tkn=token>
+func (s *private) telegramSubscribeCtrl(w http.ResponseWriter, r *http.Request) {
 	user := rest.MustGetUserInfo(r)
-	address := r.URL.Query().Get("address")
-	siteID := r.URL.Query().Get("site")
-	if address == "" {
-		rest.SendErrorJSON(w, r, http.StatusBadRequest,
-			errors.New("missing parameter"), "address parameter is required", rest.ErrInternal)
+	queryToken := r.URL.Query().Get("tkn")
```
umputun, line 366: probably we can move queryToken init closer the the if using it

backend/app/rest/telegram.go
```diff
@@ -0,0 +1,63 @@
+package rest
```
umputun, line 1: to me this file doesn't seem to be a logical part of rest package as it has very little to do with rest

umputun (changes requested):
a few things after quick review

backend/app/rest/telegram.go
paskal: Agreed, however, I have no better idea where to put it: it should not be part of notify package as it will be cut into a separate package. Suggestions are welcome.

backend/app/providers/telegram.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/umputun/remark42/backend/app/notify"
+)
+
+type tgRequester interface {
+	Request(ctx context.Context, method string, b []byte, data interface{}) error
+}
+
+// TGUpdatesReceiver used to dispatch telegram updates to multiple receivers
+type TGUpdatesReceiver interface {
+	ProcessUpdate(ctx context.Context, textUpdate string) error
```
umputun, line 22: what do you think about adding Name or Id or something identifying the receiver? this way we could report more accurately "failure from blah" instead of not very helpful "failure from destination" error

umputun: or maybe Name() method, unless you have planned to provide a functional adopter to this interface

backend/app/providers/telegram.go
```diff
@@ -0,0 +1,65 @@
+package providers
+
+// This provider provides common logic, reused by Telegram auth and notifications verification
```
umputun, line 3: From what you explained me earlier, it seems what the real reason why this Dispatch thing even needed is the nature of tg API requiring a common/single event loop, and we need it for both auth and notify. This is not that clear from the comment about "common logic".

backend/app/providers/telegram.go
```diff
@@ -0,0 +1,65 @@
...
+
+	processUpdatedTicker := time.NewTicker(period)
+	for {
+		select {
+		case <-ctx.Done():
+			processUpdatedTicker.Stop()
+			return
+		case <-processUpdatedTicker.C:
+			url := `getUpdates?allowed_updates=["message"]`
+			if updateOffset != 0 {
```
umputun, line 38: the offset thing is not super clear to me. I guess the offset 0 is relative to the start time (i.e. not the absolute 0), but it will be nice to describe this for someone not too familiar with tg api

backend/app/providers/telegram.go
```diff
@@ -0,0 +1,65 @@
...
+				url += fmt.Sprintf("&offset=%d", updateOffset)
+			}
+
+			var result notify.TelegramUpdate
```
umputun, line 42: result.Result latter seems odd, maybe rename it?

backend/app/providers/telegram.go
```diff
@@ -0,0 +1,65 @@
...
+
+			err := requester.Request(ctx, url, nil, &result)
+			if err != nil {
+				log.Printf("[ERROR] failed to fetch updates: %v", err)
```
umputun, line 46: I think this should be WARN and not ERROR, as ERROR usually means - the usual flow altered. Here you noe even doing continue (maybe missed?)

backend/app/providers/telegram.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			for _, u := range result.Result {
```
umputun, line 49: from the first sight I don't understand what this loop is doing. I see it iterates over some results and update offset. The part what is not clear - why some results here with UpdateID below the updateOffset you explicitly set above? And the second question - you seems to update offset in some cases but not altering result? Not sure what it means as I don't really understand all this loop in the first place. Maybe some little comment may clarify the mystery to me ;)

umputun (changes requested):
some questions around telegram provider implementation

umputun (approved):
LGTM

PR #1162, opened 2021-10-29
Improve telegram notifications

paskal:
These changes are designed to ease the transition into the simplified telegram notifications verification model, proposed in https://github.com/umputun/remark42/pull/1160.


backend/app/cmd/server.go
```diff
@@ -473,22 +473,24 @@ func (s *ServerCommand) newServerApp(ctx context.Context) (*serverApp, error) {
...
-	}
-
-	// we pass telegramBotUsername to Rest server only if user notifications are enabled
-	if !contains("telegram", s.Notify.Users) {
-		telegramBotUsername = ""
-	}
+	var telegramBotUsername string
+	notifyService, telegramService, err := s.makeNotify(dataService, authenticator)
 
 	if err != nil {
```
umputun, line 479: even one level of if/else makes me uncomfortable but here we have nested if/else. I'd suggest to simplify it and get rid of else with an anonymous or real function with early exit(s)

backend/app/cmd/server.go
```diff
@@ -912,11 +914,11 @@ func (s *ServerCommand) loadEmailTemplate() (string, error) {
 	return string(file), nil
 }
 
-// aside from notify.Service and error, returns telegram bot name which will be passed to the frontend
-func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *auth.Service) (*notify.Service, string, error) {
+// aside from notify.Service and error, returns telegram service
+func (s *ServerCommand) makeNotify(dataStore *service.DataStore, authenticator *auth.Service) (*notify.Service, *notify.Telegram, error) {
```
umputun, line 918: if tg needs a separate return value, maybe make it a separate func? like makeTgNotify?

backend/app/rest/api/rest_private.go
```diff
@@ -44,6 +44,11 @@ type private struct {
 	templates        templates.FileReader
 }
 
+// TelegramService is a subset of Telegram service used for setting up user telegram notifications
+type TelegramService interface {
```
umputun, line 48: is it even used anywhere?

umputun (changes requested):

umputun (approved):
LGTM

PR #1165, opened 2021-11-07
Move more parts of Readme to the docs site

paskal:


umputun (approved):

umputun:
looks like we got a conflict here

paskal:
Should be good to go now.

PR #1171, opened 2021-11-07
Use single loop for telegram auth and notify

paskal:
Second part of https://github.com/umputun/remark42/pull/1160, allowing telegram auth and notify to work at the same time.

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -454,7 +454,12 @@ func (s *ServerCommand) newServerApp(ctx context.Context) (*serverApp, error) {
 		return nil, errors.Wrap(err, "failed to make avatar store")
 	}
 	authRefreshCache := newAuthRefreshCache()
-	authenticator, err := s.makeAuthenticator(ctx, dataService, avatarStore, adminStore, authRefreshCache)
+	authenticator := s.getAuthenticator(dataService, avatarStore, adminStore, authRefreshCache)
+
+	telegramAuth := s.makeTelegramAuth(authenticator) // telegram auth requires TelegramAPI listener which is constructed below
+	telegramService, telegramBotUsername := s.startTelegramAuthAndNotify(ctx, telegramAuth)
+
+	err = s.addAuthProviders(authenticator)
 	if err != nil {
 		_ = dataService.Close()
 		return nil, errors.Wrap(err, "failed to make authenticator")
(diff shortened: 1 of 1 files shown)


backend/app/cmd/server.go
```diff
@@ -454,25 +454,14 @@ func (s *ServerCommand) newServerApp(ctx context.Context) (*serverApp, error) {
 		return nil, errors.Wrap(err, "failed to make avatar store")
 	}
 	authRefreshCache := newAuthRefreshCache()
-	authenticator, err := s.makeAuthenticator(ctx, dataService, avatarStore, adminStore, authRefreshCache)
+	authenticator := s.makeAuthenticator(dataService, avatarStore, adminStore, authRefreshCache)
+
+	telegramAuth, err := s.addAuthProviders(authenticator)
```
umputun, line 459: this return value seems very unexpected for the used-to-be generic call. I.e. getting tg-specific TGUpdatesReceiver from addAuthProviders doesn't fill right to me. I would rather split it and will have addTgAuthProvider as a separate func returning tg-related updater

backend/app/cmd/server.go
```diff
@@ -1058,8 +1058,8 @@ func (s *ServerCommand) makeSSLConfig() (config api.SSLConfig, err error) {
 	return config, err
 }
 
-func (s *ServerCommand) makeAuthenticator(ctx context.Context, ds *service.DataStore, avas avatar.Store, admns admin.Store, authRefreshCache *authRefreshCache) (*auth.Service, error) {
-	authenticator := auth.NewService(auth.Opts{
+func (s *ServerCommand) makeAuthenticator(ds *service.DataStore, avas avatar.Store, admns admin.Store, authRefreshCache *authRefreshCache) *auth.Service {
```
umputun, line 1061: this one probably need a comment indicating whay the returned auth service is not really made as it is not including any providers. Maybe it should be called prepAuthenticator instead of make, not sure

umputun (changes requested):

backend/app/cmd/server.go
```diff
@@ -493,8 +483,12 @@ func (s *ServerCommand) newServerApp(ctx context.Context) (*serverApp, error) {
 			telegramBotUsername = telegramService.GetBotUsername()
 		}
 		notifyDestinations = append(notifyDestinations, telegramService)
+		telegramReceivers := []providers.TGUpdatesReceiver{telegramService}
```
umputun, line 486: just a suggestion/idea -  maybe it make sense to extract all/some tg related code to a separate func prepping all the tg related things? i.e both tg auth, tg notify as well as the dispatcher. As of now tg prep all around in this newServerApp and I would like to keep this func as close as possible to a glue code only (call smth, check err and  assign to smth)

umputun (approved):
LGTM

Issue #1085, opened 2021-08-08
show total number of comments in profile


akellbl4:
this endpoint `/api/v1/comments` returns the total number of user's comments
we can show it on the profile screen

_If you'd like to work on this ping me and I'll bring design for it._

esvyridov:
@akellbl4 Hi, I would take this issue

akellbl4:
@esvyridov hey! I checked the UI and found out that we have the title "My recent comments" or "Recent comments".
In my opinion, the best place is right next to the title. It led me to change that title to "My comments" / "Comments".
But in this case, it could be awesome to add the ability to check all user's comments. 

As I can see method `/api/v1/comments?site={site}&user={user_id}` has only limiter with `limit={number}` but doesn't have pagination with `skip={number}`. It means we need to have that ability on the server-side and then add "show more comments" for looking at older comments of a user.

@paskal @umputun could we add `skip` parameter to `/api/v1/comments?site={site}&user={user_id}`

akellbl4:
Design proposal is [here](https://www.figma.com/file/UKUj98QJOGMnqH8Huuv3P8/Remark42-Profile-comments-counter?node-id=0%3A1)

esvyridov:
@akellbl4 looks good. I'll start to work after we get some information about `skip` param.
It looks like `count` param counts deleted comments. I think it can be confusing for a user if I put that number into UI.

![image](https://user-images.githubusercontent.com/46542370/131260172-56278b17-53e0-43af-ad6f-8278fe6aa12a.png)

umputun:
I have added support of `skip` query param to `/api/v1/comments?site=siteID&user=id&limit=123&skip=10` endpoint

umputun:
> It looks like count param counts deleted comments. I think it can be confusing for a user if I put that number into UI.

It is confusing, however, I don't have an effective solution for this problem yet. This should not affect your UI work and as soon as figure how to filter deleted, the same API will return an altered count.

esvyridov:
@umputun sounds good, thank you
@akellbl4 I'll start to work on comments counter and "load more" button very soon

akellbl4:
@esvyridov sounds great!

esvyridov:
@akellbl4 PR is ready https://github.com/umputun/remark42/pull/1176
@umputun can I try to resolve the issue with `count` property in `/comments` response which we discussed above?

umputun:
[...]

PR #1176, opened 2021-11-14
Add Load More button to profile sidebar

esvyridov:
Issue: https://github.com/umputun/remark42/issues/1085

**Profile sidebar with load more button**

Screenshot 2021-11-14 at 19 04 01

### DIFF
--- a/frontend/app/components/profile/profile.tsx
+++ b/frontend/app/components/profile/profile.tsx
@@ -1,6 +1,6 @@
 import clsx from 'clsx';
 import { h, Fragment } from 'preact';
-import { useEffect, useMemo, useRef, useState } from 'preact/hooks';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'preact/hooks';
 import { useIntl, FormattedMessage } from 'react-intl';
 
 import { getUserComments } from 'common/api';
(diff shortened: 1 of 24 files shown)


frontend/app/components/profile/profile.tsx
```diff
@@ -32,21 +32,32 @@ async function signout() {
...
+  const fetchUserComments = useCallback(
+    async (skip: number = 0) => {
+      const { comments, count } = await getUserComments(user.id, { skip, limit: commentsLimit });
+      return { comments, count };
+    },
+    [user.id, commentsLimit]
+  );
 
-  async function fetchUserComments(userId: string) {
+  const fetchUserCommentsOnMount = useCallback(async () => {
```
akellbl4, line 53: It could be moved inside `useEffect` because it is used only there and this function will be called only once.

frontend/app/components/profile/profile.tsx
```diff
@@ -32,21 +32,32 @@ async function signout() {
 export function Profile() {
   const intl = useIntl();
   const rootRef = useRef<HTMLDivElement>(null);
+  const commentsLimit = useMemo(() => 10, []);
```
akellbl4, line 35: looks unnecessary to me. it can be a simple variable or you can move it outside of the component

frontend/app/locales/ru.json
```diff
@@ -159,6 +159,7 @@
   "toolbar.unordered-list": "Добавить маркированный список",
   "user.my-comments": "Мои комментарии",
   "user.comments": "Комментарии",
+  "user.load-more": "Load more",
```
akellbl4, line 162:
```suggestion
  "user.load-more": "Загрузить ещё",
```

akellbl4 (changes requested):
Overall code looks good. Just a few comments.

frontend/app/components/profile/profile.tsx
esvyridov: yeah, it looks strange for me too now :D fixed

frontend/app/components/profile/profile.tsx
esvyridov: @akellbl4  I created this function to underline separation between `fetchUserCommentsOnMount` and `fetchMoreUserComments`. In opinion it increases readability of the component, but I can be wrong :D Anyway, please let me know if you still prefer to put the body of the function to `useEffect`, I'll do it.

akellbl4: Hm. I checked the code a bit closer.

`fetchUserCommentsOnMount` and `fetchMoreUserComments` are nearly identical.
The only difference in this `setComments([...(comments || []), ...nextComments]);`. As I understand we can use only one function for both cases and we wont need to have `fetchUserComments`.

akellbl4:
@esvyridov hey, I'm sorry for invading into the PR. I proposed some changes and would like to see your opinion about them ;)

akellbl4 (approved):

esvyridov:
@akellbl4 sure, no problem :) The code looks good, how about adding a separate "fetching more" flag to avoid the case below?

image

Just to remind: there is a small issue on the backend side https://github.com/umputun/remark42/issues/1085#issuecomment-907838347 related to the counter and load more functionality. I volunteered to fix it, but it's still in my todo list D:

akellbl4:
@esvyridov oh, valid idea. but I think it could be easier if just check `comments !== null && isLoading` ;)

esvyridov:
@akellbl4 done, but now we have two elements (Preloader and Spinner) with the same aria-labels. I would add `data-testid` to them, sounds good to you?

esvyridov:
@akellbl4 done

akellbl4 (approved):

Issue #1169, opened 2021-11-07
[site] favicon.ico is missing


paskal:
We're missing https://remark42.com/favicon.ico, and ideally should generate multiple sizes of it with a tool like https://realfavicongenerator.net.

PR #1177, opened 2021-11-16
Add favicon to site

adfslslddkds:
Issue https://github.com/umputun/remark42/issues/1169 add favicon. Favicon generated using https://favicon.io/


umputun:
we don't need a new design for favicon, as we already have one. All we need is to actually make all the sizes from it and put to the right places

umputun:
remark-1024

akellbl4 (approved):

PR #1184, opened 2021-11-23
Update golangci-lint to 1.43.0, fix discovered issues

paskal:


backend/.golangci.yml
```diff
@@ -74,6 +72,3 @@ issues:
         - gosec
         - dupl
   exclude-use-default: false
-
-service:
-  golangci-lint-version: 1.41.x
```
umputun, line 79: this will make result of linter dependd on the installed version. Not a problem for CI run, but local runs can produce different warnings depends on the local version installed

paskal: I thought it's only used for web service, as it's not documented in golangci-lint documentation and doesn't produce any difference in 1.43.0 output (when I try 1.41.x and 1.43.x values).

umputun: ah, ok. this sucks but if this is the case - fine with me

umputun (approved):
LGTM

PR #1196, opened 2021-12-19
Do not issue deprecation warning on notify.type by default

### LINKED ISSUE #1191: running container without any param shows the depreciation warn for --notify.type
@paskal  - I don't think this is expected

```
docker run -it --rm -v $(pwd)/var:/srv/var umputun/remark42:master /srv/remark42 server --secret=12345 --url=http://example.com

....
2021/12/12 19:21:27.841 [WARN]  --notify.type is deprecated since v1.9 and will be removed in the future, please use --notify.(users|admins) instead
```

As you can see I have not used `--notify.type`, probably sneaked in via default

paskal:
Default configuration for notify.type is "none", and prior to this change, it was issuing the deprecation warning which was not intended behavior.

Fixes #1191.

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -354,7 +354,9 @@ func (s *ServerCommand) HandleDeprecatedFlags() (result []DeprecatedFlag) {
 		s.ImageProxy.HTTP2HTTPS = s.LegacyImageProxy
 		result = append(result, DeprecatedFlag{Old: "img-proxy", New: "image-proxy.http2https", Version: "1.5"})
 	}
-	if len(s.Notify.Type) != 0 && (len(s.Notify.Users) != 0 || len(s.Notify.Admins) != 0) {
+	if len(s.Notify.Type) != 0 && // explicitly empty, would likely never happen due to "none" default
+		!(len(s.Notify.Type) == 1 && contains("none", s.Notify.Type)) && // ignore default, "none" notify type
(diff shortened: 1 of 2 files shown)


backend/app/cmd/server.go
```diff
@@ -354,7 +354,7 @@ func (s *ServerCommand) HandleDeprecatedFlags() (result []DeprecatedFlag) {
 		s.ImageProxy.HTTP2HTTPS = s.LegacyImageProxy
 		result = append(result, DeprecatedFlag{Old: "img-proxy", New: "image-proxy.http2https", Version: "1.5"})
 	}
-	if len(s.Notify.Type) != 0 && (len(s.Notify.Users) != 0 || len(s.Notify.Admins) != 0) {
+	if len(s.Notify.Type) != 0 && !(len(s.Notify.Type) == 1 && contains("none", s.Notify.Type)) && (len(s.Notify.Users) != 0 || len(s.Notify.Admins) != 0) {
```
umputun, line 357: this is confusing. I would guess this `!(len(s.Notify.Type) == 1` means `len(s.Notify.Type) != 1` and if so, both first lens can be `len() > 1` ? I would be also nice to add a comment here, because just from this part it is unclear what is so special about a single element s.Notify.Type slice.

paskal: I've added comments about each part of the expression, please take a look.

umputun:
well, this condition is still not too clear and I don't think we have a test for this. but this is, either way, a minor thingy, so I'll merge. If you can add a test for those conditions in meantime it would be nice

PR #1212, opened 2022-01-01
Telegram notifications format headers, resolve #1202

### LINKED ISSUE #1202: Telegram notifications: format headers as bold
I post the following comment:

```
# h1 Heading 8-)
## h2 Heading
### h3 Heading
#### h4 Heading
##### h5 Heading
###### h6 Heading

ordinary text
```

Here is what I see as a rendered comment:

image

And here is what I see as a telegram notification:

image

HTML tags [allowed](https://core.telegram.org/bots/api#html-style) in Telegram message formatting are limited and h1-h6 are not in the list. I propose to replace h1-h3 with `<b>`, and h4-h6 with `<I><b>` so headers would be distinguishable when received in Telegram notification.

stels-cs:


backend/app/notify/telegram.go
```diff
@@ -171,6 +171,19 @@ func buildMessage(req Request) ([]byte, error) {
 // returns HTML with only tags allowed in Telegram HTML message payload, also trims ending newlines
 // https://core.telegram.org/bots/api#html-style
 func telegramSupportedHTML(htmlText string) string {
+	// telegram not allow h1-h6 tags
```
paskal, line 174: Thanks for your contribution! That way of working with HTML would not work in the wild but will work here well as we process HTML which is produced by us previously, so it doesn't have anything unusual like extra spaces (`<h1 >`) or attributes (`<h1 class="red">`).

Would you be so kind as to move that code to a separate function, like `adjustHtmlTags`? Also, I would propose renaming `htmlText` to something like `adjustedHtmlText` so it will be obvious to the reader that the return statement operates with another string that was passed into that function.

backend/app/notify/telegram_test.go
```diff
@@ -496,3 +496,13 @@ func mockTelegramServer(h http.HandlerFunc) *httptest.Server {
 
 	return httptest.NewServer(router)
 }
+
+func TestTelegram_telegramSupportedHTML(t *testing.T) {
```
paskal, line 500: I propose to test it next to `// test buildMessage separately for message text`, that way it would be less painful to change the implementation later on but verify that the logic stays the same.

Ideally, we need to test these things by calling `Telegram.Send` and don't care about how it gets generated internally, but we don't have a mock server setup in the tests yet to do it.  There are fewer steps between `buildMessage` and `Send` than between the new function and it, so testing `buildMessage` will simplify including this test case for whoever will rewrite tests later on.

paskal (changes requested):
Thanks a lot for your work! Everything is good, I have only two small comments on code location.

backend/app/notify/telegram.go
```diff
@@ -188,6 +190,43 @@ func escapeTelegramText(text string) string {
 	return text
 }
 
+// telegram not allow h1-h6 tags
+// replace these tags with a combination of <b> and <i> for visual distinction
+// https://github.com/umputun/remark42/issues/1202
```
paskal, line 195: Just a little note, I think that particular link shouldn't be there: we don't have them elsewhere in the code, and it would be easy to deduct from the git annotation. If understanding a function requires you to go through the link, we are likely doing something wrong, and here nothing will break if we omit this link.

backend/app/notify/telegram_test.go
```diff
@@ -166,6 +166,26 @@ func TestTelegram_Send(t *testing.T) {
 		`\"\u003ci\u003esome parent text with a \u003ca href=\"http://example.org\"\u003elink\u003c/a\u003e and special text:\u0026amp; \u0026lt; \u0026gt; \u0026amp;\u003c/i\u003e\"\n\n`+
 		`↦  \u003ca href=\"http://example.org/\"\u003e[test title]\u003c/a\u003e","parse_mode":"HTML"}`,
 		string(res))
+
+	// special case for text with h1-h3 header https://github.com/umputun/remark42/issues/1202
+	c.Text = "<h1>Hello</h1>"
+	res, err = buildMessage(Request{Comment: c, parent: cp})
```
paskal, line 172: I propose to omit parent comment, so the text would be relevant only to what we are testing, and also, you can combine both cases into a single test.

backend/app/notify/telegram_test.go
```diff
@@ -166,6 +166,26 @@ func TestTelegram_Send(t *testing.T) {
 		`\"\u003ci\u003esome parent text with a \u003ca href=\"http://example.org\"\u003elink\u003c/a\u003e and special text:\u0026amp; \u0026lt; \u0026gt; \u0026amp;\u003c/i\u003e\"\n\n`+
 		`↦  \u003ca href=\"http://example.org/\"\u003e[test title]\u003c/a\u003e","parse_mode":"HTML"}`,
 		string(res))
+
+	// special case for text with h1-h3 header https://github.com/umputun/remark42/issues/1202
```
paskal, line 170: I propose to omit the link here as well and leave only text comments.

paskal (changes requested):

backend/app/notify/telegram.go
paskal: If you're wondering, we use links in code for external API reference and source attribution when the code is copied from somewhere.

paskal (approved):

umputun:
just a couple of lint warnings:

```
Warning: var-naming: var adjustedHtmlText should be adjustedHTMLText (revive)
Warning: var-naming: func adjustHtmlTags should be adjustHTMLTags (revive)
```

backend/test.go
```diff
@@ -0,0 +1,51 @@
+package main
```
paskal, line 1: I assume this was not supposed to be committed?

stels-cs: oops, yes, commit by mistake )

paskal (approved):

umputun (approved):
thx!

umputun:
if you could squash it to a single comment it would be nice. If I do it it will record the commit as mine

stels-cs:
hope i did squash right

paskal (approved):

PR #1235, opened 2022-01-19
Protect anonymous comments from being changed by other users

paleale:
[Direct](https://github.com/go-pkgz/auth/blob/fa21d23/provider/direct.go#L99-L102) auth from go-pkgz/auth [used](https://github.com/umputun/remark42/blob/90e5373/backend/app/cmd/server.go#L834-L861) in Remark42 for anonymous authentication. The problem is that as soon as a user logs in with the same name as another anonymous user, they get precisely the same user ID and can delete or alter messages by that other user.
To prevent that, we need to provide custom [UserIDFunc](https://github.com/go-pkgz/auth/blob/fa21d23ff361c2728f10807ba3dda386ef0c541b/provider/direct.go#L41), which I propose should return hashed IP of the user so that there will be no way another anonymous user and I get the same ID at the same time, but it would be possible to re-login right away with the same ID if I want to adjust my anonymous comment.

Issue description has been originally provided by @paskal

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -911,6 +911,8 @@ func (s *ServerCommand) addAuthProviders(authenticator *auth.Service) error {
 			}
 			return true, nil
 		}),
+			// Custom user ID generator, used to distinguish anonymous users with the same login
+			// coming from different IPs
 			func(user string, r *http.Request) string {
 				return user + r.RemoteAddr
 			})
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -887,7 +887,8 @@ func (s *ServerCommand) addAuthProviders(authenticator *auth.Service) error {
 	if s.Auth.Anonymous {
 		log.Print("[INFO] anonymous access enabled")
 		var isValidAnonName = regexp.MustCompile(`^[\p{L}\d_ ]+$`).MatchString
-		authenticator.AddDirectProvider("anonymous", provider.CredCheckerFunc(func(user, _ string) (ok bool, err error) {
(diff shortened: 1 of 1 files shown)


umputun (changes requested):
Thx for the pr!

I think it makes sense to extend user id returned by AddDirectProviderWithUserIDFunc with X-Forwarded-For and X-Real-IP. Pls note - we likely should also handle chains of reported ips from those headers as each proxy level may append its ip. 

Probably smth similar to [this](https://github.com/go-chi/chi/blob/master/middleware/realip.go)

backend/app/cmd/server.go
```diff
@@ -859,7 +859,10 @@ func (s *ServerCommand) addAuthProviders(authenticator *auth.Service) error {
 				return false, nil
 			}
 			return true, nil
-		}))
+		}),
+			func(user string, r *http.Request) string {
```
paskal, line 863: I propose to add a comment here, something like "Custom user ID generator, used to distinguish anonymous users coming from different IPs".

umputun:
@paleale pls let me know if you are planning to make the proposed change

paleale:
@umputun I am, back again, still uncertain about all the possible force majors, but I'd like to finish that at nearest holidays.

umputun:
My comment was probably inaccurate, or i have added it before we had `RealIP` middleware added. With this middleware in place your `user + r.RemoteAddr` should be fine. Pls, make sure the conflict resolved and I'll merge

paskal (approved):

umputun (approved):
LGTM, thx

PR #1243, opened 2022-01-30
Log when deprecated and new args are set at the same time

### LINKED ISSUE #1218: Notify admin not work: deprecated params are not used when new ones are set
I am using version 1.8.1, on November 24, I still received notifications in telegrams, but apparently after some update they stopped. That is, now on version 1.8.1 I do not receive telegram notifications. I also tried to set up email notifications and they don't come either.

Without changing anything in the configuration settings, I just switched to the master version and email notifications worked, but telegram notifications still do not work

paskal:
For example, when notify.telegram.token and telegram.token are both set but to different values, user might see "access denied" error in log on attempt to send telegram notification, thinking that notify.telegram.token value is used, when in fact it is ignored and only telegram.token is used.

New behavior is the same, ignoring the old param when new one is set, but issuing the error log message which explicitly tells the user about that.

Resolves #1218.

### DIFF
--- a/backend/app/cmd/cmd.go
+++ b/backend/app/cmd/cmd.go
@@ -22,6 +22,7 @@ type CommonOptionsCommander interface {
 	SetCommon(commonOpts CommonOpts)
 	Execute(args []string) error
 	HandleDeprecatedFlags() []DeprecatedFlag
+	FindDeprecatedFlagsCollisions() []DeprecatedFlag
 }
 
 // CommonOpts sets externally from main, shared across all commands
@@ -48,6 +49,10 @@ func (c *CommonOpts) SetCommon(commonOpts CommonOpts) {
 
 // HandleDeprecatedFlags sets new flags from deprecated and returns their list
 func (c *CommonOpts) HandleDeprecatedFlags() []DeprecatedFlag { return nil }
+
+// FindDeprecatedFlagsCollisions returns list of flags collisions, e.g. when both deprecated
+// and new flags are set to non-default values, and different from each other.
+func (c *CommonOpts) FindDeprecatedFlagsCollisions() []DeprecatedFlag { return nil }
 
 // fileParser used to convert template strings like blah-{{.SITE}}-{{.YYYYMMDD}} the final format
 type fileParser struct {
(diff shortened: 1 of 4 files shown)


backend/app/cmd/cmd.go
```diff
@@ -22,6 +22,7 @@ type CommonOptionsCommander interface {
 	SetCommon(commonOpts CommonOpts)
 	Execute(args []string) error
 	HandleDeprecatedFlags() []DeprecatedFlag
+	FindDeprecatedFlagsCollisions() []DeprecatedFlag
```
umputun, line 25: unlike others, this one has no default implementation in CommonOpts. I guess it may break other commands

backend/app/cmd/server.go
```diff
@@ -321,7 +321,9 @@ func (s *ServerCommand) Execute(_ []string) error {
 	return nil
 }
 
-// HandleDeprecatedFlags sets new flags from deprecated returns their list
+// HandleDeprecatedFlags sets new flags from deprecated returns their list.
+// Returned list has DeprecatedFlag.Old and DeprecatedFlag.Version set, and DeprecatedFlag.New is optional
+// (as some entries are removed without substitute).
 func (s *ServerCommand) HandleDeprecatedFlags() (result []DeprecatedFlag) {
```
umputun, line 327: maybe we don't need two methods to handle deprecated flags and should combine them?

umputun (changes requested):

backend/app/cmd/server.go
paskal: I thought of combining them and differentiating the results by flag, but found out that the complexity of resulting function is too high and there really should be two functions: one reading old flags and setting new ones if old ones are still in use, and other just detecting collisions but sets nothing.

umputun: well, i have nothing against having two functions inside the command to address the complexity, it makes sense. However I don't really see the reason the have two "entry points", i.e. interface functions doing a very similar thing. Maybe keep two and call conflict thingy from the handler without extending the interface?

umputun (approved):
LGTM

PR #1246, opened 2022-01-31
Add QR API endpoint for Telegram auth and notifications

paskal:
Telegram authentication requires you to open a chat on the phone. It's convenient to have a QR code for the case when you want to log in on the computer but have Telegram only on your phone and would be able to scan the QR instead of copy-pasting the link from the computer to the phone any other way. Generating QR codes restricted to strings starting with `https://t.me/`

Pre-requisite for #830 and #707 as in #1235 we discovered that it was discovered that generating QR on the frontend is rather expensive in terms of transferred data and increasing amount of JS code in the project, so we would just generate the QR and transfer on the backend. Server-side rendering is the future.


backend/app/rest/api/rest_public.go
```diff
@@ -364,6 +365,31 @@ func (s *public) robotsCtrl(w http.ResponseWriter, r *http.Request) {
 	render.PlainText(w, r, "User-agent: *\nDisallow: /auth/\nDisallow: /api/\n"+strings.Join(allowed, "\n")+"\n")
 }
 
+// GET /qr - generates QR for URL in provided body, used for generating telegram link QR. Provided text should start with https://t.me/.
+func (s *public) qrCtrl(w http.ResponseWriter, r *http.Request) {
+	text := r.URL.Query().Get("text")
```
umputun, line 370: Maybe we make sure this method's body is match the the method's name? I mean it called generically `qrCtrl` but in fact this is `qrTelegramCtrl` 

Another thing to consider - maybe we don't really need such a specific method and rather  keep qrCtrl to handle any urls? And maybe renaming "text" query param to "url" will make the api better match to what it does?

backend/app/rest/api/rest_public_test.go
```diff
@@ -615,6 +615,36 @@ func TestRest_Config(t *testing.T) {
...
+	assert.Equal(t, 500, code)
+	assert.Equal(t, "{\"code\":0,\"details\":\"can't generate QR\",\"error\":\"content too long to encode\"}\n", body)
+
+	// wrong request
+	body, code = get(t, ts.URL+"/api/v1/qr/telegram?url=nonsense")
+	assert.Equal(t, 400, code)
+	assert.Equal(t, "{\"code\":0,\"details\":\"text parameter should start with https://t.me/\",\"error\":\"wrong parameter\"}\n", body)
+
+	// correct request
+	r, err := http.Get(ts.URL + "/api/v1/qr/telegram?url=https://t.me/BotFather")
```
umputun, line 638: maybe it makes sense to store the QR image in testdata and compare the content here, just to make sure the body is matching? Sure, we don't want to test this library, but historically our rest tests act as some kind of semi-integration/functional and in this context the actual check of the body image makes sense

umputun (approved):
LGTM

PR #1255, opened 2022-02-04
feat(site): add anchor links

vanyauhalin:
- feat: markdown-it-anchor as local plugin
- feat: script for handling anchor

### DIFF
--- a/site/src/includes/anchor.js
+++ b/site/src/includes/anchor.js
@@ -1,0 +1,59 @@
+function createElement(message) {
+	const el = document.createElement('span');
+	el.innerHTML = message;
+	el.classList.add(
+		'animate-from-right-36',
+		'bg-brand-100',
+		'dark:bg-brand-900',
+		'dark:text-brand-400',
+		'dark:text-gray-300',
+		'fixed',
+		'font-medium',
+		'px-2',
+		'py-1',
+		'rounded',
+		'shadow',
+		'text-brand-600',
+		'text-center',
+		'w-36',
+		'z-10',
+		window.innerWidth > '768' ? 'top-20' : 'top-24',
+	);
+	el.setAttribute('role', 'alert');
+
+	document.body.after(el);
+
+	setTimeout(() => {
+		el.classList.remove('animate-from-right-36');
+		el.classList.add('animate-to-right-36');
+		setTimeout(() => {
+			el.remove();
+		}, 300);
+	}, 3000);
+}
+
+function showSuccess() {
+	createElement('Copied success');
+}
+
+function showError() {
+	createElement('Copied error');
+}
+
+function getURL(el) {
+	const url = window.location.href.replace(window.location.hash, '');
+	return `${url}#${el.parentElement.id}`;
+}
+
+function handleCopy(el) {
+	navigator.clipboard
+		.writeText(getURL(el))
+		.then(showSuccess)
+		.catch(showError);
+}
+
+window.onload = () => {
+	[...document.querySelectorAll('[data-permalink]')].forEach((el) => {
+		el.addEventListener('click', () => handleCopy(el));
+	});
+};
--- a/site/src/includes/components/head.njk
+++ b/site/src/includes/components/head.njk
@@ -17,4 +17,5 @@
     {% include "script.js" %}
   {% endset %}
   <script defer>{{ js | safe }}</script>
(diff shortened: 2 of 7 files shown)


akellbl4:
@vanyauhalin thank you for the PR! What's the reason for swapping the anchor module with your own implementation? Is it just a copy of the plugin? Then what's the reason for copying it into the project?

vanyauhalin:
> @vanyauhalin thank you for the PR! What's the reason for swapping the anchor module with your own implementation? Is it just a copy of the plugin? Then what's the reason for copying it into the project?

Yes, this is a copy of the plugin part, but I made a few changes: removed unused code and changed the `<a>` tag to the `<button`>. The button is semantically more appropriate since we're copying the link to the heading rather than going anywhere. Also, I follow the principle, if it is possible to do it yourself, there is no need to add it as a dependency.

As a notification to the user about the result of the action, I wrote a script that, like a react portal, creates a new element outside the body of the document. Unfortunately, I did not take into account all the options for using the notification (for example, multiple clicks), but I think this can be fixed in the future.

paskal (approved):

paskal:
Looks good for me. I don't have a problem embedding the code in case we need some customization.

The only thing missing is adding attribution to the author, @valeriangalliat, to the top of the file. Something like `Modified version of https://github.com/valeriangalliat/markdown-it-anchor`.

paskal:
Oh, could you also please add the screenshot before and after the change?

akellbl4:
I see few problems. New code doesn't have test but plugin does.  The plugin is being supported and new code have to be supported internally. It basically doesn't add new functionality. I mean we can achieve the same results with the plugin.

Also I suggest changing description because it doesn't describes changes. I see copy link implementation and it doesn't looks like anchor links because they already were there.

site/src/includes/anchor.js
```diff
@@ -0,0 +1,59 @@
...
+		'text-center',
+		'w-36',
+		'z-10',
+		window.innerWidth > '768' ? 'top-20' : 'top-24',
+	);
+	el.setAttribute('role', 'alert');
+
+	document.body.after(el);
+
+	setTimeout(() => {
```
akellbl4, line 26: Element have to be removed by end of animation not by timeout. Timeout should be as a fallback.

Use transitionend or animationend for subscription.

site/src/includes/components/head.njk
```diff
@@ -17,4 +17,5 @@
     {% include "script.js" %}
   {% endset %}
   <script defer>{{ js | safe }}</script>
+	<script>{% include "anchor.js" %}</script>
```
akellbl4, line 20: Why use separate place for inserting the code? We already have script file for that purpose.

akellbl4 (changes requested):

vanyauhalin:
> I see few problems. New code doesn't have test but plugin does. The plugin is being supported and new code have to be supported internally. It basically doesn't add new functionality. I mean we can achieve the same results with the plugin.
> 
> Also I suggest changing description because it doesn't describes changes. I see copy link implementation and it doesn't looks like anchor links because they already were there.

Yes, in theory, we can not add our own solution and use the plugin, but as I said before, the main reason is semantics. Plugin authors use the [`<a>`](https://github.com/valeriangalliat/markdown-it-anchor/blob/7a64eec879b7f1617800680ef87e5ac6235ed79a/permalink.js) as an anchor:

```html
<h1 id="heading">
  <a href="#heading">#</a>
</h1>
```

I don't quite understand why create a `<a>` element as an anchor if we have a heading that is already an anchor. In my implementation, I use a `<button>`, because it makes more sense:

```html
<h1 id="heading">
  <button>#</button>
</h1>
```

We can use the plugin and just replace the render function (see [`renderPermalink`](https://github.com/valeriangalliat/markdown-it-anchor/blob/7a64eec879b7f1617800680ef87e5ac6235ed79a/index.js#L97)), but I don't think it makes much sense, because in that case we can't rely on [tests](https://github.com/valeriangalliat/markdown-it-anchor/blob/7a64eec879b7f1617800680ef87e5ac6235ed79a/test.js), since they rely on the default render function. I see these options:

- I can continue to develop this request and make both the previously indicated edits and write tests.
- We sacrifice semantics and use the plugin without complicating the project.

The first option I like more.

PR #1257, opened 2022-02-06
update auth lib with support for work/school MS accounts

0xmohit:
Follow-up to https://github.com/go-pkgz/auth/pull/107


paskal:
I propose to close in favor of https://github.com/umputun/remark42/pull/1260, as you introduced unrelated changes likely due to the non-latest go version installed on the machine you used for preparing this PR.

0xmohit:
Turns out that I was using go 1.17 while the CI pipeline appears to be using 1.16

PR #1262, opened 2022-02-06
Remove explicit /srv/ usage from docker-compose files

paskal:
We have plenty of paths used in the application (like `AVATAR_FS_PATH`, `IMAGE_FS_PATH`), but two of them are hardcoded in examples all over the code for historical reasons.

I found that by default in Docker, the path would anyway resolve to the value we are setting it explicitly to, so it doesn't make sense to set a few variables we are setting now explicitly.

I was confused by these two variables when figuring out how to run the application outside of docker by looking at the default variables we propose in docker-compose.yml.

### DIFF
--- a/backend/_example/memory_store/compose-dev-memstore.yml
+++ b/backend/_example/memory_store/compose-dev-memstore.yml
@@ -29,7 +29,6 @@ services:
     environment:
       - REMARK_URL=http://127.0.0.1:8080
       - SECRET=123456
-      - BACKUP_PATH=/srv/var/backup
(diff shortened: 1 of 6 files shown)


umputun:
just to make sure: the reason to remove it is `/srv`WORKDIR in docker plus default values in cmd/server.go producing the same path anyway, correct?

paskal:
Correct, the binary starts from `/srv` and works with defaults starting with `./var/`, resulting in `/srv/var/` usage.

PR #1281, opened 2022-02-22
Remove unused CI argument from Dockerfiles

paskal:


umputun:
I don't think your assumption (i.e. unused) is accurate

as far as I can see, it is in use via /scripts/version.sh, see https://github.com/umputun/baseimage/blob/master/build.go/version.sh

PR #1318, opened 2022-04-10
add icon for paid patreon sub

### LINKED ISSUE #1125: Add support of Patreon auth

akellbl4:
Resolves #1125 

- had to fix timer tests due to flaky results


paskal:
Could you please add a screenshot?

Mavrin (dismissed):

PR #1320, opened 2022-04-10
fix copy for latest safari

### LINKED ISSUE #1312: Admin "Copy" button doesn't work in Safari
telegram-cloud-photo-size-2-5368320889772423807-x
Click on this button in current Safari on Desktop doesn't copy anything into clipboard.

akellbl4:
Resolves #1312

- updates typescript in order to have access to modern Clipboard API
- modern browsers dropped ability to copy formatter text, that led to disabling this feature. now it's only simple text
- fix for ts-ignore is needed, should be resolved in separate PR


akellbl4:
@Mavrin I tried to provide some tests for `copy` function but they are really far from ideal. If you know a better way for testing clipboard I would appreciate any suggestions.
As far as I know the best way is to test it in headless browser.

Mavrin (dismissed):

umputun:
conflict

PR #1325, opened 2022-04-12
update size-limit and move to app preset metrics

akellbl4:


akellbl4:
We've got **app preset** and it just created first record and didn't compare results from the previous result of **file preset**
CleanShot 2022-04-12 at 10 15 04@2x

umputun:
we good to merge it?

PR #1335, opened 2022-04-19
Replace errors package with fmt.Errorf

paskal:
https://gist.github.com/Peltoche/60b8b81dfbf70164d0e2b88988003229 was used for it, thanks to @Peltoche for publishing it.

I've reviewed each changed line manually and changed all wraps without checking for errors not equal to `nil` to have that check and return plain `nil` otherwise.


umputun (approved):
LGTM

PR #1338, opened 2022-04-26
Migrate notifications to the go-pkgz/notify module

paskal:
Notifications still work, I haven't tested webhook and slack but tested email and telegram:
image
image


paskal:
Done for all modules, I separated commits by modules for easier review.

umputun (commented):
LGTM. However, the part " I haven't tested webhook and slack" makes me a little bit uncomfortable. Could you please play with webhooks and slack a little bit, just in case?

paskal:
Webhook and Slack works as well:
image
image

Issue #1293, opened 2022-03-11
Apostroph character escaping issue in the profile panel


damienalexandre:
Using `umputun/remark42:master`, I get a escaping/encoding issue with the `'` character in the profile panel.

```html
<script type="[text/javascript]()">
    var remark_config = {
        host: '*******',
        site_id: '*****',
        components: ['embed'],
        url: 'https://*****/blog/lavantage-de-rouler-avant-tout',
        max_shown_comments: 20,
        theme: 'light',
        page_title: "L'avantage de rouler avant-tout",
        locale: 'fr',
        show_email_subscription: false
    };
</script>
```

The important line here is `page_title: "L'avantage de rouler avant-tout"`, the `'` is displayed like this in the user profile:

![image](https://user-images.githubusercontent.com/225704/157922964-9c429b4f-f8e6-45ab-8622-ac20357f9ded.png)

akellbl4:
@paskal we use `orig` field from a comment. It comes with `&#39;` in it. We do not render text html due to security reason. Could address this change to backend and do not encode this char like so?

paskal:
Thanks for the report! It was fixed in #1341, I've verified it:
<img width="433" alt="image" src="https://user-images.githubusercontent.com/712534/165991255-fa1b799c-4198-4e8e-a97b-41e01a3b9a4b.png">
Resolving.

PR #1341, opened 2022-04-27
Allow some escaped symbols in the title and name

### LINKED ISSUE #415: unescape step missing for the user name
This is an example:

8l221-201908-17133235-gp03k

Note: in the info wideget is seems to be handled correctly 
k8mh0-201908-17133330-pgcnn

umputun:
resolves #415 and the similar issue with the title

It allows `&`, `'` and `"` in those fields


akellbl4:
Another issue about this problem https://github.com/umputun/remark42/issues/1293

paskal (approved):

PR #1356, opened 2022-05-13
rename "store" to "storage" and "picture" to "image"

paskal:
Store sometimes was used verb (as "to store") and sometimes as a noun. This commit changes all the nouns to "storage" so that it would be easier to search through the code and documentation.

A picture was sometimes used as a synonym for an image, making it harder to find all references mentioned in the code. This commit changes a few easily changeable places to "image", making it easier to find everything about images with a single search term.

### DIFF
--- a/backend/_example/memory_store/README.md
+++ b/backend/_example/memory_store/README.md
@@ -1,4 +1,4 @@
-# sample store implementation
+# sample storage implementation
 
 `memory_store` illustrates how to make a custom storage plugin for remark42.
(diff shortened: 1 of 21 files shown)


umputun (commented):
i think this is kind of confusing and actually adds some inconsistency. It seems to change a bunch of comment from store to storage but keeps the actua vars, lile adminStore, storeEngine, etc as well as factories (makeBlahStore)

PR #1375, opened 2022-06-01
removes .git from build layer, emulates GH build

umputun:
_this related to the discussion in #269_

Remove .git from the docker build context and pass both GITHUB_REF and GITHUB_SHA from the make. It uses an [updated baseimage](https://github.com/umputun/baseimage/commit/40677b819d9b4b2082a5d0e826a46a312a88dfc8) defaulting to "local" if no git repo

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,6 +6,7 @@
 /.vscode/
 /.idea/
 /bin/
+/.git/
 
 # source files
 docker-compose.yml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM umputun/baseimage:buildgo-v1.8.0 as build-backend
+FROM umputun/baseimage:buildgo-v1.9.1 as build-backend
 
 ARG CI
 ARG GITHUB_REF
@@ -8,7 +8,6 @@ ARG SKIP_BACKEND_TEST
 ARG BACKEND_TEST_TIMEOUT
 
 ADD backend /build/backend
-ADD .git/ /build/backend/.git/
 WORKDIR /build/backend
 
 ENV GOFLAGS="-mod=vendor"
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,7 +72,7 @@ RUN if [ -z "$SKIP_FRONTEND_BUILD" ] ; then \
     ; fi
 RUN rm -rf ./node_modules
 
-FROM umputun/baseimage:app-v1.8.0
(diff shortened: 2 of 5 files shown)


umputun:
@nagimov @bnoctis pls take a look

paskal:
Doesn't it break the scenario where I just download the git repository and run `docker build`, producing `local` image instead of `master`? Even with `make rundev` in this branch I get `remark42 local-20220601T17:52:14`, I wonder if that's expected.

umputun:
yes, it does. it produces a "local" version unless it runs with `make docker`

paskal:
After last changes, `make rundev` still produces `local` build for me.

umputun:
> After last changes, `make rundev` still produces `local` build for me.

probably your local compose missing this part i have added https://github.com/umputun/remark42/pull/1375/commits/eafff01c5555dce96fbc8cf412930c9988820deb#diff-e1a5cf1531a00e0c8773ac834afc54983202fa6f7268566193ac6d09f45495d6R18

bnoctis:
I'll let @nagimov have the say, LGTM though

nagimov:
TLDR: LGTM
<hr>

Unrelated build issues:

- when building via `docker build -t remark42 --build-arg SKIP_FRONTEND_TEST=true --build-arg SKIP_BACKEND_TEST=true .` or `make docker`, getting the following error due to unset `$BUILDPLATFORM`. It seems like `Dockerfile` always expects `$BUILDPLATFORM` to be set, which only comes from `dockerx` make-target (via `--platform`), and not from `docker` make-target.
```
...
Step 15/52 : FROM --platform=$BUILDPLATFORM node:16.13.2-alpine as build-frontend-deps
failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument
make: *** [Makefile:14: docker] Error 1
```

- fixed with `sed -i 's/--platform=$BUILDPLATFORM//' Dockerfile`
- builds ok `docker build -t remark42 --build-arg SKIP_FRONTEND_TEST=true --build-arg SKIP_BACKEND_TEST=true .`, local instance `http://127.0.0.1/web/` works as expected, properly tagged with `version=gitless-docker-eafff01-YYYYMMDDTHH:MM:SS`
- `docker build -t remark42 .` gives a few failed backend tests, all frontend tests pass

paskal (approved):

PR #1378, opened 2022-06-05
show subscription buttons in simple view, add ability to hide rss button

akellbl4:
CleanShot 2022-06-05 at 15 22 54@2x

CleanShot 2022-06-05 at 15 24 19@2x


akellbl4:
cc @reorx

Mavrin (approved):

PR #1399, opened 2022-06-30
Add documentation on running remark42 on a separate domain

paskal:


site/src/docs/manuals/separate-domain/index.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Configure Instance on a different domain
+---
+
+## How to configure single remark42 instance for multiple domains
```
akellbl4, line 5: I think we should consistently write the name with capitalized `R`
```suggestion
## How to configure single Remark42 instance for multiple domains
```

akellbl4 (approved):
LGTM

umputun (approved):
good

PR #1404, opened 2022-07-01
Switch to pnpm

akellbl4:
|                                       | installation time | `node_modules` size  |    
|-----------------------|------|--------|
| NPM i          | 10.79s | 451M  |
| PNPM  |  **2.49s** | **342M**  |

It's okay that _size_ job is failing. It's because it can't install dependencies on master without `pnpm-lock.yml`


Dockerfile
```diff
@@ -45,9 +45,9 @@ RUN if [ -z "$SKIP_FRONTEND_BUILD" ] ; then \
 ADD frontend/package.json /srv/frontend/package.json
 ADD frontend/package-lock.json /srv/frontend/package-lock.json
 WORKDIR /srv/frontend
-RUN mkdir node_modules
 RUN if [ -z "$SKIP_FRONTEND_BUILD" ] ; then \
-    	CI=true npm ci --loglevel warn \
+			npm i -g pnpm && \
```
Mavrin, line 49: I  think we can use https://nodejs.org/api/corepack.html for activate pnpm
`corepack pnpm install`

Mavrin (dismissed):

umputun (dismissed):

Dockerfile
akellbl4: Hm, I think its too early for `corepack`. It was introduced in v18 and it's experimental.
But worth considering when we move towards update to node18

umputun (approved):

PR #1409, opened 2022-07-11
add turborepo

akellbl4:
- adds turborepo for package managment
- move lint-staged (it can work by package)


.gitignore
```diff
@@ -17,7 +17,7 @@ debug.test
 remark42
 /bin/
 /backend/var/
-/backend/app/var/
+/var/
```
paskal, line 20: That change doesn't seem right.

compose-dev-frontend.yml
```diff
@@ -13,8 +13,8 @@ services:
       dockerfile: Dockerfile
       args:
         - SKIP_BACKEND_TEST=true
-        - SKIP_FRONTEND_BUILD=true
-        # - NODE_ENV=development
+        # - SKIP_FRONTEND_BUILD=true
```
paskal, line 16: I think skipping frontend build should be enabled by default here, it's done that way to reduce the build time significantly.

akellbl4: dropped this change

.gitignore
akellbl4: dropped this change

paskal (approved):

Mavrin (approved):

PR #1417, opened 2022-07-13
use workspaces

akellbl4:
A part of #1409

- adds pnpm workspaces
- moves `frontend` to `frontend/apps/remark42` (99% of changes)
- updates ci jobs

**Note:** `size-limit` is crashing because code was moved and it expects it in the same place (when runs check on master)


frontend/.lintstagedrc.js
```diff
@@ -1,5 +1,14 @@
 module.exports = {
```
akellbl4, line 1: Should we use shared lint-staged config or should we have config per package?

If it is shared probably all of the lint/format packages should be shared as well

Dockerfile
```diff
@@ -105,4 +102,4 @@ HEALTHCHECK --interval=30s --timeout=3s CMD curl --fail http://localhost:8080/pi
 
 
 RUN chmod +x /srv/init.sh
-CMD ["/srv/remark42", "server"]
+CMD ["/srv/remark42", "server"]
```
Mavrin, line 105: Add new line please

Dockerfile.artifacts
```diff
@@ -77,4 +70,4 @@ RUN \
 FROM alpine
 COPY --from=build-backend /build/backend/remark42.* /artifacts/
 RUN ls -la /artifacts/*
-CMD ["sleep", "100"]
+CMD ["sleep", "100"]
```
Mavrin, line 73: Add new line please

Mavrin (approved):

paskal (changes requested):
LGTM, except for the fact that this change can't go without a single `.md` file change. Something should be altered in the frontend dev documentation and maybe a Readme file.

paskal:
Also please don't merge before #1422, I would like to rebase with newer Dockerfile.

paskal:
To be precise, I want to see edits in https://github.com/umputun/remark42/blob/master/site/src/docs/contributing/development/frontend/index.md.

paskal (dismissed):

paskal (approved):

PR #1418, opened 2022-07-13
update ci workflows

akellbl4:
unifies the way workflows defined, updates actions, gives names to jobs

Common:
- run jobs on push only for `master` (we don't need to run them for every push to every branch, when PR is opened jobs will be run for each push, now it runs twice)
- update `actions/checkout` to `v3`
- add names to jobs for better readability in PRs

Backend:
- move [ci-test-backend.yml → .github/workflows/ci-backend.yml](https://github.com/umputun/remark42/compare/workspaces...ci-workflows?expand=1#diff-d534d457d2f453ff292b326309bebf3d6b2f7e067ca4247038b61df22d09c728)  because if we have more backend jobs they will leave there, otherwise if `name` on the top of the file is the same they will be grouped
- use `name: backend` for grouping all of the backend related jobs

Frontend:
- move `size-limit` job to `ci-frontend.yml`

Site:
 - rename `build-site` to `build` because it's already prefixed with root name
 
 

CleanShot 2022-07-13 at 11 45 38@2x


.github/workflows/ci-backend.yml
```diff
@@ -1,8 +1,9 @@
-name: test_backend
+name: backend
 
 on:
   push:
     branches:
+      - master
```
umputun, line 6: I don't get it. It means one won't be able to run CI on a feature branch unless it merged to master

akellbl4: Unless you open a pr, push to master or merge to master.
Now it runs on every push (when PR is opened it runs twice because of push + pull_request)

.github/workflows/ci-frontend.yml
```diff
@@ -3,7 +3,7 @@ name: frontend
 on:
   push:
     branches:
-    tags:
+      - master
```
umputun, line 6: I'm not sure about this one either. If master pushed first and after this, we make a release tag I guess it will be fine. However, what will happen if we commit master locally, tag it locally and push the tag?

.github/workflows/ci-backend.yml
umputun: but this is exactly my flow. I do a new branch and expect pushed commits to run in CI. Postponing it till the merge request will significantly increase the feedback loop in this case

.github/workflows/ci-frontend.yml
akellbl4: I don't think we need to run this workflow in any case you mentioned, because we already ran them on PR merges. That means it's unnecessary runs.

.github/workflows/ci-backend.yml
akellbl4: rolledback

paskal (dismissed):

umputun (dismissed):

Mavrin (dismissed):

paskal (approved):

PR #1421, opened 2022-07-18
fix background color in dark mode

chuhlomin:
Fixing this:

Screen Shot 2022-07-17 at 22 45 49

### DIFF
--- a/frontend/app/styles/global.css
+++ b/frontend/app/styles/global.css
@@ -6,6 +6,7 @@ body {
   font-family: system-ui;
   font-size: 14px;
   color: rgb(var(--primary-text-color));
+  background-color: rgb(var(--primary-background-color));
   box-sizing: border-box;
 }
 


akellbl4:
@chuhlomin hey, thank you for contribution!
Can you please share info about browser in which this bug appears?

chuhlomin:
@akellbl4 it's in Safari 15.5 (17613.2.7.1.8)

akellbl4:
That's odd because I'm not getting this bug on any of those browsers.

Small explanation why the background was left without any color: We don't want to have block with comments to interfere with background of a website and it intend to be transparent. It's better to try `transparent` or `unset` in order to make the iframe injected in a site seamlessly.

akellbl4:
I haven't dug into it but this thing is causing the problem on your site. We experienced the same problem on radio-t.com and removing the prop resolved the issue.

CleanShot 2022-07-18 at 11 07 55@2x

PR #1426, opened 2022-07-21
Do not sanitise the original comment markdown

### LINKED ISSUE #1292: Code between < and > in code blocks disappears when opened for editing
When creating a comment, I can type code containing `<...>` within a backtick code clock. However, when I open this comment for editing, code between `<` and `>` disappears.

paskal:
Previously it was sanitised using the HTML sanitiser, but it had proven troublesome and unnecessary. Remark42 rendered the markdown into proper HTML, but then some pieces of it (like cited HTML code inside the code block, marked by backticks) were cut out, which then showed the incorrect markdown to a user when they were editing the comment.

For example, the comment "\`foo\<bar\>\`" became "foo" after sanitising, and despite the proper render user saw only "foo" when editing the comment.

After this change, the initial comment markdown is preserved unaltered. It could contain dangerous HTML with JS, which I assume shouldn't be a problem as it's never rendered as HTML but instead supposed to be converted to HTML by the interpreter. In Remark42, it's stored in a comment.Text field and sanitised and thus safe.

I've left information about the potential danger of rendering the original markdown as-is without an interpreter in all relevant places I could find.

Resolves #1292.

### DIFF
--- a/backend/app/rest/api/rest_private_test.go
+++ b/backend/app/rest/api/rest_private_test.go
@@ -56,7 +56,6 @@ func TestRest_Create(t *testing.T) {
 
 // based on issue https://github.com/umputun/remark42/issues/1292
 func TestRest_CreateFilteredCode(t *testing.T) {
-	t.Skip("not yet fixed")
 	ts, _, teardown := startupT(t)
 	defer teardown()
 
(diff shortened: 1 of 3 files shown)


umputun (approved):
LGTM

akellbl4 (approved):

PR #1427, opened 2022-07-22
fix: sort select background in dark mode

fish-404:
Fix sort select background in dark mode in order to see the text clearly. 

image

### DIFF
--- a/frontend/apps/remark42/app/components/select/select.module.css
+++ b/frontend/apps/remark42/app/components/select/select.module.css
@@ -17,6 +17,7 @@
 }
 
 .select {
+  background-color: rgb(var(--primary-background-color));
   position: absolute;
   top: 0;
   left: 0;


akellbl4:
@fish-404 hey, thanks for contribution. I'm curious what browser renders select dropdown like so?

fish-404:
> @fish-404 hey, thanks for contribution. I'm curious what browser renders select dropdown like so?

@akellbl4 Hey, I use Firefox 100.0.2, inspect the default `background-color` is `rgb(233, 233, 237)`.

Take Edge and Chrome to test, their default `background-color` is `rgb(255, 255, 255)` which can see the text clearly.

akellbl4:
@fish-404 is it on Windows?

akellbl4:
I've checked the problem. It should be fixed in slightly different way. The problem is in `color-scheme`. We have an issue connected with it https://github.com/umputun/remark42/issues/1430
Can you help with it?

fish-404:
> I've checked the problem. It should be fixed in slightly different way. The problem is in `color-scheme`. We have an issue connected with it #1430 Can you help with it?

OK, but I'm not sure if I am able to help, will check it later.

paskal:
thanks for the original report and patch — this one sat too long. closing in favour of #2023, which just merged and fixes the underlying `color-scheme` problem at the source: the iframe now sets `color-scheme: dark` / `light` on the inner `<html>` to match the active theme, so Firefox uses the proper dark user-agent chrome for native form controls (including the native `<select>` dropdown popup).

the one-line change here would layer on top of that by forcing `background-color` on the (invisible, `opacity: 0`) `<select>` element, which on some Firefox versions can actually suppress the `color-scheme`-driven native dropdown theming — so merging it now could partially undo #2023.

if you still see a problem with the sort select in dark Firefox after #2023, please reopen or file a fresh issue with a screenshot against current master — happy to take another look.

PR #1431, opened 2022-07-24
Move common options to cmd.go

paskal:
Timeout, admin password and site id are set in many commands, and we need to take care of synchronising the descriptions and flags between them.

This change moves these standard options to cmd.go importing them in the same manner CommonOpts imported by all commands already.

### DIFF
--- a/backend/app/cmd/cmd.go
+++ b/backend/app/cmd/cmd.go
@@ -31,6 +31,14 @@ type CommonOpts struct {
 	Revision     string
 }
 
+// SupportCmdOpts is set of commands shared among similar commands like backup/restore and such.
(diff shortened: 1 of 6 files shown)


backend/app/cmd/cmd.go
```diff
@@ -31,6 +31,20 @@ type CommonOpts struct {
 	Revision     string
 }
 
+// AdminPasswordOpt is shared among many commands for setting the server password
+// and then using it for backup/restore and such
+type AdminPasswordOpt struct {
```
umputun, line 36: This one seems like a little bit too much inderection. I would rather put it to SupportCmdOpts and keep it as a "naked" (direct) option in the primary server opts

PR #1434, opened 2022-07-24
Add docker image labels, add docker image badge to readme

paskal:
Specification for labels: https://github.com/opencontainers/image-spec/blob/main/annotations.md


Dockerfile.artifacts
```diff
@@ -67,6 +67,18 @@ RUN \
 
 
 FROM alpine
+
+ARG GITHUB_SHA
+
+LABEL org.opencontainers.image.authors="Umputun <umputun@gmail.com>" \
```
umputun, line 73: Unless I forgot something, I don't see why it is needed in the artifact container. This one is an intermediate thingy and does not produce any image, but rather the binary archives

umputun (approved):
LGTM

PR #1435, opened 2022-07-24
docs(manuals): add ts version of react component

openscript:
Some time ago Gatsby started to support Typescript natively. Also the frontend of remark42 is written in Typescript. So I thought it's a good idea to add a Typescript version of the component.


akellbl4 (approved):
Thanks for contribution!

umputun (approved):
thx

Issue #103, opened 2018-06-23
Add index.html with instructions


umputun:
_proposed by https://github.com/umputun/remark/issues/26#issuecomment-399681446_

We need a static HTML file with some details on how to integrate remark42, how to customize and so on. 

_related to #27_

Guria:
@umputun could you create html stub that will be served by remark at root.
Currently it only serves WebRoot at `/web/`. I am not feel that I can extend routing config to serve single file at root myself.

Guria:
I think I'll use serving a folder. Can you suggest how to make a redirect from root to '/getstarted'?

umputun:
I can add routing from `/index.html` to `/web/getstarted.html` file. Can you keep it a single file?

PR #1442, opened 2022-07-27
Get rid of getstarted.html mentions

paskal:
@umputun introduced that reference in 70649b271, but I can't find any references to it or the file itself now.


umputun:
This was done back in days we had very limited docs. I think it is fine to drop it

https://github.com/umputun/remark42/issues/103

umputun (approved):

PR #1444, opened 2022-07-27
Move templates from rakyll/statik to go:embed

paskal:
There is no need for the rakyll/statik package starting with Go 1.16, which provides us with tools for embedding files without third-party libraries.

I've verified that binary reads files from the disk in case they are present with the same name as embedded ones. Changes in files are immediately seen on the page after refresh. To verify, try opening http://127.0.0.1:8080/email/unsubscribe.html and putting file `error_response.html.tmpl` next to the library and altering it.


backend/app/rest/api/rest_private_test.go
```diff
@@ -610,7 +610,6 @@ func TestRest_EmailAndTelegram(t *testing.T) {
 	ts, srv, teardown := startupT(t)
 	defer teardown()
```
umputun, line 612: do we still need MockFS?

Dockerfile
```diff
@@ -102,7 +102,6 @@ ADD backend/scripts/import.sh /usr/local/bin/import
 RUN chmod +x /entrypoint.sh /usr/local/bin/backup /usr/local/bin/restore /usr/local/bin/import
 
 COPY --from=build-backend /build/backend/remark42 /srv/remark42
-COPY --from=build-backend /build/backend/templates /srv
 COPY --from=build-frontend /srv/frontend/apps/remark42/public/ /srv/web/
```
umputun, line 105: I don't recall why we needed templates copied in the final container, but I think this was not due to statik?

paskal: It's not necessary, and I explained how to replace any template file with your own as clearly as I could in #1450.

umputun (approved):

PR #1446, opened 2022-07-27
Move static web files from rakyll/statik to go:embed 

paskal:
There is no need for the rakyll/statik package starting with Go 1.16, which provides us with tools for embedding files without third-party libraries.


site/src/docs/contributing/backend/index.md
```diff
@@ -30,21 +30,19 @@ Run tests in your IDE, and re-run `make rundev` each time you want to see how yo
...
 
-In order to have working Remark42 installation you need once to copy frontend static files to `./backend/web` directory from `master` docker image, and also copy files from `./templates` to the `./backend` as they are expected to be where application starts:
+In order to have working Remark42 installation you need once to copy frontend static files to `./backend/web` directory from `master` docker image, as it is expected to be where application compiles:
 
 ```shell
 # frontend files
 docker pull umputun/remark42:master
 docker create -ti --name remark42files umputun/remark42:master sh
-docker cp remark42files:/srv/web/ ./backend/
+docker cp remark42files:/srv/web/ ./backend/app/cmd/
```
paskal, line 39: This might be not working anymore as the /web/ directory is now embedded into binary and not present in the Dockerfile. I'll figure out what to do with that a bit later.

docker-init.sh
```diff
@@ -3,11 +3,6 @@ echo "prepare environment"
 # replace {% REMARK_URL %} by content of REMARK_URL variable
 find . -regex '.*\.\(html\|js\|mjs\)$' -print -exec sed -i "s|{% REMARK_URL %}|${REMARK_URL}|g" {} \;
```
umputun, line 5: what was the  reason to remove this replacement with sed?

paskal: Because I didn't add the `web` directory to the Dockerfile. But I decided to keep it there for simplicity.

umputun (approved):

PR #1447, opened 2022-07-27
Add Traditional Chinese (zh-tw) translations

orangesobeautiful:
The original chinese translations (zh) is only for Simplified Chinese.
It has different character sets  from Traditional Chinese .
I add translations file for Traditional Chinese .


paskal (approved):

umputun (approved):
thx

PR #1448, opened 2022-07-27
Do not run size limit job outside of PRs

### LINKED ISSUE #1433: [CI] Frontend task "Check bundle size" is broken in master
Example: [link](https://github.com/umputun/remark42/runs/7461670696?check_suite_focus=true#step:4:9)

> Error: No PR found. Only pull_request workflows are supported.

We should disable this step run in case CI runs outside of the pull request.

paskal:
Currently, this step emits the "Error: No PR found. Only pull_request workflows are supported." message when run on the master branch commits (after the merge), so the change prevents it from being run there.

Resolves #1433.

### DIFF
--- a/.github/workflows/ci-frontend.yml
+++ b/.github/workflows/ci-frontend.yml
@@ -150,6 +150,7 @@ jobs:
   size-limit:
     name: Size limit
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


akellbl4 (approved):
Awesome!

umputun (approved):

PR #1463, opened 2022-08-22
fix: anon cannot request data removal

### LINKED ISSUE #1460: [UI] do not show "Request my data removal" for anonymous users
image

That button yields 403 for anonym users and shouldn't be displayed.

akellbl4:
Resolves #1460


paskal (approved):
Button disappeared, as expected.

umputun (approved):

PR #1464, opened 2022-08-23
e2e tests

akellbl4:
- adds playwright as testing framework
- adds a test with comment posting as dev user

### DIFF
--- a/compose-e2e-test.yml
+++ b/compose-e2e-test.yml
@@ -1,0 +1,36 @@
+# compose for running e2e tests in CI
+version: "2"
+
+services:
+  remark42:
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        - SKIP_BACKEND_TEST=true
+        - SKIP_FRONTEND_TEST=true
+
+    image: umputun/remark42:dev
+    container_name: "remark42"
+
+    environment:
+      # remark42 hostname used for proper dev auth work for tests container which connects to it
+      # using the such hostname (unlike local development where 127.0.0.1 is used)
+      - REMARK_URL=http://remark42:8080
+      - SECRET=12345
+      - DEBUG=true
+      - ADMIN_PASSWD=password
+      - AUTH_DEV=true # activate local OAuth "dev" listening on http://remark42:8084
+      - ADMIN_SHARED_ID=dev_user # set admin flag for default user on local oauth2
+      - AUTH_ANON=true
+      - AUTH_EMAIL_ENABLE=true
+    volumes:
+      - ./var:/srv/var
+
+  tests:
+    build:
+      context: ./frontend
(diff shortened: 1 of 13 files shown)


akellbl4:
@umputun @paskal how can I pass `$BUILDPLATFORM` to [here](https://github.com/umputun/remark42/pull/1464/files#diff-194218c48b9a0cdd03974145733804c2d992ca818529fe2fa69a501d8b5b1cc3R24)
 https://github.com/umputun/remark42/runs/7965644294?check_suite_focus=true#step:4:9

Mavrin (dismissed):

paskal (changes requested):
Please tidy up the commit history. Twenty-eight commits, including merge ones, is too much.

Also, I expect something about e2e testing to be mentioned in the frontend Readme file.

compose-e2e-test.yml
```diff
@@ -0,0 +1,37 @@
...
+    container_name: "remark42"
+
+    environment:
+      - REMARK_URL=http://remark42:8080
+      - SECRET=12345
+      - DEBUG=true
+      - ADMIN_PASSWD=password
+      - AUTH_DEV=true # activate local oauth "dev"
+      - ADMIN_SHARED_ID=dev_user # set admin flag for default user on local oauth2
+      - POSITIVE_SCORE=false # restricts comment's score to be only positive
```
paskal, line 24: False by default, no point adding it here.

compose-e2e-test.yml
```diff
@@ -0,0 +1,37 @@
...
+      - EDIT_TIME=5m # edit window
```
paskal, line 25: The default value as well should be removed.

compose-e2e-test.yml
akellbl4: Then it should be fixed on dev composes.

compose-e2e-test.yml
akellbl4: Same here

compose-e2e-test.yml
paskal: Good point. Feel free to do so: it indeed doesn't make sense. I assume some frontend developer was playing with non-default values of these options and then didn't exclude them from the commit: 5809419bcece68350d55e4fe95076edfcf917250.

compose-e2e-test.yml
```diff
@@ -0,0 +1,37 @@
+# compose for running e2e tests in CI
+version: "2"
+
+services:
+  remark42:
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        - BUILDPLATFORM=linux/amd64
```
paskal, line 10: Is that line needed?

paskal (approved):

Mavrin (approved):

PR #1473, opened 2022-08-26
Bump go modules, make auth dev hostname customisable

paskal:
After this commit, dev auth would start working with the `REMARK_URL` hostname instead of the previously hardcoded 127.0.0.1.

Breaks development setup where `REMARK_URL` was set to a non-standard value and dev auth was running on 127.0.0.1 and working, as, after that change, it would stop working.


umputun:
I think this change is potentially breaking (well, a little bit) and may need some explanation in docs. The only use case I can think of - is some local testing user sets a fake domain in remark_url but want dev auth to go to 127.0.0.1. I don't think such a rare/exotic case warrants an extra parameter, but maybe we want to indicate in docs what prior to version x.y.z.tThis always went to 127.0.0.1

paskal:
@umputun I've updated development documentation as well as development docker-compose files; please review once more.

umputun (approved):
LGTM

PR #1474, opened 2022-08-26
Add compose files to dockerignore, remove defaults from frontend compose

paskal:
Before that change, docker would create a new image on docker-compose file changes.

Frontend docker-compose file change removes options set to the same values as their default values.


umputun (approved):
LGTM

PR #1485, opened 2022-09-11
Add locale Thai language

kaikunz:
th

PR #1486, opened 2022-09-11
Update supportedLocales.json Add Thai language

kaikunz:
add th


paskal (approved):

paskal:
It should be merged after #1485.

PR #1490, opened 2022-09-14
fix iframe resize on auth dropdown opening

akellbl4:
Mutation observer was listening for all of the elements. When Auth dropdown appears it renders images and triggering height recalculations.
Resize trigger on image load was introduced in favor of resizing iframe for users with slow internet (when images load later and we need to resize iframe accordingly)

Solution:
Listen to mutations only on block with comments.
Plus remove unnecessary triggers.


paskal:
I installed it on https://terrty.net/2022/self-hosted-comments-remark42/

Safari, Chrome first load in Incognito:
image
After refresh:
image

akellbl4:
@paskal thank you for testing the fix. I see in prod env it behave slightly different.
Probably should be a good idea to introduce e2e here

Mavrin (approved):

PR #1500, opened 2022-09-28
Call admin store with a proper key

### LINKED ISSUE #1499: Key() method invoked with wrong siteID
Hello,
I noticed that in some places the `Key()` method of admin `Store` is called with an incorrect site id, precisely:

https://github.com/umputun/remark42/blob/master/backend/app/store/service/service.go#L986

```go
func (s *DataStore) getSecret(siteID string) (secret string, err error) {
	if secret, err = s.AdminStore.Key("any"); err != nil {
```

https://github.com/umputun/remark42/blob/master/backend/app/cmd/server.go#L1110

```go
SecretReader: token.SecretFunc(func(aud string) (string, error) { // get secret per site
	return admns.Key("")
}),
```

This is not a problem for the "shared" implementation since it supports only one set of data for all sites, however that's not the case for my custom RPC backend... moreover, the fix seems trivial in both cases:

https://github.com/umputun/remark42/blob/master/backend/app/store/service/service.go#L986

```go
func (s *DataStore) getSecret(siteID string) (secret string, err error) {
	if secret, err = s.AdminStore.Key(siteID); err != nil {
```

https://github.com/umputun/remark42/blob/master/backend/app/cmd/server.go#L1110

```go
SecretReader: token.SecretFunc(func(aud string) (string, error) { // get secret per site
        // go-pkgz/auth.go says:
        //       "SecretReader   token.Secret        // reader returns secret for given site id (aud), required"
	return admns.Key(aud)
}),
```

Thanks

paskal:
Previously it was set to fixed strings, likely a test artefact.

Resolves #1499.

### DIFF
--- a/backend/app/store/service/service.go
+++ b/backend/app/store/service/service.go
@@ -981,9 +981,9 @@ func (s *DataStore) prepVotes(c store.Comment, user store.User) store.Comment {
 }
 
 // get secret for given siteID
-// Note: secret shared across sites, but some sites can be disabled.
+// Note: siteID ignored for the default admin.Static store
 func (s *DataStore) getSecret(siteID string) (secret string, err error) {
-	if secret, err = s.AdminStore.Key("any"); err != nil {
+	if secret, err = s.AdminStore.Key(siteID); err != nil {
 		return "", fmt.Errorf("can't get secret for site %s: %w", siteID, err)
 	}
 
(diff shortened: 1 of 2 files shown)


backend/app/store/service/service.go
```diff
@@ -983,7 +983,7 @@ func (s *DataStore) prepVotes(c store.Comment, user store.User) store.Comment {
 // get secret for given siteID
 // Note: secret shared across sites, but some sites can be disabled.
 func (s *DataStore) getSecret(siteID string) (secret string, err error) {
-	if secret, err = s.AdminStore.Key("any"); err != nil {
+	if secret, err = s.AdminStore.Key(siteID); err != nil {
```
umputun, line 986: would be nice to add a comment like "note: siteID ignored for the default admin.Static store". The current comment is misleading as after this change secret is not shared (well, for some admin stores) anymore

paskal: I've altered the comment as you proposed, but I haven't looked into the issue deeply to know the precise results of that change. I verified that it doesn't seem to be a breaking change and settled on that.

umputun (approved):
LGTM

backend/app/store/service/service.go
umputun: i have checked what it can affect quickly, and it looks safe to me

PR #1507, opened 2022-10-02
Clean up comment HasReplies cache on child comment deletion

### LINKED ISSUE #1481: Cannot delete/edit comment with replies even if they're removed
Steps to reproduce:

1. Login as a non-administrator;
2. Post one comment "a";
3. Reply to that comment with another comment "b";
4. Delete the "b" comment;
5. Delete the "a" comment;

An error occurs on the last step: "parent comment with reply can't be edited"... but this shouldn't be thrown because the child comment "b" is no longer visible. Furthermore, the "Edit" action is not restored on the parent comment once the child comment has been deleted.

No errors are thrown if the user is an administrator instead, and also the "Edit" is restored after refreshing the page. 
I noticed that the "Delete" button calls two different APIs depending if the user is an administrator or not:

- **Administrator**:  DELETE /admin/comment/{id}
- **Non administrator**: PUT /comment/{id}

The problem seems related to the `repliesCache` used by `HasReplies()` method that is not updated when the child comment is removed.

paskal:
Previously, the cache kept the entry and deletion of the parent comment after child deletion was not possible for the rest of cache life (5m) duration. Now it's possible to delete a parent comment after the deletion of the child comment by a user or admin.

Resolves #1481


umputun (commented):
looking good. the only thing i would suggest is to add tests on the service level, not just functional test on the rest level, as you did. Those test should be relatively easy to do as you have access to the cache and can check if comments added/removed. Alternatively, you can define a local interface for repliesCache.LoadincCache, mock it and check the interaction with the mock.

umputun (approved):

PR #757, opened 2020-08-02
Full text search

vdimir:
Resolves #734

This PR suggests full text search implementation. 
There are two search backends [bleve](https://github.com/blevesearch/bleve) and ElasticSearch (via [go-elasticsearch](https://github.com/elastic/go-elasticsearch) client).

Also I have considered implementation of inverted index by hands. It is required only implementation `indexer` interface with one method `IndexDocument` because the maintenance (such as batching, startup/shutdown etc.) is already done in front of bleve.

I have performed kind of load testing. I downloaded a bunch of reddit comments (to be close to real data) converted it to remark backup format, restored it, and ran request from local computer to droplet with disabled ratelimiter. Script for downloading and processing data and python notebook with testing can be found in [here](https://gist.github.com/vdimir/662a76926717c49b8b554af79792211e).

I was testing on ~80.000 comments restored from backup generating and sending requests from 8 threads.

Elastic behaves good.

Bleve slow downs remark.
When search and comment requests were sent, comment had timing 400ms @ p99 and 100 ms @ p50, search – 1000ms @ p99 and 200 @ p50, total rps ~40.

Also when indexing was enabled and only commenting was performed, latency spikes slowed down requests to >1000 ms every 5-10 sec. But rest 95% requests are good (~100 ms), total rps ~100.
So, I think spikes are connected with batch flushing, but I think this place requires attention.

Actually I think Elastic is an overkill for remark and I have implemented it only to ensure that interfaces are extensible to support such things.

Also I feel that my implementation is weak at some points but I'm not sure which things need improvement. 
**So, all comments and review will be really appreciated**. Especially I worry about architecture and working with concurrency. I am ready to make any changes starting from minor fixes to changing architecture at all.

Also sorry for mistakes in text, ask me about everything, I will explain if something isn't clear.

PR #1527, opened 2022-11-06
Remove duplicate type definitions from function signatures

paskal:
I haven't found a linter for these, so I had to catch these manually. I found #757 to fix one of these, and I thought it would be good to fix everything at once.

### DIFF
--- a/backend/app/migrator/backup.go
+++ b/backend/app/migrator/backup.go
@@ -83,7 +83,7 @@ func (ab AutoBackup) removeOldBackupFiles() {
 			backFiles = append(backFiles, info)
 		}
 	}
-	sort.Slice(backFiles, func(i int, j int) bool { return backFiles[i].Name() < backFiles[j].Name() })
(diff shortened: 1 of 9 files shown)


backend/app/rest/api/ssl.go
```diff
@@ -111,15 +111,11 @@ func (s *Rest) getRemarkHost() string {
 
 func (s *Rest) makeTLSConfig() *tls.Config {
 	return &tls.Config{
-		PreferServerCipherSuites: true,
```
paskal, line 114: This is not used since 1.17, https://github.com/golang/go/commit/9d0819b27ca248f9949e7cf6bf7cb9fe7cf574e8

VladimirZaets:
@paskal I personally prefer full declaration:)

umputun:
this pr is a little bit confusing. I don't know how safe the part about TLS change (will need to check) but  for sure it is not a part of "remove duplicate definition from func"

can you pls keep only those lint related changes here and make a separate pr for tls config?

paskal:
@umputun done, feel free to re-review.

umputun (approved):

PR #1528, opened 2022-11-06
Remove the unused DataStore.Put method

paskal:
It was initially done in #757, cutting into a separate change.

### DIFF
--- a/backend/app/store/service/service.go
+++ b/backend/app/store/service/service.go
@@ -154,12 +154,6 @@ func (s *DataStore) Get(locator store.Locator, commentID string, user store.User
 	return s.alterComment(c, user), nil
 }
 
-// Put updates comment, mutable parts only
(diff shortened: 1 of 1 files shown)


VladimirZaets:
@paskal Hi, just to clarify for myself. Even if it unused it is part of the contract and it is backward incompatible changes.

paskal:
It's an end product and not a library. I would consider CLI options and user-facing behaviour as part of the compatibility contract, which is why we never deprecated any CLI options. Also, storage interfaces whose usage is demonstrated in [example memory store](https://github.com/umputun/remark42/tree/master/backend/_example/memory_store) are part of the compatibility contract.

I don't consider unused and untested functions from Datastore as part of it.

umputun (commented):
I'm not sure about this one. The method is kind of usefull and I can see us using it some day. I'd rather keep it

PR #1544, opened 2022-12-03
Deprecate secret CLI param requirement from all commands but "server"

paskal:
It was planned to be used in the import, restore, and backup commands, but later, Umputun ditched that idea. After that, one less CLI option to set for all the commands other than `server`.

Backwards compatibility preserved: providing the secret won't fail the command execution.

### DIFF
--- a/backend/app/cmd/avatar.go
+++ b/backend/app/cmd/avatar.go
@@ -19,6 +19,7 @@ type AvatarCommand struct {
 
 	migrator AvatarMigrator
 	CommonOpts
+	DeprecatedSharedSecret
 }
 
 // AvatarMigrator defines interface for migration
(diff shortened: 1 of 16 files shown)


umputun:
Still, I'm not too fond of it. You are probably correct - this change won't break back compatibility, but it adds extra complexity, and I wonder why we care about it. Keeping a shared secret field in `CommonOpts` doesn't hurt anything, but all the code to handle `DeprecatedSharedSecret` (needed for back compact) feels like something artificial and unnatural to me. 

I'd say - let's keep it in `CommonOpts`, and if you think the presence of this flag is causing some confusion for a reader - we can add a comment saying, "currently used by `ServerCommand` only and ignored by others". 

_Generally, I agree with you that this flag is not common anymore, however, we want to keep it back compact, and the best way to do it is to change nothing and accept the fact - it is still here and (maybe) will have a use by some other commands some day._

PR #1547, opened 2022-12-04
refac: use resize observer instead of mutation observer

akellbl4:
Refactor resizing mechanism with more sutable browser API.

### DIFF
--- a/frontend/apps/remark42/app/components/auth/auth.hooks.ts
+++ b/frontend/apps/remark42/app/components/auth/auth.hooks.ts
@@ -76,11 +76,11 @@ export function useDropdown(disableClosing?: boolean) {
 
     handleChangeIframeSize(dropdownElement);
 
-    const observer = new MutationObserver(() => {
+    const observer = new ResizeObserver(() => {
       handleChangeIframeSize(dropdownElement);
     });
 
-    observer.observe(dropdownElement, { attributes: true, childList: true, subtree: true });
+    observer.observe(dropdownElement);
 
     return () => {
       document.body.style.removeProperty('min-height');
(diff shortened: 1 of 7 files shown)


frontend/apps/remark42/app/components/root/root.tsx
```diff
@@ -358,6 +344,14 @@ export function ConnectedRoot() {
   const props = useSelector(mapStateToProps);
   const actions = useActions(boundActions);
 
+  useEffect(() => {
+    const observer = new ResizeObserver(updateIframeHeight);
+
+    updateIframeHeight();
+    observer.observe(document.body);
```
akellbl4, line 351: Not sure if it's a good idea to listen on body.

frontend/apps/remark42/app/components/auth/auth.hooks.ts
```diff
@@ -76,11 +76,11 @@ export function useDropdown(disableClosing?: boolean) {
 
     handleChangeIframeSize(dropdownElement);
 
-    const observer = new MutationObserver(() => {
+    const observer = new ResizeObserver(() => {
```
Mavrin, line 79: I am not sure. Maybe it isn't necessary to have this `observer` and ResizeObserver on `body` will be enough.

Mavrin (approved):

frontend/apps/remark42/app/components/auth/auth.hooks.ts
akellbl4: Unfortunately we need it. I attached the root observer to `body` because I hoped it should see size changes for this thing too. As it turned out it can't spot size change because dropdowns positioned absolutely. I haven't dug into it maybe there is a way to have only one observer. For now I hope this thing will work better than our previous method.

Mavrin: Yep, got it

PR #1555, opened 2023-01-01
Update comment-form__subscribe-by-email.tsx

markmonroy:
Adds clarity to sign in token placeholder


paskal:
I've prepared #1557 with the proper fix, unfortunately, updating just the files you changed is not enough.

PR #1566, opened 2023-01-08
Move frontend nvmrc to /frontend/

paskal:
Previously we had two .nvmrc files in two locations out of three using node.


akellbl4 (approved):
@paskal thanks for the good call and PR

Issue #651, opened 2020-04-08
Update subscribe methods to get params on backend by request body


akellbl4:
Now we have methods like `/email/subscribe`, `/email/confirm` and these API methods implemented as POST and PUT. But now they are getting parameters from query string.

I think they should get params from request body because it is usual way to pass data over POST/PUT requests. Also, it will be good to send an email/token as is without URL encoding.

paskal:
@akellbl4, the change is merged. Feel free to change the frontend calls to utilise the body. `tkn` became `token` in the body, and everything else stayed the same.

paskal:
@Jaskon this one already has backend support as well, feel free to pick it up if you'll have time.

PR #1572, opened 2023-01-08
Use the request body for email subscription endpoints

paskal:
Previously, the endpoints were using query parameters. After this change, the body is tried to be parsed.

Backend part for #651.


backend/app/rest/api/rest_private.go
```diff
@@ -307,28 +307,42 @@ func (s *private) getEmailCtrl(w http.ResponseWriter, r *http.Request) {
 }
 
 // sendEmailConfirmationCtrl gets address and siteID from query, makes confirmation token and sends it to user.
-// GET /email/subscribe?site=siteID&address=someone@example.com
+// POST /email/subscribe with site and address in json body
```
umputun, line 310: This change is not back-compatible, and I don't see a good reason for this. I'd suggest supporting both existing query params and fallback to the body

paskal: I've already done that. The difference is that old behaviour is not mentioned in the documentation anymore.

I even left the tests with the old behaviour to verify it still works.

umputun: my bad, i noticed the removal of address and siteID in diff and have not looked further ;)

umputun (approved):
LGTM

Issue #809, opened 2020-11-05
Users posting links with destination "URL" instead of intended one


ninedraft:
Not sure if bug or UI misuse, but sometimes links are replaced with `$REMARK_DOMAIN/web/url` URL.

For example checkout this page: [https://radio-t.com/p/2020/11/03/prep-727/#comments](https://radio-t.com/p/2020/11/03/prep-727/#comments)

Examples of generated HTML:
+ `<a href="url" rel="nofollow">https://winxp.now.sh/</a>`
+ `<a href="url" rel="nofollow">https://samy.pl/slipstream/</a>`

If somebody can reproduce, please answer - I can spent some time to fix this bug (if presented).

umputun:
I can confirm - this is not an imaginary issue but the real one. I have seen it a few times on the demo site, but can't reproduce it. 

this is what I have tried

```
- http://google.com
- https://google.com
- [link](http://google.com)
- <a href="https://google.com" blah="xyz">link</a>
```

All the above make the expected URL. The only thing remotely similar to the reported issue is an attempt to publish a relative link, for example:

- `[link](google.com)` -> `https://demo.remark42.com/web/google.com`
- `<a href="google.com" blah="xyz">link</a>` -> `https://demo.remark42.com/web/google.com`
- `[link](/google.com)` -> `https://demo.remark42.com/google.com`

None of those really surprising, and such links make no sense in general.

umputun:
ok,  from what I see the incorrect one was submitted as this ` {"text":"[https://samy.pl/slipstream/](url)\n\n**NAT** **Slipstreaming
** - новый метод обхода NAT и файервола. \n\nПозволяет получить удаленный доступ к любой службе TCP/UDP путем посещения жертвой веб-сайта. "}`

I think this is **the user's error**. Highly unlikely something like this may happen if the user typing URL in MD manually, so I guess the user clicked on the button (toolbar), which populates `[](url)` by default and typed the link to the wrong place (inside `[]`). If this is the case, it is probably a usability issue but not a bug. We can address it by prepopulating both parts by default, I.e., something like `[title](URL)` or, even better - on a click, open a dialog box with a pair of text inputs and asking for both fields.

cc @akellbl4 

_note: just in case if a reader wonders how I got access to the log - this is not because remark42 reports to the mothership, but because I'm the admin for that particular site (radio-t.com)_

ninedraft:
Maybe a linter-like functionality for markdown renderer can be a good idea.

[...]

PR #1578, opened 2023-01-09
Don't allow relative links in comments

paskal:
(url) is a text inserted by default and never an intended URL.

That additional validation will ensure that users won't post relative links because they are rarely intended.

This is how it looks without frontend changes:

image

Backend part for #809.

### DIFF
--- a/backend/app/rest/api/rest_private_test.go
+++ b/backend/app/rest/api/rest_private_test.go
@@ -166,6 +166,25 @@ func TestRest_CreateWithRestrictedWord(t *testing.T) {
 	assert.Equal(t, "invalid comment", c["details"])
 }
 
+func TestRest_CreateRelativeURL(t *testing.T) {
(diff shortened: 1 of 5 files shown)


akellbl4:
Are we gonna throw an error in case link provided without protocol? Like `[google](google.com)`. Because it results in a link that refers to `${currentUrl}/google.com`

paskal:
This change prohibits posting relative links. I have to think about how to parse markdown correctly, but the change is already working and can be tested. @umputun, what do you think about the whole idea of blocking the relative URLs?

umputun:
Using relative links intentionally is an extremely rare use case. One would need to have a file server on the same subdomain as Remark42 to make any sense of this. I believe disabling such links is a good idea and will work well in 99.9% of cases.

umputun (approved):
LGTM

Issue #556, opened 2020-01-16
Cache control for UI assets


umputun:
UI updates may not be picked up by browsers or even worse - picked up partially due to the caching. 

Currently, we let browsers decide what and how to cache and not enforcing any control. I have tried to do it with https://github.com/umputun/remark/commit/b0898cd1f3cd29d3c4fb8ddcc4df370bbb7278d7 and technically it does what I've planned - sets etag as a mix of url & version and checks`If-None-Match` tag matches. Probably it is the right thing to do, but it doesn't help with refresh/update issues.

#554 is trying to address it from the frontend side.

paskal:
I've observed following behavior on reloading http://127.0.0.1:8080/web/remark.js without cache (Cmd + Shift + R on MacOS) using Firefox with added debug logging:

    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:583 api.cacheControl.func2.1} cacheControl request for /web/remark.js
    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:475 api.addFileServer.func1} addFileServer request for /web/remark.js
    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:583 api.cacheControl.func2.1} cacheControl request for /web/remark.js
    remark42-dev | 2020/01/18 06:35:14.217 [DEBUG] {api/rest.go:475 api.addFileServer.func1} addFileServer request for /web/remark.js

For some reason request for a static asset is handled by all involved handlers twice. I've checked the content of the request, and it's the same for both requests.

On reload `addFileServer` request handler is not called because cacheControl return 304 answer, but it's still getting called twice for some reason:

    remark42-dev | 2020/01/18 06:36:07.041 [DEBUG] {api/rest.go:583 api.cacheControl.func2.1} cacheControl request for /web/remark.js
    remark42-dev | 2020/01/18 06:36:07.041 [DEBUG] {api/rest.go:583 api.cacheControl.func2.1} cacheControl request for /web/remark.js

---

However, after a few application rebuilds, I see single request to every endpoint again and couldn't reproduce it. Continuing investigation.

paskal:
Three distinct requests to http:`//127.0.0.1:8080/web/` with Firefox.

### First load after force-reload

| URL                       | `addFileServer` handler call | `cacheControl` call      |
| ------------------------- | -------------------------- | ---------------------- |
[...]

PR #1592, opened 2023-02-05
clarify frontend files cache policy

paskal:
Previously Cache-Control was set to "max-age=3600, no-cache", but max-age is not used in case no-cache is set.

cc @Ivan-Chupin for review.

Addresses https://github.com/umputun/remark42/discussions/1549#discussioncomment-4750537.


akellbl4 (changes requested):
It overrides other parts of code which makes this change really confusing and hard to debug. https://github.com/umputun/remark42/issues/556#issuecomment-763946927

Issue #99, opened 2018-06-23
Separate frontend & backend


igoradamenko:
I think we should separate frontend & backend. Right now fronted is just a folder `web` inside a project, which looks a bit awkward. And we have a lot of go's files inside a project root, etc.

@umputun, what do you think about explicit separation? Like folders `frontend` and `backend` in project root? Is it possible for your part, or not?

umputun:
Yeah, planned to do this. We can start by renaming `app` to `backend` and `web` to `frontend`. On my side, this will need some refactoring since go kind of sensitive to the path, but not a big deal.

umputun:
the backend code was moved to `backend` top-level directory

umputun:
@igoradamenko - you can rename `web` as you ready

umputun:
@Reeywhaar - how hard is to rename `web` to `frontend`? Obviously not that critical, but will be nice to have backend and frontend directories

Reeywhaar:
I think it's possible

PR #1633, opened 2023-05-02
Add the ability to change the primary color of the comments widget and other styles

husky-dev:
First of all, thank you for this comments engine. I have added it to my blog, and it looks great. However, I have not found a way to change the primary color. The default green color is not the best choice for my website, and it would be better to use blue as the primary color for [https://radio-t.com/](https://radio-t.com/). Therefore, I have decided to add this functionality.

This PR adds the ability to change the primary color by using `remark_config`:

```js
var remark_config = {
  // ...other configs
  styling: {
    colors: {
      primary: {
        light: '#0aa',
        dark: '#099',
      },
    },
  },
};
```

Or change it dynamically:

```js
window.REMARK42.changeStyling({
  colors: {
    primary: '#099',
  },
})
```

Initially, I planned to use the `theme` config key for this feature, but it is already used for selecting a theme (light/dark). Although it is possible to use this key with backward compatibility, such as `light`, `dark`, or `{colors: {}}`. But it might overcomplicate the source code. Additionally, the `styles` key is used for iframe's custom styles. Therefore, I decided to use the `styling` key instead.

I saw some commented code which indicated that the initial plan was to add something like a `__colors__` key and allow the user to provide colors by name, such as `--color0`, `--color9`, and so on. However, it might be difficult for the user to change each color in this way. It would be much easier to provide one primary color and generate the required tones.

The config object data structure is designed to be easily extensible in the future, allowing for the addition of properties such as text color, background color, font configurations, and more. However, I have not implemented all of this functionality because I would like to hear your feedback first.

The code applies styles by changing appropriate CSS variables.

The code allows for the application of a single color for both themes (like: `primary: '#099'`) or the use of separate colors for each (`primary: {light: '...', dark: '...'}`). This may not be as useful for the primary color, but it can be very useful for the background color, which might be completely different in those cases.

I've added some utilities to make type checking easier (`app/utils/types.ts`), which I'm using in the type guards, as well as a utility for parsing and working with colors (`app/utils/colors.ts`). This is not difficult to implement, so I decided not to add external dependencies.

Please let me know your thoughts, as I am open to any constructive criticisms or suggestions you may have.

### DIFF
--- a/frontend/apps/remark42/app/common/settings.ts
+++ b/frontend/apps/remark42/app/common/settings.ts
@@ -1,5 +1,6 @@
 import { parseQuery } from 'utils/parse-query';
 import { THEMES, MAX_SHOWN_ROOT_COMMENTS } from './constants';
+import { themeStylingFromUrlSearchParams } from './theme';
 
 function parseNumber(value: unknown) {
   if (typeof value !== 'string') {
@@ -27,3 +28,4 @@ export const url = rawParams.url;
 export const token = rawParams.token;
 export const locale = rawParams.locale || 'en';
 export const noFooter = rawParams.no_footer === 'true';
+export const styling = themeStylingFromUrlSearchParams(rawParams);
--- a/frontend/apps/remark42/app/common/theme.spec.ts
+++ b/frontend/apps/remark42/app/common/theme.spec.ts
@@ -1,0 +1,65 @@
+import { themeStylingFromUrlSearchParams, themeStylingToUrlSearchParams } from './theme';
+
+describe('themeStylingToUrlSearchParams', () => {
+  test('returns an empty object when styling is undefined', () => {
+    expect(themeStylingToUrlSearchParams(undefined)).toEqual({});
+  });
+
+  test('converts primary color string to "colorsPrimary" parameter', () => {
+    const styling = { colors: { primary: '#ff0000' } };
+    const params = themeStylingToUrlSearchParams(styling);
+    expect(params).toEqual({ colorsPrimary: '#ff0000' });
+  });
+
+  test('converts primary color object with light/dark properties to "colorsPrimary", "colorsPrimaryLight", and "colorsPrimaryDark" parameters', () => {
+    const styling = { colors: { primary: { light: '#ffffff', dark: '#000000' } } };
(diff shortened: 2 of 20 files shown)


paskal:
I can't check the technical implementation, but the idea looks great and simple enough to be used by many users potentially.

@akellbl4 can you please take a look?

akellbl4:
@husky-dev hey, sorry I'm coming to late here but we have ability to pass css variables of colors in JS config. Have you seen that? https://github.com/umputun/remark42/blob/master/frontend/apps/remark42/templates/demo.ejs#L129 this is the thing, it's not documented but we are not gonna chage it if we don't go by any other styling route.

husky-dev:
@akellbl4 hey, yes, I saw an example with `__colors__`. My idea was to implement theme configuration similar to popular UI libraries, with `primary` and `secondary` colors, the ability to change fonts, and so on. That's why my proposed config looks like this:

```js
var remark_config = {
  // ...other configs
  styling: {
    colors: {
      primary: {
        light: '#0aa',
        dark: '#099',
      },
    },
    // ...fonts,
    // ...icons,
    // ...etc
  },
};
```

Here is the current implementation:

```js
var remark_config = {
  site_id: 'remark',
  host: '<%= htmlWebpackPlugin.options.REMARK_URL %>',
  url: window.location.href,
  components: ['embed', 'counter'],
  // __colors__: {
  //   "--color0": "red",
  // },
}
```

The idea was to set one base color, and other colors would be calculated from it. Right now, there's a bunch of different colors and their hues, and it's unclear which is responsible for what. Names like `color0`, `color1`, and `color2` don't help with that. I think users of the widget won't understand why the property is named `__colors__` with underscores, while other props aren't. It looks like a hack. What if you want to add font configuration because your website uses custom fonts? With the current code style, you'll probably have to add a `__fonts__` prop. I don't think it's a good idea to use separate config props when most libraries simply use a theme prop. Since theme is already taken, the next logical choice was styling. This way, you could consolidate all styling configurations under a single prop, like colors, fonts, icons, and so on.

But this PR has been open for almost 6 months, and I'm not sure that I want to keep working on it anymore. So feel free to close it.

Issue #1456, opened 2022-07-31
Lack of auth providers doesn't show errors in the UI


paskal:
Here is what one would see by running Remark42 like `./remark42.darwin-arm64 server --secret=12345 --url=http://127.0.0.1:8080 `:

<img width="345" alt="image" src="https://user-images.githubusercontent.com/712534/182046258-275c1490-6ed5-40ea-a20a-653d1eff09f5.png">

Let's name it "Read-only mode" and write some helping text like "No authentication methods are set up, the comments are in the read-only mode".

akellbl4:
Why would somebody run comment system without auth providers at all?
Should backend throw an error if no providers were configured?

paskal:
Frontend behavior currently is not helpful for someone trying to set up Remark42 for the first time, I would insist on showing some proper message about lack of providers in that case.

akellbl4:
It is not friendly because in my opinion Remark42 shouldn't be started without providers and backend should provide info about it in any possible form. Responsibility of backend configuration and errors shouldn't be shifted to frontend.

I have this position because we don't have requirement for starting an instance without providers and it makes no sense to me. Why would we need it?

umputun:
I think It does make sense to run the backend-only without any auth providers if basic auth mode turned on. This is useful at least for development and also for some internal tests/operation one may want to do without setting up the proper auth

umputun:
frontend has nothing to do with this. From the fronted POV such configuration is invalid (or at leas incomplete), but if user run it this way for whatever testing they did, and forgot to enable auth providers - frontend may help that poor user a little bit. Probably, logging WARN to frontend log is more apropriate place technically (i.e. "WARN no auth backend enabled", or smth like this) but this won't prevent users from doing it anyway and asking after this "why no auth working?"

akellbl4:
Backend-only mode probably means that FE shouldn't be exposed at all. I think it should has specific option and an error/warning won't appear in BE logs in this case.

paskal:
The backend doesn't fail in such a scenario as it's a valid case to run it in such a scenario. However, the front is broken in that case and doesn't help the user understand what is wrong. Some error message in case there are zero providers should be displayed.

umputun:
> Backend-only mode probably means that FE shouldn't be exposed at all. I think it should has specific option and an error/warning won't appear in BE logs in this case.

[...]

PR #1640, opened 2023-06-09
fix: ui error when no auth providers configured

Jaskon:
[This issue](https://github.com/umputun/remark42/issues/1456)

image

Please correct me if i'm wrong: view variable (telegram/token/email) is only exist after some action in the dropdown menu. So its not necessary to check here that view is equal to something and checks provided in my PR is enough


paskal (dismissed):

frontend/apps/remark42/app/components/auth/auth.messsages.ts
```diff
@@ -65,4 +65,8 @@ export const messages = defineMessages<string>({
     id: 'auth.signout',
     defaultMessage: 'Sign Out',
   },
+  noProviders: {
```
paskal, line 68: I think you need to add a translation string and regenerate translations in case you're adding new text like that.

paskal:
@Jaskon can you please finalise that PR so that we could merge it?

paskal:
Rebased onto current master to clear the conflict (the messages file was renamed from `auth.messsages.ts` to `auth.messages.ts` after this PR was opened). Added a small follow-up commit: corrected the message grammar ("May be" → "Maybe") and ran `translation:generate` to propagate the new `auth.no-providers` key into all locale catalogs. Verified locally — eslint, type-check, translations check, and `jest` (including the new test) all pass.

paskal:
Updated the follow-up commit: replaced the English placeholders with proper translations for all 23 non-English locales (the new `auth.no-providers` key), and added a test asserting the error is hidden when providers *are* configured (covers both branches of the new conditional). Local checks green: translations check, type-check, eslint, jest (30 tests).

Note on `codecov/project`: codecov reports "all modified and coverable lines are covered" (`codecov/patch` passes). The `-1.72%` on `codecov/project` is a stale-base artifact — the report is 8 commits behind head and shows Files −8 / Lines −148 / Partials +108 across unrelated code, not anything introduced here.

PR #1643, opened 2023-06-12
RSS: Fix snippet generation for parent comment

koteyur:
PR fixes the following issues with snippet generation for parent comments in RSS (couldn't find existing issues to tag them...):

1) Current implementation simply removes the last word, without truncating up to limit length.
2) In case if even the first word (e.g. magnet link or some base64) is too long don't add extra space.

STR:
1) Run dev version (https://remark42.com/docs/contributing/backend/)
2) Write very long comment (chatgpt helps :))
3) Reply to the comment
4) Check RSS

ER:
Text of original comment in reply item is truncated to 300 chars.

AR:
Only the last word is truncated.

### DIFF
--- a/backend/app/rest/api/rest_test.go
+++ b/backend/app/rest/api/rest_test.go
@@ -301,7 +301,7 @@ func TestRest_cacheControl(t *testing.T) {
 	for i, tt := range tbl {
 		tt := tt
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


paskal (dismissed):
LGTM, thank you!

koteyur:
It looks like new version of golangci-lint found some issues in existing code

paskal (approved):

umputun (approved):
LGTM, thx

Issue #1174, opened 2021-11-11
Easier email subscription for email users


paskal:
The user who logged in with the email has to send and paste a token from the email to confirm the replies subscription.

The desired flow is confirmation of subscription without sending the token if the email is the same user logged in with. To make it live, we need to add a backend check for the address and some special flow for instant confirmation, either by returning something else than token after the first step and requiring the second API call as well as we do now or by just doing the confirmation in the first API call.

paskal:
@akellbl4 passing to you, once you'll switch to params in the body for /email/subscribe, `{"updated": true}` in response would mean that the user is subscribed without the necessity to enter the token from email.

paskal:
@Jaskon would be great if you'll be able to look into it, the backend part is already implemented.

PR #1646, opened 2023-06-25
Skip confirmation step on email subscription

Jaskon:
When user is logged in with the same email he tries to subscribe
According to #1174


Jaskon:
#1645 changes are also here. They were required for this change and I was forced to merge that branch into this one.

paskal (dismissed):

paskal (commented):
Here is what I get after trying to subscribe with the same email once I've already subscribed (but interface didn't update for some reason):

image

The core functionality is fine, but I wonder if we can handle this code 409 edge case differently so that interface will just update with "you've subscribed" instead of showing me the error before I reload the page.

Jaskon:
@paskal Could you please review it again?
Added a handler for 409 http error.

paskal (dismissed):

paskal (approved):

PR #1657, opened 2023-07-18
Update documentation to support Caddy V2

tardisx:
* Change Caddy example config to be compatible with Caddy V2 config format 
* Caddy V2 was released in 2020 and most new users are likely to be using that
* Existing config snippet is not compatible with V2 at all

### DIFF
--- a/site/src/docs/manuals/subdomain/index.md
+++ b/site/src/docs/manuals/subdomain/index.md
@@ -58,17 +58,15 @@ The `nginx.conf` would then look something like:
 
 ### Caddy configuration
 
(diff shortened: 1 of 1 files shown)


umputun (commented):
thx. Maybe we better keep the old one under "caddy version 1" and add your change with "caddy version 2" ?

Proxy is such a thing people may use for a long time without upgrading

tardisx:
> thx. Maybe we better keep the old one under "caddy version 1" and add your change with "caddy version 2" ?
> 
> Proxy is such a thing people may use for a long time without upgrading

Excellent point, I have updated the PR.

umputun (approved):
thx

Issue #1673, opened 2023-09-16
Lack of frontend information about relative URL


paskal:
Previewing a comment with the relative URL:

<img width="693" alt="image" src="https://github.com/umputun/remark42/assets/712534/0de900b1-9802-4ead-a576-8420ce118b6b">

The answer from the backend is `400`:

```json
{
    "code": 4,
    "details": "invalid comment",
    "error": "links should start with mailto:, http:// or https://"
}
```

Submitting a comment with the relative URL:

<img width="657" alt="image" src="https://github.com/umputun/remark42/assets/712534/fefd06d4-f8b4-45e1-a738-384f51a5d6b8">

The answer from the backend is `400` and the same text as above.

We need to alter the frontend to show the problem to the user instead of showing generic error messages, which do not help determine what went wrong and on which side. The preview answer for an unknown `400` error should be as well altered to `Invalid comment data.` so that the user would know something is off with the comment and not the Remark42 server.

thevinitgupta:
Is it open? I would love to work on this.

paskal:
Yes, feel free to pick it up.

thevinitgupta:
When I am testing the above, I am getting the same text displayed as the second image.
So is the issue solved??

PR #1674, opened 2023-09-17
Extract error message from API response for preview.

sharief007:
### Change Log
- For preview errors, only `isErrorShown` is set to `true` and `errorMessage` is not provided. Hence the comment form shows `Something went wrong. Please try again a bit later.` (generic message) by default for all preview errors.
- Refactored the exception handling to derive error message from the response code (similar to `send` method). Now both Send and Preview shows the same error message.

Resolves: https://github.com/umputun/remark42/issues/1673


akellbl4 (approved):
@sharief007 thank you!

paskal:
Thanks a lot! I've checked now and it works as expected.

Issue #1677, opened 2023-10-10
[CWE-918] Blind SSRF in /api/v1/comment


jet-pentest:
Hi! I found an issue in remark42 engine. 
When field "title" is missing in comment request and attacker send a malicious URL which is a part of locator struct on POST request to ```/api/v1/comment?site=<SITE_ID>```,  it leads to Blind SSRF vulnerability.
As example, malicious URL looks like this:
{"text":"qweqwe","locator":{"site":"remark","url":"https://j5pxshabxb5037lms6z182pkjbp4d01p.oastify.com"}}
This malicious JSON in request to ```/api/v1/comment?site=<SITE_ID>``` leads to Blind SSRF
<img width="435" alt="Снимок экрана 2023-10-10 в 11 59 07" src="https://github.com/umputun/remark42/assets/71512502/f10ce363-e968-4a6d-a0e9-dd60c930caab">
As another way, malicious field "title", which should be fully cutted by sanitizer, also leads to blind SSRF.

paskal:
Thank you for the report!

> As another way, malicious field "title", which should be fully cutted by sanitizer, also leads to blind SSRF.

Could you please clarify this point? Title is not sanitized now but I don't understand how does it lead to blind SSRF.

jet-pentest:
> Could you please clarify this point? Title ~is not sanitised now~ is sanitised except for some HTML tags like `<b>` and `<a>` but I don't understand how does it lead to blind SSRF.

@paskal Hi! In this case, a minimal payload that I was able to reproduce was: `<script>alert();</script>`. An sanitizer will fully clean "title" field.

jet-pentest:
Also, in my example, I was using a Burp Suite Collaborator as a part of Burp Suite. For development purposes, you may try to use it too. Another pay-free variant is http://webhook.site/. I hope it'll be helpful for you.

paskal:
What's the problem with `<script>alert();</script>` being converted to a empty title? It's not a vulnerability for sure.

jet-pentest:
> What's the problem with `<script>alert();</script>` being converted to a empty title? It's not a vulnerability for sure.
@paskal 
Sure, but SSRF with non empty title was not working. I was highlighted this, because not only removing `title` triggers SSRF. It's will be sanitizer's result of work. Also, it doesn't work, if you'll send empty `title` field due initial checks.

jet-pentest:
Thanks for fast resolve! You're the best:))

PR #1680, opened 2023-10-10
Remove all HTML tags from comment title and username

paskal:
Previously, we stripped unsafe HTML tags but left some like `<a>`, but it's not expected to have a link in a title or user's name, so the new behaviour is stripping everything.

The issue was highlighted in #1677.


umputun:
LGTM

PR #1681, opened 2023-10-10
Limit TitleExtractor to allow only Remark42 whitelisted domains

### LINKED ISSUE #1677: [CWE-918] Blind SSRF in /api/v1/comment

paskal:
Allowed domains consist of `REMARK_URL` second-level domain (or whole IP in case it's IP like `127.0.0.1`) and `ALLOWED_HOSTS`. That is needed to prevent Remark42 from asking arbitrary servers and storing the page title as the comment.PostTitle.

Previous behaviour allowed the caller of the API to create a comment with an arbitrary URL and learn the title of the page, which might be accessible to the server Remark42 is installed on but not to the user outside that network (CWE-918).

Resolves #1677.

### DIFF
--- a/backend/app/cmd/server_test.go
+++ b/backend/app/cmd/server_test.go
@@ -726,6 +726,27 @@ func Test_splitAtCommas(t *testing.T) {
 	}
 }
 
+func Test_getAllowedDomains(t *testing.T) {
+	tbl := []struct {
+		s              ServerCommand
+		allowedDomains []string
+	}{
+		// correct example, parsed and returned as allowed domain
+		{ServerCommand{AllowedHosts: []string{}, CommonOpts: CommonOpts{RemarkURL: "https://remark42.example.org"}}, []string{"example.org"}},
+		{ServerCommand{AllowedHosts: []string{}, CommonOpts: CommonOpts{RemarkURL: "http://localhost"}}, []string{"localhost"}},
+		// incorrect URLs, so Hostname is empty but returned list doesn't include empty string as it would allow any domain
+		{ServerCommand{AllowedHosts: []string{}, CommonOpts: CommonOpts{RemarkURL: "bad hostname"}}, []string{}},
+		{ServerCommand{AllowedHosts: []string{}, CommonOpts: CommonOpts{RemarkURL: "not_a_hostname"}}, []string{}},
+		// test removal of 'self', multiple AllowedHosts and deduplication
+		{ServerCommand{AllowedHosts: []string{"'self'", "example.org", "test.example.org", "remark42.com"}, CommonOpts: CommonOpts{RemarkURL: "https://example.org"}}, []string{"example.org", "remark42.com"}},
+	}
+	for i, tt := range tbl {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			assert.Equal(t, tt.allowedDomains, tt.s.getAllowedDomains())
+		})
+	}
+}
+
 func chooseRandomUnusedPort() (port int) {
 	for i := 0; i < 10; i++ {
 		port = 40000 + int(rand.Int31n(10000))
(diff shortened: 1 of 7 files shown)


backend/app/cmd/server_test.go
```diff
@@ -726,6 +726,27 @@ func Test_splitAtCommas(t *testing.T) {
 	}
 }
 
+func Test_getAllowedDomains(t *testing.T) {
+	tbl := []struct {
+		s              ServerCommand
+		allowedDomains []string
+	}{
+		// correct example, parsed and returned as allowed domain
+		{ServerCommand{AllowedHosts: []string{}, CommonOpts: CommonOpts{RemarkURL: "https://remark42.example.org"}}, []string{"example.org"}},
```
umputun, line 735: would be nice to add http:// case just ... in case

backend/app/store/service/title.go
```diff
@@ -19,14 +20,16 @@ const (
...
 }
 
 // NewTitleExtractor makes extractor with cache. If memory cache failed, switching to no-cache
-func NewTitleExtractor(client http.Client) *TitleExtractor {
+func NewTitleExtractor(client http.Client, allowedDomains []string) *TitleExtractor {
 	res := TitleExtractor{
-		client: client,
+		client:         client,
+		allowedDomains: allowedDomains,
 	}
```
umputun, line 33: maybe add log debug here, to know what domain were allowed?

backend/app/store/service/title.go
```diff
@@ -38,33 +41,48 @@ func NewTitleExtractor(client http.Client) *TitleExtractor {
...
-func (t *TitleExtractor) Get(url string) (string, error) {
+func (t *TitleExtractor) Get(pageURL string) (string, error) {
+	// parse domain of the URL and check if it's in the allowed list
+	u, err := url.Parse(pageURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse url %s: %w", pageURL, err)
+	}
+	allowed := false
+	for _, domain := range t.allowedDomains {
+		if strings.HasSuffix(u.Hostname(), domain) {
```
umputun, line 52: isn't it risky a little bit? I mean can it allow ab.example.com for allowed b.example.com?

umputun (commented):
solid job; just a few minor things

backend/app/store/service/title.go
paskal: It's a bit risky, however I see no way around for allowing case like example.org when remark42.example.com is used for Remark42 installation for comments on example.com. I think the solution here is outlining that risk in the documentation.

Also, the current case allowed requests to notexample.org if my domain is example.org (e.g. suffixes match): I've fixed that.

paskal: Any other behaviour would be a breaking change. It's a small part of functionality so maybe a requirement for listing your domain(s) explicitly in AllowedDomains in case Remark42 runs on a separate domain is not bad.

I can alter the logic to get second-level domain only from `REMARK_URL` and use `ALLOWED_HOSTS` as-is, that seem to make sense.

umputun:
LGTM

PR #1685, opened 2023-10-18
Collect /find Info for tree and plain types consistently

paskal:
MakeTree calculated Info locally for historical reasons, and the results were consistent with the dataService.Info call but calculated differently. Requesting plain format comments, on the other hand, had a problem with not returning the read-only status of the page if it doesn't have any comments yet.

That change fixes that, ensuring that Info is requested in the same manner and the read-only status is returned correctly in case the post doesn't have comments but has such a status.


backend/app/store/service/tree.go
```diff
@@ -106,7 +84,8 @@ func (t *Tree) proc(comments []store.Comment, node *Node, rd *recurData, parentI
 	sort.Slice(node.Replies, func(i, j int) bool {
 		return node.Replies[i].Comment.Timestamp.Before(node.Replies[j].Comment.Timestamp)
 	})
-	return node, rd.tsModified, rd.tsCreated
+	node.tsModified, node.tsCreated = rd.tsModified, rd.tsCreated
```
paskal, line 87: This line is magic, I have no idea why it doesn't mess up tsCreated or tsModified of any of the nodes. Maybe we just don't have good enough test cases to catch it.

umputun (approved):
makes sense

PR #1688, opened 2023-10-28
update node to v20 lts

akellbl4:
- updates node to v20
- enable use of pnpm v8
- replaces nvmrc with npmrc ([pnpm manages node version for you](https://pnpm.io/npmrc#use-node-version))
- moves env requirements to the root `package.json`


frontend/packages/api/lib/fetcher.ts
```diff
@@ -79,6 +79,7 @@ export const createFetcher = (site: string, baseUrl: string): Client => {
 				.catch(Object)
 				.then((data: string) => {
 					if (res.status < 200 || res.status > 299) {
+						console.log(data)
```
Mavrin, line 82: please remove console.log

frontend/packages/api/tests/clients/admin.test.ts
```diff
@@ -6,7 +6,7 @@ interface Context {
 	client: ReturnType<typeof createAdminClient>
 }
 
-describe<Context>('Admin Client', (adminClient) => {
+describe.skip<Context>('Admin Client', (adminClient) => {
```
Mavrin, line 9: why do you skip this test

frontend/e2e/screenshot-firefox.png
Mavrin, line 1: why does firefox show only one comment? I see that other browsers render 5 comments.

Mavrin (changes requested):

frontend/e2e/screenshot-firefox.png
akellbl4: All of the tests started to fail after update, so I skip them. This code is not in use so I wanted to fix it as follow up

frontend/packages/api/tests/clients/admin.test.ts
akellbl4: reworked the whole thing, tests are fixed.

frontend/e2e/screenshot-firefox.png
akellbl4: Ops, those shouldn't get into the repo.
Those were created while local testing and FF had some problems.

frontend/packages/api/package.json
```diff
@@ -1,57 +1,68 @@
...
   "author": "Paul Mineev",
   "license": "MIT",
+  "type": "module",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.js",
+      "types": "./dist/index.d.ts"
```
Mavrin, line 21: https://nodejs.org/api/packages.html#community-conditions-definitions
> "types" - can be used by typing systems to resolve the typing file for the given export. This condition should always be included first.

frontend/packages/api/src/lib/fetcher.ts
```diff
@@ -90,5 +87,23 @@ export const createFetcher = (site: string, baseUrl: string): Client => {
...
+
+function getSearchParams(query?: QueryParams) {
+	if (!query) {
+		return ''
+	}
+
+	// overrides type of init in URLSearchParams constructor because it's not correct and accepts QueryParams
+	const searchParams = new URLSearchParams(
+		Object.entries(query).reduce<Record<string, string>>(
+			(acc, [k, v]) => Object.assign(acc, { [k]: v.toString() }),
```
Mavrin, line 101: I think we don't need to create new object on every iterations and we can mutate `acc`
```
acc[k] = v.toString();
```

Mavrin (changes requested):

Issue #782, opened 2020-10-02
Paginate/limit comments from /find endpoint


CodeWithOz:
Is there a built-in way to paginate when fetching comments from the /find endpoint? For instance if a particular thread has 1000 replies and I only want to get them in batches of 50 each time. Do I need my own custom code in the backend to implement this?

I can see the [since](https://github.com/umputun/remark42/blob/c02590665f3a5d9d3d598bb70168ede290d4a9ff/backend/app/rest/api/rest_public.go#L64) param but ideally I'd like to also specify a maximum number of replies to include in the response.

Thanks!

umputun:
As of now `since`is the only way to limit the response.

ingria:
Would've been a great feature. I have some threads with over 10k comments, and there's no way to load them page by page or with infinite loader. Loading everything at once is not an option, because of a) large amount of traffic b) heavy cpu load while parsing JSON.

Example implementation: [Disqus](https://disqus.com/embed/comments/?base=default&f=disqus&t_i=8465986837&t_u=https%3A%2F%2Fblog.disqus.com%2Fintroducing-collapsed-replies&t_d=%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20Introducing%20Collapsed%20Replies!%0A%20%20%20%20%20%20%20%20%20%20%20%20&t_t=%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20Introducing%20Collapsed%20Replies!%0A%20%20%20%20%20%20%20%20%20%20%20%20&s_o=default#version=4c17dbce49f331e55ae0e3654575eb6a)

paskal:
I'll look into it at some point, as it seems like a must-have for bigger installations.

paskal:
I just wanted to let you know that I plan to work on this issue shortly. Current endpoint:

```go
// GET /find?site=siteID&url=post-url&format=[tree|plain]&sort=[+/-time|+/-score|+/-controversy]&view=[user|all]&since=unix_ts_msec&limit=100&limit_start={id}
// find comments for the given post. Returns in tree or plain formats, sorted
```

What I plan to add:

```go
// GET /find?<...>&limit=100&limit_start={id}
// When `limit` is set, first {limit} comments are returned. When `limit_start` is set, comments are returned starting
// from the comment with the given id.
```

[...]

PR #1699, opened 2023-11-18
add pagination to GET /api/v1/find endpoint

paskal:
`format=tree` pagination provides top-level comments with all replies and returns the last top-level comment as `last_comment` to be used as `offset` for the next page. If comments and replies overflow the limit, the one stepping out of the limit will not be returned. If the first comment and its replies after the given offset overflow the limit, it will be returned with all the replies.

`format=plain` pagination works by providing all comments and returning the last comment as `last_comment` to be used as `offset` for the next page.

Requires #1685 to be merged first. Backend part for #782.

### DIFF
--- a/backend/app/rest/api/rest_public.go
+++ b/backend/app/rest/api/rest_public.go
@@ -16,6 +16,7 @@ import (
 	cache "github.com/go-pkgz/lcw/v2"
 	log "github.com/go-pkgz/lgr"
 	R "github.com/go-pkgz/rest"
+	"github.com/google/uuid"
 	"github.com/skip2/go-qrcode"
 
 	"github.com/umputun/remark42/backend/app/rest"
@@ -48,10 +49,18 @@ type pubStore interface {
 	Counts(siteID string, postIDs []string) ([]store.PostInfo, error)
 }
 
-// GET /find?site=siteID&url=post-url&format=[tree|plain]&sort=[+/-time|+/-score|+/-controversy]&view=[user|all]&since=unix_ts_msec
-// find comments for given post. Returns in tree or plain formats, sorted
+// GET /find?site=siteID&url=post-url&format=[tree|plain]&sort=[+/-time|+/-score|+/-controversy]&view=[user|all]&since=unix_ts_msec&limit=100&offset_id={id}
+// find comments for given post. Returns in tree or plain formats, sorted.
 //
 // When `url` parameter is not set (e.g. request is for site-wide comments), does not return deleted comments.
+//
+// When `limit` is set, first {limit} comments are returned. When `offset_id` is set, comments are returned starting
+// after the comment with the given id.
+// format="tree" limits comments by top-level comments and all their replies,
+// and never returns parent comment with only part of replies.
+//
+// `count` in the response refers to total number of non-deleted comments,
+// `count_left` to amount of comments left to be returned _including deleted_.
 func (s *public) findCommentsCtrl(w http.ResponseWriter, r *http.Request) {
 	locator := store.Locator{SiteID: r.URL.Query().Get("site"), URL: r.URL.Query().Get("url")}
 	sort := r.URL.Query().Get("sort")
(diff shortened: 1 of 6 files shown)


backend/app/rest/api/rest_public.go
```diff
@@ -70,7 +79,24 @@ func (s *public) findCommentsCtrl(w http.ResponseWriter, r *http.Request) {
...
+
+	offsetID := r.URL.Query().Get("offset_id")
+	if offsetID != "" {
+		if _, err = uuid.Parse(offsetID); err != nil {
+			rest.SendErrorJSON(w, r, http.StatusBadRequest, err, "bad offset_id value", rest.ErrCommentNotFound)
+			return
+		}
+	}
+
+	log.Printf("[DEBUG] get comments for %+v, sort %s, format %s, since %v, limit %d, offset %s", locator, sort, format, since, limit, offsetID)
```
umputun, line 99: I think %q for the offset will be less confusing in the log if it was input.

backend/app/rest/api/rest_public.go
```diff
@@ -70,7 +79,24 @@ func (s *public) findCommentsCtrl(w http.ResponseWriter, r *http.Request) {
 		since = time.Time{} // since doesn't make sense for tree
 	}
 
-	log.Printf("[DEBUG] get comments for %+v, sort %s, format %s, since %v", locator, sort, format, since)
+	limitParam := r.URL.Query().Get("limit")
+	var limit int
```
umputun, line 83: nit: personally, I'd prefer limit:=0 instead, as it communicates the default value more clearly.

backend/app/rest/api/rest_public.go
```diff
@@ -432,3 +466,25 @@ func (s *public) parseSince(r *http.Request) (time.Time, error) {
...
+
+// limitComments returns limited list of comments and count of comments left after limit.
+// If offsetID is provided, the list will be sliced starting from the comment with this ID.
+// If offsetID is not found, the full list will be returned.
+// It's used for only "
+func limitComments(c []store.Comment, limit int, offsetID string) (comments []store.Comment, countLeft int) {
+	if offsetID != "" {
+		for i, comment := range c {
+			if comment.ID == offsetID {
+				c = c[i+1:]
```
umputun, line 478: This is confusing and even scary a bit. Generally, the idea of changing the slice we are ranging over inside the loop feels wrong to me. Probably it is safe to do now, as we are breaking right away, but someday, as we add code to this function, it may not be the case. I'd rather add a new lcs variable for the result slice.

backend/app/store/service/tree.go
```diff
@@ -145,3 +159,81 @@ func (t *Tree) sortNodes(sortType string) {
...
+// starting with comment next after offsetID.
+//
+// If offsetID is empty or invalid, it starts from the beginning. If limit is 0, it doesn't limit anything.
+//
+// Limit is applied to top-level comments only, so top-level comments only returned with all replies,
+// and lastLimitedComment is set to the last top-level comment and not last reply in it.
+//
+// In case limit is less than the number of replies to first comment after given offset, that first comment is
+// returned completely with all replies.
+func (t *Tree) limit(limit int, offsetID string) {
```
umputun, line 173: I would suggest calling it setLimit or updateWithLimit, as this function modifies nodes and does not return anything.

backend/app/store/service/tree.go
```diff
@@ -145,3 +159,81 @@ func (t *Tree) sortNodes(sortType string) {
 		}
 	})
 }
+
+// limit limits number of comments in tree and sets countLeft and lastLimitedComment,
```
umputun, line 163: I think there is no direct test for this non-trivial function, but rather only via MakeTree? It's up to you, but I will feel better if we have some.

umputun (changes requested):
few minor things. I have not checked the tree part of this, and asked for some tests to make sure everyting is working as designed.

PR #1700, opened 2023-11-18
Allow disabling fancy HTML formatting

### LINKED ISSUE #1666: Double-quote characters in comments show as guillemets in output
When double-quote characters are entered in a comment "like this", they appear in the output as guillemets  «like this».

This is presumably intended, but it is not appropriate for some languages, and doesn't seem to be configurable.

paskal:
It might be necessary if the comments should preserve original quotes instead of replacing them with angled ones.

Resolves #1666.


akellbl4 (dismissed):
LGTM! One thing about the name of the param. SmartyPants is about punctuation "enhancement" and it has nothing to do with HTML itself.

umputun (approved):
LGTM

PR #1702, opened 2023-11-18
Cleanup images from deleted comments

### LINKED ISSUE #1698: Image are not deleted once admin deletes a comment with it
I discovered that deleting a comment does not delete the image file attached to the comment on the server. I have quite a bit of space on the server to store dead information. It would be logical to delete the image file along with the comment.

paskal:
Previously, images were deleted only from comments deleted before EditDuration expiration. After this change, any deletion of the comment deletes images if they are not used elsewhere in comments under the same page.

FS store got a mutex lock to prevent missing the file in the transition from staging to permanent storage, which otherwise would result in the file being stuck in the permanent storage forever.

Resolves #1698.

### DIFF
--- a/backend/app/store/image/bolt_store.go
+++ b/backend/app/store/image/bolt_store.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"time"
 
@@ -116,6 +117,18 @@ func (b *Bolt) Load(id string) ([]byte, error) {
 		return nil, err
 	}
 	return data, nil
+}
+
+// Delete image from storage
+func (b *Bolt) Delete(id string) error {
+	return b.db.Update(func(tx *bolt.Tx) error {
+		// deleting a non-existing key doesn't return an error, so joining errors from deleting an image
+		// from both buckets is safe and will return nil if there are no errors on the real delete
+		// or image is absent in both buckets
+		err := tx.Bucket([]byte(imagesBktName)).Delete([]byte(id))
+		err = errors.Join(err, tx.Bucket([]byte(imagesStagedBktName)).Delete([]byte(id)))
+		return err
+	})
 }
 
 // Cleanup runs scan of staging and removes old data based on ttl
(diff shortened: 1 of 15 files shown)


backend/app/store/image/bolt_store.go
```diff
@@ -118,6 +119,16 @@ func (b *Bolt) Load(id string) ([]byte, error) {
 	return data, nil
 }
 
+// Delete image from storage
+func (b *Bolt) Delete(id string) error {
+	return b.db.Update(func(tx *bolt.Tx) error {
+		// deleting non-existing key doesn't return an error
+		err := tx.Bucket([]byte(imagesBktName)).Delete([]byte(id))
```
umputun, line 126: combo of this code and the comment above is confusing. If deleting non-existing key doesn't return an error, why we bother to assign it to err? If it is needed, why we ignore (overwrite) the first error?

backend/app/store/service/service.go
```diff
@@ -796,6 +797,34 @@ func (s *DataStore) Delete(locator store.Locator, commentID string, mode store.D
 		s.repliesCache.Delete(commentID)
 		s.repliesCache.Delete(comment.ParentID)
 	}
+
+	// delete images from the comment if they are not reused elsewhere in comments to the same page
```
umputun, line 801: could be my misunderstanding but why we even care about the "if they are not reused elsewhere" part. Isn't is is safe to assume the image is unique to the particular comment and not bother with all those checks? You likely had some use-case (potential collision) in mind, but I fail to see it

umputun (commented):
a few things I don't get. Probably it is just me, but anyway

backend/app/store/image/bolt_store.go
paskal: I'll adjust it to say:
```
		// deleting a non-existing key doesn't return an error, so joining errors from deleting an image
		// from both buckets is safe and will return nil if there are no errors on the real delete
		// or image is absent in both buckets
```
My bad for not making the intention clear first.

backend/app/store/service/service.go
paskal: From the top of my mind, I can quote your message, and then the same image would be used in two places. Deleting the quoting message is not supposed to delete the original comment message.

umputun (approved):

Issue #1629, opened 2023-04-26
Long one-line `code` block causes content to be hidden


cowsay1:
reproduce in https://remark42.com/demo/
![image](https://user-images.githubusercontent.com/60675651/234499881-5e76b841-6d1c-4b0c-8a32-2c1adf8d4dc9.png)

I think it can be fixed with css white-space:pre-line or pre-wrap, not sure what is better:

![image](https://user-images.githubusercontent.com/60675651/234500353-282dd034-9ad8-4176-9578-d89a4d44587b.png)

paskal:
I've checked, and it's indeed a problem: that test results in a long line of text which is not visible and not scrollable.
@Jaskon @goooseman, maybe one of you can tackle this one?

PR #1704, opened 2023-11-25
Comment content styles

akellbl4:
- Resolves https://github.com/umputun/remark42/issues/1629
- Composes dark and light theme styles in one place
- Moves syntax styles into content styles file
- Adds colors to block with code only when `pre > code` (the way chroma generates html)
- Fixes size of text in code block. Previously `pre` had size 85% and `code` had 85% from it.
- Fixes code syntax highlight colors in dark mode (previously colors from light theme were used)
- Minor syntax highlight colors adjustmests in dark mode for better contrast

### Before
Screenshot 2023-11-25 at 3 16 47 AM

### After
Screenshot 2023-11-25 at 3 15 51 AM


frontend/apps/remark42/app/components/raw-content/raw-content.css
```diff
@@ -1,129 +0,0 @@
-.raw-content {
-  & .root_theme_dark {
```
akellbl4, line 2: Wrong nesting, should've been `.root_theme_dark &`

frontend/apps/remark42/app/components/comment-form/comment-form.tsx
```diff
@@ -515,9 +515,7 @@ export class CommentForm extends Component<Props, State> {
           !!preview && (
             <div className="comment-form__preview-wrapper">
               <div
-                className={b('comment-form__preview', {
-                  mix: b('raw-content', {}, { theme }),
-                })}
+                className="comment-form__preview raw-content"
```
akellbl4, line 518: theme set to root tag as `dark` so we don't need additional theming tag

Mavrin (approved):
👍

PR #1723, opened 2024-01-08
fix: Add missing Korean and Japanese translations for updated English text

NavyStack:
Korean translation is provided "as is".
---
* Rearranged the word order to better match Korean sentiment.
* Made adjustments to clarify subtle semantic differences in the original text, which may have identical translations in Korean.
* We have striven to maintain the original sentence structure as closely as possible.
* If the original text lacks a period, the translation into Korean also omits it, rendering the content as a short phrase rather than a full sentence.


NavyStack:
Japanese translation also provided "as is".
---
* Add missing Japanese translation.

paskal (approved):

PR #1734, opened 2024-01-28
Fix lack of error on file site export, improve lack of ADMIN_PASSWD error message

paskal:
Previously, status 200 was set for file export, which is used for backup, which resulted in an inability to set an error status code in case of a problem with file generation.

After this change, status code 200 would be written automatically by Go before we start writing the response's body.

Also, add meaningful error for lack of auth on import, remap and backup.

Previously, the error printed was just the following:

    error response "401 Unauthorized", Unauthorized"

New error:

    error response "401 Unauthorized", ensure you have set ADMIN_PASSWD and provided it to the command you're running: Unauthorized

### DIFF
--- a/backend/app/rest/api/rss.go
+++ b/backend/app/rest/api/rss.go
@@ -56,7 +56,6 @@ func (s *rss) postCommentsCtrl(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/xml; charset=utf-8")
 	w.WriteHeader(http.StatusOK)
-
 	if _, err = w.Write(data); err != nil {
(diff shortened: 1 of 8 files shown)


backend/app/rest/api/rss.go
```diff
@@ -56,7 +56,6 @@ func (s *rss) postCommentsCtrl(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/xml; charset=utf-8")
 	w.WriteHeader(http.StatusOK)
-
```
paskal, line 59: This is just for consistency. I've checked all `WriteHeader(http.StatusOK)` for the bug's presence, and that is the only place in the `rss.go` file with that extra newline, while two others don't.

umputun (approved):

PR #1742, opened 2024-02-20
Update to lcw v2 with generic types

paskal:
Now the types cast would be explicit thanks to generics.


backend/app/cmd/server.go
```diff
@@ -1341,5 +1343,5 @@ func (c *authRefreshCache) Get(key interface{}) (interface{}, bool) {
 
 // Set implements cache setter with key converted to string
 func (c *authRefreshCache) Set(key, value interface{}) {
-	_, _ = c.LoadingCache.Get(key.(string), func() (interface{}, error) { return value, nil })
+	_, _ = c.LoadingCache.Get(key.(string), func() (string, error) { return value.(string), nil })
```
umputun, line 1346: Do we need value interface{} here for real, or just a leftover ? the underlying cache is typed string now

backend/app/rest/api/admin_test.go
```diff
@@ -348,7 +348,7 @@ func TestAdmin_Block(t *testing.T) {
 	assert.Equal(t, "test test #1", comments.Comments[2].Text, "comment not removed and not cleared")
 	assert.False(t, comments.Comments[2].Deleted, "not deleted")
 
-	srv.pubRest.cache = cache.NewScache(cache.NewNopCache()) // TODO: with lru cache it won't be refreshed and invalidated for long
+	srv.pubRest.cache = cache.NewScache[[]byte](cache.NewNopCache[[]byte]()) // TODO: with lru cache it won't be refreshed and invalidated for long
```
umputun, line 351: I would expect it to infer the type for `NewNopCache`

umputun (commented):
a couple of minor things

backend/app/cmd/server.go
paskal: We do because go-pkg/auth, as of now, doesn't support generics and expects `interface{}`.

backend/app/rest/api/admin_test.go
paskal: Apparently it doesn't work:
```
Cannot infer V
Cannot use 'cache.NewScache(cache.NewNopCache[[]byte]())' (type *Scache[V]) as the type LoadingCache Type does not implement 'LoadingCache' need the method: Get(key lcw.Key, fn func() ([]byte, error)) (data []byte, err error) have the method: Get(key Key, fn func() (V, error)) (data V, err error)
```

PR #1763, opened 2024-05-02
Probe /auth/status from frontend to avoid 401 on /user (closes #1188)

### LINKED ISSUE #1188: Page Experience/Core Web Vitals: Browser errors were logged to the console
Regarding the old issue #328 there is a problem that if the user is not logged in API responds with a 401 error.

Google uses a lot of metrics to range sites. And from November 2021, not only Page Speed matters, but also user experience. More info https://developers.google.com/search/blog/2020/11/timing-for-page-experience

Using https://web.dev you can check the issues you have on the page. That's the new google page speed.
Logging errors to the console is one of the official issues. 

Example for page https://radio-t.com/p/2021/11/30/prep-783/:

Screenshot 2021-12-07 at 11 44 56

p.s. Also, some other console/SSL errors found on this site relates to not using GTM Server Side. Some of them you can fix with the correct policy headers, but not all of them.

paskal:
Resolves #1188.

`GET /api/v1/user` requires auth and returns `401` for anonymous visitors, which the browser logs to console even when JS catches it. The frontend now probes `GET /auth/status` first (always `200`, no console noise) and only fetches `/user` when the response says `"logged in"`.

When `/auth/status` reports `"not logged in"`, stale `JWT` and `XSRF-TOKEN` cookies are cleared in `getUser()` to preserve the cleanup-on-probe behaviour previously triggered by the `/user` 401.

No backend changes — `/user` keeps its current contract. Branch is `+81 / -13` across 4 frontend files (api client + tests, both `apps/remark42` and `packages/api`).

Continuation of accidentally closed #1752.

### DIFF
--- a/frontend/apps/remark42/app/common/api.test.ts
+++ b/frontend/apps/remark42/app/common/api.test.ts
@@ -1,5 +1,40 @@
-import { getUserComments } from './api';
-import { apiFetcher } from './fetcher';
+import { getUser, getUserComments } from './api';
+import { apiFetcher, authFetcher, JWT_COOKIE_NAME, XSRF_COOKIE } from './fetcher';
+import * as cookies from './cookies';
+
+describe('getUser', () => {
+  beforeEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('should fetch /user when /auth/status reports logged in', async () => {
+    const user = { id: '1', name: 'user' };
+    jest.spyOn(authFetcher, 'get').mockResolvedValue({ status: 'logged in', user: 'user' });
+    const apiSpy = jest.spyOn(apiFetcher, 'get').mockResolvedValue(user);
+
+    await expect(getUser()).resolves.toEqual(user);
+    expect(apiSpy).toHaveBeenCalledWith('/user');
+  });
+
+  it('should return null and clear auth cookies when /auth/status reports not logged in', async () => {
+    jest.spyOn(authFetcher, 'get').mockResolvedValue({ status: 'not logged in' });
+    const apiSpy = jest.spyOn(apiFetcher, 'get');
+    const clearSpy = jest.spyOn(cookies, 'clearAuthCookie').mockImplementation(() => {});
+
+    await expect(getUser()).resolves.toBeNull();
+    expect(apiSpy).not.toHaveBeenCalled();
+    expect(clearSpy).toHaveBeenCalledWith(JWT_COOKIE_NAME);
+    expect(clearSpy).toHaveBeenCalledWith(XSRF_COOKIE);
+  });
+
+  it('should return null when /auth/status request fails', async () => {
+    jest.spyOn(authFetcher, 'get').mockRejectedValue(new Error('boom'));
+    const apiSpy = jest.spyOn(apiFetcher, 'get');
+
+    await expect(getUser()).resolves.toBeNull();
+    expect(apiSpy).not.toHaveBeenCalled();
+  });
+});
 
 describe('getUserComments', () => {
   it('should call apiFetcher.get with /comments endpoint and default skip and limit query params', () => {
--- a/frontend/apps/remark42/app/common/api.test.ts
+++ b/frontend/apps/remark42/app/common/api.test.ts
@@ -27,12 +27,14 @@ describe('getUser', () => {
     expect(clearSpy).toHaveBeenCalledWith(XSRF_COOKIE);
   });
 
-  it('should return null when /auth/status request fails', async () => {
+  it('should return null without clearing cookies when /auth/status request fails', async () => {
     jest.spyOn(authFetcher, 'get').mockRejectedValue(new Error('boom'));
     const apiSpy = jest.spyOn(apiFetcher, 'get');
+    const clearSpy = jest.spyOn(cookies, 'clearAuthCookie').mockImplementation(() => {});
 
     await expect(getUser()).resolves.toBeNull();
     expect(apiSpy).not.toHaveBeenCalled();
+    expect(clearSpy).not.toHaveBeenCalled();
   });
 });
 
--- a/frontend/apps/remark42/app/common/api.ts
+++ b/frontend/apps/remark42/app/common/api.ts
@@ -64,8 +64,12 @@ export const getPreview = (text: string): Promise<string> => apiFetcher.post('/p
 export async function getUser(): Promise<User | null> {
   // probe auth state via /auth/status (always 200, no console 401)
   const status = await authFetcher.get<{ status: string }>('/status').catch(() => null);
-  if (status?.status !== 'logged in') {
-    // clear stale auth cookies — preserves the cleanup-on-probe behaviour previously triggered by /user 401
+  if (status === null) {
+    // probe failed (network blip, transient 5xx) - leave cookies alone, treat as unknown
+    return null;
+  }
+  if (status.status !== 'logged in') {
+    // explicit "not logged in" - clear stale cookies, preserving cleanup-on-probe behaviour previously triggered by /user 401
     clearAuthCookie(JWT_COOKIE_NAME);
     clearAuthCookie(XSRF_COOKIE);
     return null;
--- a/frontend/apps/remark42/app/common/api.ts
+++ b/frontend/apps/remark42/app/common/api.ts
@@ -11,7 +11,8 @@ import {
   Image,
   EmailSubVerificationStatus,
 } from './types';
-import { apiFetcher, adminFetcher } from './fetcher';
+import { apiFetcher, adminFetcher, authFetcher, JWT_COOKIE_NAME, XSRF_COOKIE } from './fetcher';
+import { clearAuthCookie } from './cookies';
 
 /* API methods */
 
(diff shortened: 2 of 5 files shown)


akellbl4 (changes requested):
I don't think this is a proper solution

backend/app/rest/api/rest.go
```diff
@@ -291,10 +291,17 @@ func (s *Rest) routes() chi.Router {
 			rauth.Use(middleware.Timeout(30 * time.Second))
 			rauth.Use(tollbooth_chi.LimitHandler(tollbooth.NewLimiter(10, nil)))
 			rauth.Use(authMiddleware.Auth, matchSiteID, middleware.NoCache, logInfoWithBody)
-			rauth.Get("/user", s.privRest.userInfoCtrl)
```
esev, line 294: Could this line be moved into the open routes group of handlers above, rather than creating a new group for it?

backend/app/rest/api/rest_private.go
```diff
@@ -236,9 +236,21 @@ func (s *private) updateCommentCtrl(w http.ResponseWriter, r *http.Request) {
 	render.JSON(w, r, res)
 }
 
-// GET /user?site=siteID - returns user info
+// GET /user?site=siteID&unauthorised200=false - returns user info, with unauthorised200=true returns 200 with error message
 func (s *private) userInfoCtrl(w http.ResponseWriter, r *http.Request) {
-	user := rest.MustGetUserInfo(r)
+	user, err := rest.GetUserInfo(r)
+	if err != nil {
+		if r.URL.Query().Get("unauthorised200") == "true" {
```
esev, line 243: Is the unauthorised200 parameter necessary? It looks like the frontend code will already accept a 200 response with `null` as the JSON response.

Please correct me if I'm wrong, I'm new here, but I think if `null` can be returned here, it looks like no frontend changes would be needed.

esev (commented):
I'm interested in seeing a fix for #1188. Just a couple drive-by comments in case they can help move this forward.

paskal:
Thanks @esev, your suggestion was spot on -- the simpler shape works without any frontend churn. Reworked the PR along those lines:

- moved `/user` into the existing open routes group instead of creating a new `rauthOptional` group
- dropped the `unauthorised200` query parameter; unauthenticated `/user` requests now return `200` with a `null` body, which the existing `getUser()` `.catch(() => null)` handles transparently
- inlined the site check (mirroring `matchSiteID`) for authenticated requests, replacing the fragile `matchSiteID(...).ServeHTTP(w, r)` call inside the handler
- frontend `getUser()` reverted to its original one-liner -- no API surface change
- updated the test accordingly (renamed to `TestRest_UserInfoUnauthenticated`) covering the unauthenticated-200 path and the authenticated site-mismatch 403

Net diff vs master is now `+52 / -4` across three backend files, no frontend touch.

umputun (commented):
looked through this and have a few things to discuss before merging.

**main question**: in #1188 you yourself suggested switching the frontend to the `/status` endpoint I added to `go-pkgz/auth` ([78e5d29](https://github.com/go-pkgz/auth/commit/78e5d290333d18075f2d316ffd22c4cd1b3482d2)) precisely so we wouldn't have to change the `/user` contract. what made you change direction here? swapping the call site on the frontend feels like the cleaner path, and it doesn't break anyone consuming `/user` with the documented `401 unauthorized` behavior. also worth noting akellbl4's CHANGES_REQUESTED on this PR is still open from the original revision.

**API docs not updated**: `site/src/docs/contributing/api/index.md:111` still says `GET /api/v1/user — auth required`. if we're going with this contract change it has to be reflected in the docs in this same PR.

**DRY**: copilot already raised this and I agree. the inlined site check at `rest_private.go:248-253` is a verbatim copy of `matchSiteID` (`rest.go:580-592`), including the `user.Name == "admin" && user.ID == "admin"` predicate. f8ba3877 just hardened `matchSiteID` to require an explicit `?site=`, and now there are two copies to keep in sync. extract a small helper like `siteAllowed(user, siteID)` and call it from both places.

**stale title**: `unauthorised200 parameter` doesn't match the implementation anymore (no such param). pls update.

**frontend side effect worth flagging**: `frontend/apps/remark42/app/common/fetcher.ts:124-128` clears the JWT cookie on `[401, 403]`. with the new behavior, an expired session calling `/user` returns `200 null`, so this probe no longer triggers cookie cleanup. other authenticated calls still hit 401 and clear it, so it's not a session-leak bug, but the cleanup-on-probe behavior silently changes.

**test gaps**:
1. the new test bundles two unrelated scenarios. split into separate tests or use `t.Run` subtests.
2. missing: basic-auth admin path (the `admin/admin` exemption you just inlined into the handler has zero coverage)
3. missing: authenticated user without `?site=` for non-admin (the 403 your comment promises)
4. missing: invalid/expired JWT - should also yield `200 null`
5. `assert.Equal(t, "null\n", string(body))` is brittle (copilot flagged). better:
   ```go
   var v any
   require.NoError(t, json.Unmarshal(body, &v))
   assert.Nil(t, v)
   ```
6. nits: use `http.MethodGet` not `"GET"` (rest of the file uses the constant); `require.NoError` on `io.ReadAll` not `assert.NoError`; `http.Client{Timeout: 1 * time.Second}` is tight and may flake under CI load.

**minor**: the godoc on `userInfoCtrl` doesn't mention the new 403 site-mismatch branch, and the comment "no auth → 200/null" actually fires on any `GetUserInfo` error including malformed JWT - either narrow the check or fix the comment.

ci is green and the security boundary holds (`Trace` still validates JWTs fully when present, `NoCache` is on the open group). but the strategic question about `/status` vs `/user` reshape is the main thing I'd like to settle before going further.

paskal:
Pivoted to `/auth/status` as you originally suggested in #1188 -- you were right, this is the cleaner path. Force-pushed; branch is now `+81 / -13` across 4 frontend files, no backend changes.

What changed:

- `getUser()` in both `apps/remark42/app/common/api.ts` and `packages/api/clients/public.ts` now probes `/auth/status` first via the auth fetcher; only calls `/user` when the response is `"logged in"`
- when `/auth/status` reports `"not logged in"`, stale `JWT` / `XSRF-TOKEN` cookies are cleared in `getUser()` -- this preserves the cleanup-on-probe behaviour you flagged, since the probe no longer hits the 401/403 path in `fetcher.ts:124-128`
- `/user` contract is untouched, so `site/src/docs/contributing/api/index.md:111` stays accurate

Addressing your other points:

- **DRY**: gone, no backend duplication anymore
- **Stale title**: renamed to `Probe /auth/status from frontend to avoid 401 on /user (closes #1188)`
- **Test gaps**: new shape is covered -- 3 tests in `api.test.ts` (logged in -> fetches `/user`; not logged in -> null + cookies cleared; status request fails -> null) and 3 in `public.test.ts` (same matrix). The basic-auth admin / site-mismatch / expired-JWT branches don't apply now since `/user` is back to its original handler
- **Copilot's brittle `"null\n"` assertion**: gone with the test rewrite

Pre-existing: `packages/api` vitest fails locally on Node 25 due to msw URL parsing, but the same failures show up on master, so unrelated to this change.

@akellbl4 -- since the original approach you objected to is dropped, please take another look when you get a chance.

umputun (commented):
nice pivot, this is the right shape — `/user` contract preserved, no docs churn, no backend changes.

one subtle behavior change worth a look: in `frontend/apps/remark42/app/common/api.ts:64-68`, when the `/auth/status` probe itself fails (network blip, transient 5xx), `.catch(() => null)` falls through into the cookie-clear branch and wipes `JWT` / `XSRF-TOKEN`. previously a transient `/user` failure didn't clear cookies — only an explicit 401/403 in `fetcher.ts:124-128` did. so a flaky network on the probe now silently logs the user out on next page load.

probably want to only clear when the response is unambiguously "not logged in":

```ts
const status = await authFetcher.get<{ status: string }>('/status').catch(() => null);
if (status === null) {
  // probe failed — don't touch cookies, just return null for this call
  return null;
}
if (status.status !== 'logged in') {
  clearAuthCookie(JWT_COOKIE_NAME);
  clearAuthCookie(XSRF_COOKIE);
  return null;
}
return apiFetcher.get<User | null>('/user').catch(() => null);
```

your "status request fails" test would then want a negative assertion (`expect(clearSpy).not.toHaveBeenCalled()`) so the distinction is locked in.

`packages/api/clients/public.ts` doesn't have this concern — it doesn't manage cookies in either branch, which is consistent with its existing fetcher (`packages/api/lib/fetcher.ts` doesn't clear on 401/403 either). good call keeping the asymmetry.

minor: the trailing-comma changes in `packages/api/tests/clients/public.test.ts:46/49/52` look like prettier sweeping in unrelated lines. not a blocker, but ideally trim those out.

otherwise good to go after the cookie-clear tweak.

paskal:
Good catch on the probe-failure path -- you're right, a transient network blip shouldn't silently log the user out. Pushed `7442ac29`:

- `getUser()` now distinguishes `null` (probe failed -> leave cookies, return null) from explicit `"not logged in"` (clear cookies + return null)
- the probe-failure test now asserts `clearSpy).not.toHaveBeenCalled()` to lock the distinction in
- trailing-comma noise in `public.test.ts` is gone; root cause was `packages/api/.prettierrc` not specifying `trailingComma`, so prettier 3 defaulted to `'all'` and `eslint --fix` swept unrelated lines on commit. Set it to `'es5'` to match `apps/remark42`, plus dropped one stray trailing comma in `clients/public.ts:113` that the new config flagged

`packages/api/clients/public.ts` keeps the symmetric short-circuit (no cookie management either way), as you noted.

umputun (approved):
lgtm, both points resolved cleanly, and nice job root-causing the prettier-config drift instead of just deleting the noise.

PR #1766, opened 2024-05-08
Add dependabot updates for Go, TypeScript, GitHub Actions

paskal:
Currently, only security updates and only for remark42 Go app and /site TS code are provided, and after that PR, Go _example and all five TS applications will start receiving updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,74 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      "GitHub Actions updates":
+        patterns:
+          - "*"
+  - package-ecosystem: "gomod"
+    directory: "/backend"
+    schedule:
+      interval: "monthly"
+    groups:
+      "Go modules updates":
+        dependency-type: "production"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,74 @@
...
+  - package-ecosystem: "npm"
+    directory: "/frontend/e2e"
+    schedule:
+      interval: "monthly"
+    groups:
+      "NPM modules updates":
+        dependency-type: "production"
+      "NPM modules updates for tests":
+        dependency-type: "development"
+  - package-ecosystem: "npm"
```
akellbl4, line 57:
```suggestion
  - package-ecosystem: "pnpm"
```

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM node:16.20.2-alpine AS frontend-deps
+FROM --platform=$BUILDPLATFORM node:16-alpine AS frontend-deps
```
akellbl4, line 1: I think getting rid of patch is fine, but we should preserve minor

frontend/Dockerfile.e2e
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/playwright:v1.40.1-jammy
+FROM mcr.microsoft.com/playwright:v1.44.0-jammy
```
akellbl4, line 1: probably fine to drop patch here too
```suggestion
FROM mcr.microsoft.com/playwright:v1.44-jammy
```

akellbl4 (changes requested):
LGTM! I like it being automatic, I hope we have time to check all of the PRs that can pop. The only thing is the correct package manager for the frontend.

.github/dependabot.yml
paskal: package-ecosystem is still npm, there is no pnpm option.

PR #1792, opened 2024-06-30
Add escaping of comment text in webhook default JSON template

### LINKED ISSUE #1791: Admin Notification Webhook receives an additional \n that causes a malformed JSON
Version: `1.13.0`
OS: Linux
Envvar:
```
NOTIFY_ADMINS=webhook
NOTIFY_WEBHOOK_HEADERS="Content-Type:application/json"
NOTIFY_WEBHOOK_URL=http://xxx```
```

### Issue
When configuring an admin notification webhook, remark seem to add an additional `\n` to the webhook content causing the JSON to be malformed at the webhook side.

image

When the webhook is called, the raw body reads as

```
"{\"text\": \"<p>testme</p>\n\"}"  
```

Even though the comment was submitted without any linebreaks after `testme`.

This causes a JSON parsing error on the webhook end.

### Some initial investigation

I've also looked into the `notify` library that remark uses https://github.com/go-pkgz/notify/blob/master/webhook.go#L52-L53 and with the following test that hardcodes a `\n` in the text that is sent, the webhook receives it with the proper escaped `\\n`.

```golang
func main() {
        httpposturl := "http://xxx"
        fmt.Println("HTTP JSON POST URL:", httpposturl)


        var txt = `{"name": "morpheus<p>meow</p>\n"}`

        request, error := http.NewRequest("POST", httpposturl, bytes.NewBufferString(txt))
         request.Header.Set("Content-Type", "application/json")

        client := &http.Client{}
        response, error := client.Do(request)
        if error != nil {
                panic(error)
        }
        defer response.Body.Close()

        fmt.Println("response Status:", response.Status)
        fmt.Println("response Headers:", response.Header)
        body, _ := ioutil.ReadAll(response.Body)
        fmt.Println("response Body:", string(body))

}
 ```

### Expected

Even if there's a linebreak, the linebreak should be escaped (`\\n`) when sending it over to the webhook.

I'm not sure why remark is sending an unescaped `\n`. Is there a way to prevent the linebreak from being sent over?

paskal:
Resolves #1791.

### DIFF
--- a/backend/app/notify/webhook.go
+++ b/backend/app/notify/webhook.go
@@ -3,6 +3,7 @@ package notify
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"text/template"
 	"time"
@@ -12,7 +13,7 @@ import (
 )
 
 const (
-	webhookDefaultTemplate = `{"text": "{{.Text}}"}`
+	webhookDefaultTemplate = `{"text": {{.Text | escapeJSONString}}}`
 )
 
 // WebhookParams contain settings for webhook notifications
@@ -49,7 +50,7 @@ func NewWebhook(params WebhookParams) (*Webhook, error) {
 		params.Template = webhookDefaultTemplate
 	}
 
-	payloadTmpl, err := template.New("webhook").Parse(params.Template)
+	payloadTmpl, err := template.New("webhook").Funcs(template.FuncMap{"escapeJSONString": escapeJSONString}).Parse(params.Template)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse webhook template: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


davidchua:
Thanks for this! Really appreciate the work in resolving #1791 so promptly! 🙇 

Just wondering if moving the escape just prior to sending out the webhook might be better? Unless there is an instance where the user might not want to escape their string prior to sending it out as a webhook?

Pardon me if I'm missing any additional context.

https://github.com/umputun/remark42/blob/ff0efd6fd496e66dda6893caa67b71c6cfdb2d2e/backend/app/notify/webhook.go#L70-L74

paskal:
The thing is you can send anything and not only JSON and escaping everything in JSON style might lead to unexpected consequences.

umputun (approved):

PR #1796, opened 2024-07-01
Add content type check for images endpoint

umputun:
This PR adds a check in image proxy download for validating content type.


umputun:
note: I had to add a way to alter public routes limiter because on a fast local box without enforcing sequential tests execution, we run into limiter's 429. This limiter value is not exposed, used for tests only and shouldn't change anything else

paskal (approved):

PR #1804, opened 2024-08-04
Disallow image proxy for private IPs, add blacklist support

paskal:
This mitigates the problem where a user might probe machines that are unavailable to the user directly but accessible to the server hosting Remark42.

Scenarios addressed:

1. A malicious user could learn about the presence of specific software or hardware running on an internal address. For example, the presence of an image at `http://192.168.0.1/img/container_bottom_shade_login.png` can expose the type of router you have.

2. A malicious user might receive non-timeout `invalid content type` responses from internal addresses, enabling them to scan and identify HTTP servers running in the internal IP range without revealing their content but indicating their presence.

The new functionality is breaking, but I assume no one intends to expose only private network images to the outside world. The old behavior can be restored by setting the `--image-proxy.allow-private-networks` flag.

Additionally, this change adds the `--image-proxy.blacklist` flag to allow blacklisting private parts of the infrastructure from being accessed by the image proxy. You can blacklist IPs (e.g., `8.8.8.8`), CIDR subnets (e.g., `8.8.8.8/31`), and domains (e.g., `private.example .com`). Note that all subdomains of a given domain will also be blacklisted.

`127.0.0.0/8` and `::1/128` ranges are not included as they are most commonly used for local testing, and it would be cumbersome to prohibit them. Localhost is considered less of a security threat than probing other hosts in the network. If desired, localhost can be restricted using the blacklist functionality.

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -133,8 +133,10 @@ type ServerCommand struct {
 
 // ImageProxyGroup defines options group for image proxy
 type ImageProxyGroup struct {
-	HTTP2HTTPS    bool `long:"http2https" env:"HTTP2HTTPS" description:"enable HTTP->HTTPS proxy"`
-	CacheExternal bool `long:"cache-external" env:"CACHE_EXTERNAL" description:"enable caching for external images"`
+	HTTP2HTTPS           bool     `long:"http2https" env:"HTTP2HTTPS" description:"enable HTTP->HTTPS proxy"`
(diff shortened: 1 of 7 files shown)


paskal:
Tested manually, works as I would expect it to work. Responses:

```
{"code":18,"details":"private network access not allowed","error":"private network access not allowed"}
{"code":18,"details":"URL is blacklisted","error":"blacklisted URL"}
```

paskal:
Too much complexity, I dislike it. Closing.

PR #1806, opened 2024-08-12
fix: Apple login integration

### LINKED ISSUE #1789: Sign in with Apple breaks
After configuring all the required variables, Sign in with Apple still fails. The behavior is identical to the demo site, so I suspect this is not because of misconfiguration.

```log
remark42  | 2024/06/25 00:44:58.852 [ERROR] prepare login url for [apple] provider failed
remark42  | 2024/06/25 00:44:58.852 [INFO]  prepare login url for [apple] provider failed - response_mode must be form_post if scope is not empty - 500 - ***.***.***.*** - /auth/apple/login?from=https://***.***.***/web/iframe.html?selfClose&site=blog [caused by auth/provider/apple.go:282 provider.(*AppleHandler).LoginHandler]
remark42  | 2024/06/25 00:44:58.852 [INFO]  GET - /auth/apple/login?from=https://***.***.***/web/iframe.html?selfClose&site=blog - ***.***.*** - 9865eb6e1655 - 500 (58) - 206.795µs
remark42  | 2024/06/25 00:44:58.852 [ERROR] prepare login url for [apple] provider failed
```

After some tracing, I noticed this line, which was added recently: 

https://github.com/umputun/remark42/blob/c04705947a2a611aef51df82fb20d4cbc6fcefb5/backend/app/cmd/server.go#L911

It used to be `form_post` but somehow got changed to `query` without any document or commit message.

tomy0000000:
As mentioned in #1789, Apple login was broken in #1617.

I removed the problematic lines, recompiled a docker image, and confirmed the bug was indeed caused by the wrong `ResponseMode` params.

In addition, I've also improved some sections of docs about how to configure credentials used by Apple integration. One thing worth mentioning is that the term "Client ID" used by remark42 does not appear in Apple's developer dashboard but is used in [some documentation](https://developer.apple.com/documentation/sign_in_with_apple/request_an_authorization_to_the_sign_in_with_apple_server), which can cause some confusion. Therefore, I added an annotation indicating that the "Client ID" can be either "App ID" or "Service ID".

resolves #1789

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -139,7 +139,7 @@ type ImageProxyGroup struct {
 
 // AppleGroup defines options for Apple auth params
 type AppleGroup struct {
-	CID                string `long:"cid" env:"CID" description:"Apple client ID"`
+	CID                string `long:"cid" env:"CID" description:"Apple client ID (App ID or Services ID)"`
 	TID                string `long:"tid" env:"TID" description:"Apple service ID"`
 	KID                string `long:"kid" env:"KID" description:"Private key ID"`
 	PrivateKeyFilePath string `long:"private-key-filepath" env:"PRIVATE_KEY_FILEPATH" description:"Private key file location" default:"/srv/var/apple.p8"`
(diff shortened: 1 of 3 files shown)


tomy0000000:
close #1789

paskal (approved):
LGTM!

Could you please also update server.go CLI param comment to match your changes in the parameters/index.md, and prepare a PR which would update the https://github.com/go-pkgz/auth?tab=readme-ov-file#apple-auth-provider documentation as well?

Also a nitpick, please keep the list numeration to `1.` everywhere: that way you never need to update it in case elements are changing order or removed from the middle.

tomy0000000:
@paskal @umputun Got some time to review this?

paskal (approved):

umputun (approved):

PR #1870, opened 2024-12-05
Implement function to prune string keeping HTML closing tags

### LINKED ISSUE #1587: Reduce number of symbols in Telegram messages
It makes sense not to cite long messages in Telegram, leaving 50 or 100 messages should be enough.

aliksend:
Fixes #1587

### DIFF
--- a/backend/app/notify/telegram.go
+++ b/backend/app/notify/telegram.go
@@ -3,12 +3,16 @@ package notify
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	log "github.com/go-pkgz/lgr"
 	ntf "github.com/go-pkgz/notify"
 	"github.com/hashicorp/go-multierror"
+	"golang.org/x/net/html"
 )
+
+const comment_text_length_limit = 100
 
 // TelegramParams contain settings for telegram notifications
 type TelegramParams struct {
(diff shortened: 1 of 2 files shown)


paskal (changes requested):
Looks good! Could you please move the prune function to a different file? It's not telegram-specific and now increases complexity of telegram file without necessity.

aliksend:
Sure. Sorry for the long wait

Moved function to separated file and rebased my branch onto current master

backend/app/notify/prune_html.go
```diff
@@ -0,0 +1,122 @@
+package notify
+
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/net/html"
+)
+
+const commentTextLengthLimit = 100
```
paskal, line 10: This is the only consonant to move to telegram.go, otherwise LGTM.

paskal (dismissed):

backend/app/notify/prune_html.go
aliksend: you're right, my bad. moved back to telegram.go

aliksend:
I don't know why build fails on CI/CD. Locally I can do both `go build -ldflags "-X main.revision=0.0.0 -s -w"` from `backend/_example/memory_store` directory and `docker buildx build --build-arg SKIP_BACKEND_TEST=true --build-arg SKIP_FRONTEND_TEST=true -f backend/_example/memory_store/Dockerfile .` from root directory. Can you help me with it?

umputun:
> I don't know why build fails on CI/CD. Locally I can do both `go build -ldflags "-X main.revision=0.0.0 -s -w"` from `backend/_example/memory_store` directory and `docker buildx build --build-arg SKIP_BACKEND_TEST=true --build-arg SKIP_FRONTEND_TEST=true -f backend/_example/memory_store/Dockerfile .` from root directory. Can you help me with it?

fixed, passed

paskal (approved):

umputun:
I have generated some unit tests, as the one it had didn't seem too convincing to me.

umputun:
pls note - the test is generated to match the code behavoir, however some of them look strange to me. for example this one

```go
		{
 			name:      "text exceeding limit",
 			html:      "<p>Hello world, this is a long text</p>",
 			maxLength: 15,
 			expected:  "<p>...</p>",
 		},
```

I have not checked the code closely but the common sense expectation would be `<p>Hello...</p>` in this case.

@aliksend can you take a look pls and adjust tests/code if needed?

aliksend:
Current implementation will generate string `"<p>Hello ...</p>"` which length is 16, exceeding the limit. But it make sense to remove space in cases like this. 
Also I'll fix another tests to make behavior better.

paskal (dismissed):

Issue #1877, opened 2024-12-20
SendJWTHeader leads to {"error":"failed to get token"}


nmattia:
I'm not sure how to make `send-jwt-header` work. I'm enabling it as an env var:

```
      - AUTH_SEND_JWT_HEADER=true
```

but unfortunately after going through the GitHub auth flow, I end up on this page (`https://remark42.example.com/auth/github/callback?code=...&state=...`)

```json
{"error":"failed to get token"}
```

This seems to come from [`AuthHandler`](https://github.com/umputun/remark42/blob/3f0789fd90c90841574a6b99bb78f272a03a4a26/backend/vendor/github.com/go-pkgz/auth/v2/provider/oauth2.go#L142) which reads either a cookie or a header (which should get set in [`LoginHandler`](https://github.com/umputun/remark42/blob/3f0789fd90c90841574a6b99bb78f272a03a4a26/backend/vendor/github.com/go-pkgz/auth/v2/provider/oauth2.go#L123)).

It looks like[ frontend support was added](https://github.com/umputun/remark42/pull/834/files) for `send-jwt-header` though it's unclear to me how this works when redirects are used (AFAICT the `fetcher.ts` module is not actually being used in this case).

Any help appreciated!

Note: I'm trying to enable `send-jwt-header` because -- as far as I understand -- it should allow me to have remark42 on a different domain than my website (and not rely on cookies which don't seem to work too well inside iframes on Safari due to `SameSite` not being supported).

paskal:
Based on my investigation, the problem is:

1. The Remark42 backend correctly passes the SendJWTHeader option to the authentication service in server.go (line 1208):
   ```
   SendJWTHeader: s.Auth.SendJWTHeader,
   ```

2. This setting is also exposed to the frontend via the /api/v1/config endpoint (line 464):
   ```
   SendJWTHeader: s.SendJWTHeader,
   ```

3. In the authentication flow, when a user initiates OAuth (e.g., with GitHub), the JWT token is initially set during the login phase:
   ```go
   // in oauth2.go LoginHandler
   if _, err := p.JwtService.Set(w, claims); err != nil {
       rest.SendErrorJSON(w, r, p.L, http.StatusInternalServerError, err, "failed to set token")
       return
   }
   ```

4. The problem is in the token/jwt.go implementation. When SendJWTHeader is true, it correctly sends the JWT as a header (line 261-264):
   ```go
   if j.SendJWTHeader {
[...]

PR #1929, opened 2025-04-27
Fix login persistence with AUTH_SEND_JWT_HEADER enabled

paskal:
## Summary
- When using AUTH_SEND_JWT_HEADER=true, the JWT token in header needed to be persisted to maintain login after page reload
- The frontend now properly handles JWT tokens received in headers by:
  - Storing them in client-side cookies with the name 'JWT' to match what backend expects
  - Extracting and storing the JWT ID as XSRF token in 'XSRF-TOKEN' cookie
  - Setting cookies with Secure flag when on HTTPS connections
- Documentation has been updated to clarify that with this flag enabled, the frontend stores the JWT in a client-side cookie

First part of the fix for #1877, pre-requirement for fixing OAUTH login in #241

## Technical Details
The issue was occurring because the JWT token was received in X-JWT header with AUTH_SEND_JWT_HEADER=true, but the token was only stored in memory. When the page reloaded, the token would be lost, causing users to be logged out.

This implementation allows the frontend to handle authentication consistently regardless of whether AUTH_SEND_JWT_HEADER is enabled or not, making it transparent to users.

## Demo

Before:

https://github.com/user-attachments/assets/7990ca72-5e7c-4dcf-aacc-ad9b6b9d3aa7

After:

https://github.com/user-attachments/assets/c9d7bbf0-a83c-4881-be36-a114a27a39d7

## Concerns

Stored cookie is not [HTTPOnly](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#httponly) as it's written from JS, so could be read from JS on the page as well. I thought it would be better than storing it in local storage, at least we know that the only sensitive stuff is still in the cookie.

### DIFF
--- a/frontend/apps/remark42/app/common/cookies.ts
+++ b/frontend/apps/remark42/app/common/cookies.ts
@@ -8,15 +8,14 @@ interface CookieOptions {
   path?: string;
   domain?: string;
   secure?: boolean;
+  sameSite?: 'Strict' | 'Lax' | 'None';
 }
 
 export function setCookie(name: string, value: string, options: CookieOptions = {}) {
   if (options.expires) {
+    // Convert number (seconds) or Date to UTC string
     if (typeof options.expires === 'number') {
-      const d = new Date();
-      d.setTime(d.getTime() + options.expires * 1000);
-      options.expires = d;
-      options.expires = options.expires.toUTCString();
+      options.expires = new Date(Date.now() + options.expires * 1000).toUTCString();
     } else if (options.expires instanceof Date) {
       options.expires = options.expires.toUTCString();
     }
@@ -27,13 +26,61 @@ export function setCookie(name: string, value: string, options: CookieOptions = 
   let updatedCookie = `${name}=${value}`;
 
   for (const [key, value] of Object.entries(options)) {
-    updatedCookie += `; ${key}`;
-    if (value !== true) {
-      updatedCookie += `=${value}`;
+    // For boolean attributes like 'secure', only add them if true, otherwise skip
+    if (value === true) {
+      updatedCookie += `; ${key}`;
+    }
+    if (typeof value !== 'boolean') {
+      updatedCookie += `; ${key}=${value}`;
     }
   }
 
   document.cookie = updatedCookie;
+}
+
+/**
+ * Sets a cookie with enhanced security options for authentication
+ * @param name The name of the cookie
+ * @param value The value to set
+ * @param options Additional cookie options
+ */
+export function setAuthCookie(name: string, value: string, options: CookieOptions = {}) {
+  const isSecure = window.location.protocol === 'https:';
+  const cookiePrefix = isSecure ? '__Host-' : '';
+
+  // Default options for auth cookies with strong security
+  const authOptions: CookieOptions = {
+    path: '/',
+    sameSite: 'Strict',
+    secure: isSecure,
+    ...options,
+  };
+
+  setCookie(`${cookiePrefix}${name}`, value, authOptions);
+}
+
+/**
+ * Clears an authentication cookie by setting its expiration to the past
+ * @param name The name of the cookie to clear
+ */
+export function clearAuthCookie(name: string) {
+  const isSecure = window.location.protocol === 'https:';
+  const cookiePrefix = isSecure ? '__Host-' : '';
+
+  setCookie(`${cookiePrefix}${name}`, '', {
+    path: '/',
+    secure: isSecure,
+    expires: new Date(0), // Set to epoch time to expire immediately
+  });
+
+  // Also try to clear the non-prefixed version to be thorough
+  if (cookiePrefix) {
+    setCookie(name, '', {
+      path: '/',
+      secure: isSecure,
+      expires: new Date(0),
+    });
+  }
 }
 
 export function getCookie(name: string) {
--- a/frontend/apps/remark42/app/common/fetcher.test.ts
+++ b/frontend/apps/remark42/app/common/fetcher.test.ts
@@ -4,7 +4,16 @@ jest.mock('./settings', () => ({
 
 import { RequestError } from 'utils/errorUtils';
 import { API_BASE, BASE_URL } from './constants.config';
-import { apiFetcher, authFetcher, adminFetcher, JWT_HEADER } from './fetcher';
+import {
+  apiFetcher,
+  authFetcher,
+  adminFetcher,
+  JWT_HEADER,
+  JWT_COOKIE_NAME,
+  XSRF_COOKIE,
+  AUTH_COOKIE_TTL_SECONDS,
+} from './fetcher';
+import * as cookies from './cookies';
 
 type FetchImplementationProps = {
   status?: number;
@@ -31,6 +40,16 @@ function mockFetch({ headers = {}, data = {}, ...props }: FetchImplementationPro
 }
 
 describe('fetcher', () => {
+  // Mock cookies for the test environment
+  beforeEach(() => {
+    // Mock getCookie to always return undefined for XSRF_COOKIE
+    jest.spyOn(cookies, 'getCookie').mockImplementation(() => undefined);
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
   const headers = {};
   const apiUri = '/anything';
   const apiUrl = `${BASE_URL}${API_BASE}/anything?site=remark`;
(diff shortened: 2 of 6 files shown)


umputun (commented):
## Overview
 PR #1929 implements a solution for login persistence when using `AUTH_SEND_JWT_HEADER=true`. The change stores JWT tokens received from server headers in client-side  cookies, allowing login sessions to persist across page reloads.

 ## Key Findings

### Security Concerns

1. **XSS Vulnerability Risk (High)**
         - JWTs are stored in client-accessible cookies without `HttpOnly` flag
         - This makes tokens vulnerable to theft via XSS attacks
         - This is an inherent tradeoff with the `AUTH_SEND_JWT_HEADER` approach

2. **Cookie Security Settings**
         - ✅ `SameSite=Strict` is correctly used for CSRF mitigation
         - ✅ `Secure` flag is added for HTTPS connections
         - ❌ Missing `__Host-` prefix on cookies for additional security

3. **CSRF Protection**
         - ✅ JWT ID (`jti`) is extracted and used as XSRF token
         - ✅ Implementation follows Double Submit Cookie pattern

### Code Quality

 1. **Hardcoded Values**
         - 30-day cookie expiration is hardcoded (`60*60*24*30`)
         - This may not align with server-side JWT TTL configuration

2. **Error Handling**
         - ✅ Comprehensive try/catch blocks for token parsing and cookie setting
         - ✅ Errors are logged but silently fail for users

3. **Documentation**
         - ✅ Updated in server.go and parameters documentation
         - ❌ Missing explicit security warning about XSS risks

## Recommendations

 1. **Security Enhancements**
         - Add `__Host-` prefix to cookies when on HTTPS
         - Add explicit documentation warning about XSS risks
         - Consider aligning cookie expiration with JWT expiry

 2. **Code Improvements**
         - Extract cookie setting to a helper function
         - Use a constant for the cookie expiration time
         - Use a more robust JWT parsing method that handles URL-safe base64

3. **Cleanup Handling**
         - Add explicit cookie cleanup on auth errors/logout

      This implementation provides the needed functionality, but the security tradeoffs should be clearly documented for users enabling this feature.

frontend/apps/remark42/app/common/cookies.ts
```diff
@@ -27,15 +26,66 @@ export function setCookie(name: string, value: string, options: CookieOptions =
...
+    // For boolean attributes like 'secure', only add them if true, otherwise skip
+    if (typeof value === 'boolean') {
+      if (value) {
+        updatedCookie += `; ${key}`;
+      }
+      // Skip false boolean values entirely
+    } else if (value !== undefined) {
+      // For non-boolean values
+      updatedCookie += `; ${key}=${value}`;
     }
```
akellbl4, line 38:
```suggestion
    if (value === true) {
        updatedCookie += `; ${key}`;
      // Skip false boolean values entirely
    }
    if (typeof value !== 'boolean') {
      // For non-boolean values
      updatedCookie += `; ${key}=${value}`;
    }
```

frontend/apps/remark42/app/common/cookies.ts
```diff
@@ -27,15 +26,66 @@ export function setCookie(name: string, value: string, options: CookieOptions =
...
+  });
+
+  // Also try to clear the non-prefixed version to be thorough
+  if (cookiePrefix) {
+    setCookie(name, '', {
+      path: '/',
+      secure: isSecure,
+      expires: new Date(0),
+    });
+  }
```
akellbl4, line 86: these 2 calls are basically identical, the only difference is the name.
Consider calling the function once with different cookie name passed.

frontend/apps/remark42/app/common/cookies.ts
```diff
@@ -27,15 +26,66 @@ export function setCookie(name: string, value: string, options: CookieOptions =
...
+  setCookie(`${cookiePrefix}${name}`, value, authOptions);
+}
+
+/**
+ * Clears an authentication cookie by setting its expiration to the past
+ * @param name The name of the cookie to clear
+ */
+export function clearAuthCookie(name: string) {
+  const isSecure = window.location.protocol === 'https:';
+  const cookiePrefix = isSecure ? '__Host-' : '';
```
akellbl4, line 71: code duplication in `setAuthCookie` and `clearAuthCookie`, consider moving this part into a function

frontend/apps/remark42/app/common/fetcher.test.ts
```diff
@@ -125,6 +150,133 @@ describe('fetcher', () => {
 
       expect(window.fetch).toHaveBeenCalledWith(apiUrl, { method: 'get', headers });
     });
+
+    it('should store JWT token in a cookie when received in header', async () => {
+      // Mock the auth cookie helper
+      const setAuthCookieSpy = jest.spyOn(cookies, 'setAuthCookie').mockImplementation(() => {});
```
akellbl4, line 156: if function called only once per test I'd recommend using methods with `Once` suffix, so no `restoreAllMocks` is needed.
If test fails then the mock cleaning is not called and all the following test cases may be affected.

tip: you can perform expectations on `cookies.setAuthCookie` because the function is wrapped into the spy and the variable is not needed for the expectations.
```suggestion
      const setAuthCookieSpy = jest.spyOn(cookies, 'setAuthCookie').mockReturnValueOnce(undefined);
```
```suggestion
      const setAuthCookieSpy = jest.spyOn(cookies, 'setAuthCookie').mockImplementationOnce(jest.fn());
```

frontend/apps/remark42/app/common/fetcher.test.ts
```diff
@@ -125,6 +150,133 @@ describe('fetcher', () => {
...
+
+      // Create test JWT token - we'll mock the parsing
+      const jwtToken = 'test.jwt.token';
+
+      // Mock parseJwtPayload implementation since it's not directly accessible
+      jest.spyOn(window, 'atob').mockImplementation(() => JSON.stringify({ jti: 'test-jti-id', sub: '1234567890' }));
```
akellbl4, line 162:
```suggestion
      jest.spyOn(window, 'atob').mockReturnValueOnce(JSON.stringify({ jti: 'test-jti-id', sub: '1234567890' });
```

frontend/apps/remark42/app/common/fetcher.test.ts
```diff
@@ -125,6 +150,133 @@ describe('fetcher', () => {
...
+      );
+
+      // Clean up
+      jest.restoreAllMocks();
+    });
+
+    it('should handle errors when setting cookies', async () => {
+      // Mock console.error
+      const originalConsoleError = console.error;
+      console.error = jest.fn();
```
akellbl4, line 217: Doesn't `jest.spyOn(console, 'error')` works for this case?

frontend/apps/remark42/app/common/fetcher.test.ts
```diff
@@ -125,6 +150,133 @@ describe('fetcher', () => {
...
+      await apiFetcher.get(apiUri);
+
+      // Now trigger a 401 response
+      mockFetch({ status: 401 });
+
+      try {
+        await apiFetcher.get(apiUri);
+      } catch (e) {
+        // Expected to fail with 401
+      }
```
akellbl4, line 265: on the first test run snapshot of the error is going to be generated
```suggestion
       expect(() => apiFetcher.get(apiUri)). toThrowErrorMatchingInlineSnapshot();
```

frontend/pnpm-lock.yaml
```diff
@@ -4994,7 +4994,7 @@ packages:
       normalize-path: 3.0.0
       readdirp: 3.6.0
     optionalDependencies:
-      fsevents: 2.3.2
+      fsevents: 2.3.3
```
akellbl4, line 4997: pnpm lock update doesn't seem to be related to the change

frontend/apps/remark42/app/common/cookies.ts
paskal: I've applied it but I am almost sure it doesn't work the same as it's not accounting for undefined.

frontend/apps/remark42/app/common/cookies.ts
paskal: I don't get you.

frontend/apps/remark42/app/common/cookies.ts
paskal: This is okay.

frontend/pnpm-lock.yaml
paskal: As intended,  it's to make `npx update-browserslist-db@latest` go away.

frontend/apps/remark42/app/common/cookies.ts
akellbl4:
```ts
setCookie(`${cookiePrefix ?? ""}${name}`, '', {
    path: '/',
    secure: isSecure,
    expires: new Date(0), // Set to epoch time to expire immediately
  });
```

paskal: I'm cleaning up both here, but the code you're proposing cleaning only one variant.

PR #1947, opened 2025-09-04
docs: improve backup instructions; closes #1613

### LINKED ISSUE #1613: Improve backup docs by clarifying requirements to create a fully functional backup
[The Backup docs state:](https://remark42.com/docs/backup/backup/)

> Remark42 by default makes daily backup files under `${BACKUP_PATH}`

I was under the impression that by downloading the contents of the `${BACKUP_PATH}`, I have a series of fully functional backups of my Remark42 server. After playing around with moving Remark42 hosts and restoring those backups, I found that avatars of existing comments were broken, however. At first I thought I had to somehow re-generate / re-cache avatars, but I think that isn't possible.

Under **Technical Details** of the [Backend Docs](https://remark42.com/docs/contributing/backend/) it's stated that:

> To migrate/move Remark42 to another host, boltdb files and avatars directory AVATAR_FS_PATH should be transferred. Optionally, boltdb can be used to store avatars as well.

Looking over the [CLI parameters docs](https://remark42.com/docs/configuration/parameters/), I think that `IMAGE_FS_PATH` (or `IMAGE_BOLT_FILE` if `IMAGE_TYPE` is `bolt`) also has to be considered (I think this is where inline comment images are stored?).

What about `IMAGE_FS_STAGING`?

---

I think the backup docs aren't crystal clear on what else needs to be backed up besides automatic/manual backups. We should at least mention the relevant paths to backup when running Remark with the default configuration:

- `AVATAR_FS_PATH`, i.e., `./var/avatars`
- `IMAGE_FS_PATH`, i.e., `./var/pictures`

Anything I missed?

If you agree on this issue, I can open a PR to extend the docs.

schnerring:
I just migrated Remark42 to another host, so I decided to finally send the PR! Only took me two years, but better late than never 😅

### DIFF
--- a/site/src/docs/backup/backup/index.md
+++ b/site/src/docs/backup/backup/index.md
@@ -18,3 +18,5 @@ This command creates `userbackup-{site ID}-{timestamp}.gz` file by default.
 ## Backup format
 
 The backup file is a text file with all exported comments separated by EOL. Each backup record is a valid JSON with all key/value unmarshaled from the `Comment` struct (see [here](https://remark42.com/docs/contributing/api/#commenting)).
+
+Avatars and images are stored separately. By default avatars are stored in `./var/avatars` and images in `./var/pictures` and need to be backed up, as well. For more information, check the [technical details of the backend documentation](https://remark42.com/docs/contributing/backend/#technical-details).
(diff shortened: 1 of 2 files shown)


paskal (approved):
Super, thanks a lot!

PR #1960, opened 2025-10-14
fix: Modify the example to avoid misunderstanding

km0e:
In the section `On a separate domain`, it may be misleading to see that `ALLOWED_HOSTS` is set to some domain names. Although the domain names here refer to URLs, it is recommended to modify them slightly.


paskal:
Found relevant documentation: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#host-source

paskal (approved):

PR #1962, opened 2025-10-22
Romanian language added

notmalicik:
I added the romanian language, and thx for your service
*the packages lock is changed beacuse the npm install didnt work on my pc, maybe because of arch liux idk.
Instantaneu din 2025-10-22 17-10-56


frontend/package-lock.json
Mavrin: It looks like you are using npm as your package manager. You should use pnpm instead. Please remove the package-lock.json file and squash your commits.

notmalicik: removed

notmalicik: done

PR #1966, opened 2025-11-06
Premoderation

fedmag:
Hello, this is a first draft for the addition of pre-moderation.
It uses an "approved" flag to define if a comment should be visible to non-admin users. 
Only admins can approve/disapprove comments.

There are 3 modes: 
- __none__: this is the default and it basically "approves" all comments by default
- __all__: every new comment must be approved before it becomes visible to non-admin users.
- __first__: only the first comment from a given user must be approved. All subsequent and previous messages are approved as one of them is approved. If an approved comment is disapproved, only that specific comment becomes unapproved.

In the UI, I added a red border for unapproved comments so that they are easy to spot:
image
I'd assume this solution is not desirable, but I did not have any other idea.


fedmag:
> @fedmag Hi, is it possible to add modes "with links" or/and "with images"?

Hi @cowsay, if this goes in and the code owners want such a feature, I can certainly look into that. 
I can see why it might be useful.

Issue #2011, opened 2026-02-23
iframe background pure white on firefox (even in dark mode)


nephatrine:
The comment widget looks great when I'm using a Chromium-based browser like Chrome or Edge... but if I use a Firefox-based browser the iFrame background is just plain white and it looks hideous.

in Chrome/Edge (Chromium-Based):

<img width="1511" height="309" alt="Image" src="https://github.com/user-attachments/assets/daa77bd7-3dd7-4cf1-8413-87a2ec549307" />

in Firefox/Zen (Gecko-Based):

<img width="1440" height="285" alt="Image" src="https://github.com/user-attachments/assets/ea117b58-9348-4117-8d1b-aa7d89c9a68d" />

These images are both with theme set to dark mode.

PR #2023, opened 2026-03-21
Fix Firefox dark mode white background on comment iframe

### LINKED ISSUE #1430: support color-scheme
- add new option for theme to the config in order to understand if site supports auto switch between light and dart theme
- add `color-scheme: light dark` when the param is `true`,  add `light` or `dark` when `false` and switch `coloe-scheme` value accordingly
- remove this change https://github.com/umputun/remark42/pull/1429

### Auto

```ts
var remark_config = {
  theme: 'auto'
}
```
```css
color-scheme: light dark;
```
Add support for auto switch between themes

### Dark
```ts
var remark_config = {
  theme: 'dark'
}
```
```css
color-scheme: dark;
```

### Light
```ts
var remark_config = {
  theme: 'light'
}
```
```css
color-scheme: light;
```

amdevz:
## Summary

- Firefox renders a white background on the comment iframe in dark mode because `color-scheme: none` is not a valid CSS value. Chrome silently ignores it, but Firefox treats it literally and uses the default light canvas.
- Sets `color-scheme` on the iframe's inner `<html>` element to match the active theme (`dark` or `light`), which tells Firefox to use the correct canvas color.
- Also updates the outer `<iframe>` element's `color-scheme` on creation and on runtime theme changes via `changeTheme()`.
- Resolves #1430, #2011, and builds on the groundwork from #1427 and #1429.

## Changes

- **`remark.tsx`**: Sets `document.documentElement.style.colorScheme` inside the iframe on init and when receiving theme change messages
- **`embed.ts`**: Updates `iframe.style.colorScheme` when `changeTheme()` is called at runtime
- **`create-iframe.ts`**: Replaces invalid `colorScheme: 'none'` with a theme-aware value (`'dark'`, `'light'`, or `'light dark'`)

## Test plan

- Open `http://127.0.0.1:8080/web/` in Firefox and Chrome
- Toggle dark mode and verify no white background appears on the iframe in either browser
- Switch back to light mode and verify it renders correctly
- Verify `changeTheme('dark')` and `changeTheme('light')` in the console both update the iframe canvas color
#### Dark
image

#### Dark
image

#### Light
image

#### Dark
image

#### Light
image

#### Dark
image

### DIFF
--- a/frontend/apps/remark42/app/embed.ts
+++ b/frontend/apps/remark42/app/embed.ts
@@ -116,6 +116,7 @@ function createInstance(config: typeof window.remark_config) {
 
   function changeTheme(theme: Theme) {
     window.remark_config.theme = theme;
+    iframe.style.colorScheme = theme;
     postMessageToIframe(iframe, { theme });
   }
 
--- a/frontend/apps/remark42/app/remark.tsx
+++ b/frontend/apps/remark42/app/remark.tsx
@@ -40,15 +40,20 @@ async function init(): Promise<void> {
 
     if (data.theme === 'light') {
       document.body.classList.remove('dark');
+      document.documentElement.style.colorScheme = 'light';
     }
 
     if (data.theme === 'dark') {
       document.body.classList.add('dark');
+      document.documentElement.style.colorScheme = 'dark';
     }
   });
 
   if (theme === 'dark') {
     document.body.classList.add('dark');
+    document.documentElement.style.colorScheme = 'dark';
+  } else {
+    document.documentElement.style.colorScheme = 'light';
   }
 
   boundActions.fetchHiddenUsers();
--- a/frontend/apps/remark42/app/utils/create-iframe.ts
+++ b/frontend/apps/remark42/app/utils/create-iframe.ts
@@ -20,7 +20,7 @@ export function createIframe({ __colors__, styles, ...params }: Params) {
     padding: 0,
     margin: 0,
     overflow: 'hidden',
-    colorScheme: 'none',
+    colorScheme: params.theme === 'dark' ? 'dark' : params.theme === 'light' ? 'light' : 'light dark',
     ...styles,
   });
 
--- a/frontend/apps/remark42/app/utils/create-iframe.ts
+++ b/frontend/apps/remark42/app/utils/create-iframe.ts
@@ -20,7 +20,7 @@ export function createIframe({ __colors__, styles, ...params }: Params) {
     padding: 0,
     margin: 0,
     overflow: 'hidden',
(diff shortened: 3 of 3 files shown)


paskal:
Thanks for the fix, the problem is real - Firefox does handle the invalid `color-scheme: none` differently from Chrome.

Some context on the history here:

- [#1421](https://github.com/umputun/remark42/pull/1421) - original dark mode background issue
- [#1427](https://github.com/umputun/remark42/pull/1427) - Firefox sort select background fix attempt, led to discovering the `color-scheme` root cause
- [#1429](https://github.com/umputun/remark42/pull/1429) - `colorScheme: 'none'` was added intentionally as a workaround to stop browser auto-theming until proper support is implemented
- [#1430](https://github.com/umputun/remark42/issues/1430) - the plan for full `color-scheme` support including `theme: 'auto'`

`colorScheme: 'none'` is not a valid CSS value, but it worked as a "disable" hack in Chrome. Firefox treating it differently is the actual bug, and your fix for explicit `dark`/`light` themes is correct.

One concern with the fallback in `create-iframe.ts`:

```ts
colorScheme: params.theme === 'dark' ? 'dark' : params.theme === 'light' ? 'light' : 'light dark'
```

When theme is not set (undefined), this falls through to `'light dark'`, which tells the browser to pick based on `prefers-color-scheme`. The problem is that the `.dark` class on body won't be applied, so you'd get a dark canvas with light-themed UI on top - exactly the mismatch that [#1429](https://github.com/umputun/remark42/pull/1429) was meant to prevent.

The fallback should be `'light'` instead of `'light dark'`, matching the default light theme.

amdevz:
> One concern with the fallback in `create-iframe.ts`:
> 
> ```ts
> colorScheme: params.theme === 'dark' ? 'dark' : params.theme === 'light' ? 'light' : 'light dark'
> ```
> 
> When theme is not set (undefined), this falls through to `'light dark'`, which tells the browser to pick based on `prefers-color-scheme`. The problem is that the `.dark` class on body won't be applied, so you'd get a dark canvas with light-themed UI on top - exactly the mismatch that [#1429](https://github.com/umputun/remark42/pull/1429) was meant to prevent.
> 
> The fallback should be `'light'` instead of `'light dark'`, matching the default light theme.

Of course, thanks for taking a look. That's a valid concern, I think for this PR defaulting to `light` makes sense. Long term might be best to add auto detection using [`window.matchMedia`](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia#:~:text=Use%20this%20object%27s%20properties%20and%20events%20to%20detect%20matches%20and%20to%20monitor%20for%20changes) potentially if the theme is undefined

paskal (approved):

PR #2041, opened 2026-04-12
Fix #2040: preserve orig verbatim in edit textarea

### LINKED ISSUE #2040: &lt; converts to < and after edit to nothing
\&lt;script\&gt;Hacked you!\&lt;/script\&gt; 

This displays as:
<script>Hacked you!</script>

Which is maybe the problem in the first place.

But when I edit this, the edit input box contains:
<script>Hacked you!</script>

instead of
\&lt;script\&gt;Hacked you!\&lt;/script\&gt; 

If I save the edit, the entire line is gone and empty.

This seems to be a display bug.

BTW I've just edited this GitHub issues and it works as expected.

paskal:
## Summary

Fixes #2040. Removes a stale HTML-entity decode from the edit textarea flow. When a user posted a comment containing literal HTML entities (e.g. `&lt;script&gt;Hacked you!&lt;/script&gt;`), the edit textarea would show the decoded `<script>...</script>`. Saving that decoded text made bluemonday strip the now-real `<script>` tag on the way back to storage, collapsing the comment body to an empty string.

## Root cause

`frontend/apps/remark42/app/components/comment/comment.tsx:306-313` was running `props.data.orig` through the browser's HTML parser via a detached `<span>.innerHTML` to "decode entities":

```tsx
orig: isEditing
  ? props.data.orig &&
    props.data.orig.replace(/&[#A-Za-z0-9]+;/gi, (entity) => {
      const span = document.createElement('span');
      span.innerHTML = entity;
      return span.innerText;
    })
  : props.data.orig,
```

That block was added in commit `78b9a10` (2018) because the backend `Comment.Sanitize()` at the time ran bluemonday on **both** `Text` and `Orig`. bluemonday HTML-escapes raw `<`/`>`/`&` in text content, so a user typing `1 < 2` got `orig = "1 &lt; 2"` back from the API, and the frontend compensated.

Commit `243c835` (2022, _"do not sanitise the original comment markdown"_) removed bluemonday from `Orig` in the backend. After that, `Orig` is stored and returned byte-for-byte. The frontend decode was never removed alongside — it has been a silent corruption path for ~3.5 years, only surfacing when a user actually types entities they want preserved.

See issue #2040 for the full investigation trail.

## Fix

Delete the decode block. The edit textarea now receives `props.data.orig` unchanged, matching the backend contract documented at `backend/app/store/comment.go:118-120`: _"Comment.Orig … is not sanitized as we expect to never render it as HTML"_.

## Test coverage

Added a 45-case table-driven regression test in `comment.test.tsx` — `<Comment /> › edit textarea preserves data.orig verbatim`. Payloads come from three independent brainstorming passes covering:

- **HTML entities**: canonical issue repro, doubly-escaped (`&amp;lt;`), decimal/hex numeric, named entity mix, malformed entities, real `<` adjacent to `&lt;`, recursive-looking sequences
- **Unicode**: zero-width characters, bidi mix, RTL overrides, decomposed vs precomposed, ZWJ emoji, supplementary-plane surrogate pairs, LS/PS separators, BOM, homoglyphs, full-width characters, whitespace variety
- **Markdown interactions**: code spans with raw/entity script tags, fenced code blocks, javascript: links, images with entity alt/title, backslash escapes, entities inside emphasis/headers/blockquotes, raw autolinks, iframe/object/embed chains, kitchen sink

Verified the test fails as intended by reverting the fix and re-running — 24 of 45 cases (every entity-containing payload) fail against the buggy code, all 45 pass against the fixed code.

Also verified end-to-end in a real browser against `make rundev`: typed the issue's canonical payload, saw it display as literal text, clicked Edit, confirmed the textarea shows `&lt;script&gt;...` (previously `<script>...`), saved, confirmed the body survives (previously empty).

## Behavioural note

The test found one incidental DOM-level normalisation: HTML5 mandates that `<textarea>.value` normalises `\r\n` and lone `\r` to `\n`. This happens in the browser regardless of any remark42 code, so the guarantee is _"byte-equal after newline normalisation"_ — documented inline in the test.

## Screenshots

image

Edit without this change on https://remark42.com/demo/:

image

Edit with this change:
image

### DIFF
--- a/frontend/apps/remark42/app/components/comment/comment.test.tsx
+++ b/frontend/apps/remark42/app/components/comment/comment.test.tsx
@@ -7,6 +7,7 @@ import { render } from 'tests/utils';
 import { StaticStore } from 'common/static-store';
 
 import { Comment, CommentProps } from './comment';
+import { CommentForm } from 'components/comment-form';
 import { CommentMode } from 'common/types';
 
 function CommentWithIntl(props: CommentProps) {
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm

PR #2044, opened 2026-04-17
fix(ssrf): apply ssrf-safe transport to TitleExtractor + restore gosec G70x rules

paskal:
SSRF in `TitleExtractor` plus the lint hygiene that was needed to catch this class of bug going forward.

This is one of four PRs split out from the original audit work; the others are #2045 (path traversal), #2046 (matchSiteID bypass), #2047 (TZ-agnostic store tests). They are independent — merge in any order.

The fifth audit finding — OAuth open-redirect — lives upstream in `go-pkgz/auth/v2` ([branch ready](https://github.com/go-pkgz/auth/pull/new/fix-oauth-from-open-redirect)) and will land here after the auth release.

## SSRF in TitleExtractor — incomplete fix from `aca0cff3`

**Where:** `store/service/title.go:64` (production wiring at `cmd/server.go:524`)

The image proxy got an SSRF-safe transport in `aca0cff3` — same one we now apply to TitleExtractor. The image proxy was hardened, the title fetcher was missed, and the gosec G704 rule was excluded project-wide as part of the same commit so the linter never flagged the gap.

`TitleExtractor` was left using `http.DefaultTransport`. Its hostname allowlist checks the parsed URL host but never the resolved IP, so a domain suffix-matching an allowed host (or `127.0.0.1` itself when `AllowedHosts` is empty) reaches cloud metadata and any other internal endpoint.

**Fix:** extract the SSRF transport from `proxy/image.go` into a new shared `safehttp` package (one source of truth) and apply `safehttp.Transport()` to the `TitleExtractor`'s `http.Client` at construction. The image proxy now imports the same helper instead of having its own copy — no behaviour change for the image path.

**Reproduction** in `title_test.go` (`TestTitle_GetBlocksPrivateIPViaSafeTransport`) demonstrates both the protection (with safehttp) and the original vulnerability (control case without it):

```go
t.Run("with safehttp transport: blocked", ...)   // PASSES — fix works
t.Run("control: default transport leaks", ...)   // PASSES — confirms vulnerability without fix
```

### What does TitleExtractor do, and what's the impact of stricter behaviour?

It fetches the `<title>` tag of the page where the comment was posted (`comment.Locator.URL`), so each comment carries the post's title (`PostTitle`) for use in the admin UI, RSS feeds, and admin notifications. Two trigger points: every new comment without a client-supplied title, and the admin `PUT /admin/title/{id}` endpoint.

**Failure mode after the fix is degraded UX, not data loss:** comments are still created and still display correctly on the post page — only `PostTitle` stays empty when the configured `Locator.URL` resolves to a private IP. The official embed widget already sends `PostTitle` in the request body (read from `document.title` client-side), so the server-side fetch is a fallback for clients that don't. For self-hosted Docker Compose / private-network setups that *do* depend on it, the existing `--allowed-hosts` config can be reused as an SSRF bypass for exact host matches if needed (would be a small follow-up).

## Re-enable gosec `G703`/`G704`/`G705` + clean up all related findings

`aca0cff3` excluded path-traversal, SSRF and XSS taint-analysis rules project-wide. With the SSRF closed (#2045 takes care of path traversal), restoring the rules brings back coverage for future regressions. Every finding the restored rules raised is addressed:

- Image-proxy `http.NewRequest`, QR PNG `Write`, two RSS XML `Write`s — `//nolint:gosec` with reason
- `cmd/{backup,cleanup,import,remap}.go` HTTP calls — operator-controlled URLs from CLI flags, not user input — `//nolint:gosec` with reason
- `migrator.go` temp-file `os.Remove` (path from `os.CreateTemp`) — `//nolint:gosec` with reason
- `rest_private.savePictureCtrl` and `migrator` `ParseMultipartForm` — wrapped with `http.MaxBytesReader` (real fix, not suppression)
- `fs_store.go` Walk-callback `os.Remove` in staging — `//nolint:gosec` (staging dir is server-only)

## Verification

```
$ cd backend && /tmp/golangci-lint run app/...     # CI's v2.10.1
0 issues.
```

`go test ./...` — all packages pass.

### DIFF
--- a/backend/.golangci.yml
+++ b/backend/.golangci.yml
@@ -26,9 +26,6 @@ linters:
     gosec:
       excludes:
         - G117 # false positive: struct field name matches "secret" pattern
-        - G703 # false positive: path traversal via taint analysis
-        - G704 # false positive: SSRF via taint analysis
-        - G705 # false positive: XSS via taint analysis
     gocritic:
       disabled-checks:
         - wrapperFunc
--- a/backend/app/cmd/backup.go
+++ b/backend/app/cmd/backup.go
@@ -47,7 +47,7 @@ func (ec *BackupCommand) Execute(_ []string) error {
 	req.SetBasicAuth("admin", ec.AdminPasswd)
 
 	// get with timeout
-	resp, err := client.Do(req.WithContext(ctx))
+	resp, err := client.Do(req.WithContext(ctx)) //nolint:gosec // exportURL is built from operator-supplied CLI flags, not user input
 	if err != nil {
 		return fmt.Errorf("request failed for %s: %w", exportURL, err)
 	}
--- a/backend/app/cmd/cleanup.go
+++ b/backend/app/cmd/cleanup.go
@@ -199,7 +199,7 @@ func (cc *CleanupCommand) deleteComment(c store.Comment) error { //nolint:dupl /
 
 	client := http.Client{}
 	defer client.CloseIdleConnections()
-	r, err := client.Do(req)
+	r, err := client.Do(req) //nolint:gosec // RemarkURL comes from operator CLI flag, not user input
 	if err != nil {
 		return fmt.Errorf("delete request failed for comment %s, %s: %w", c.ID, c.Locator.URL, err)
 	}
@@ -221,7 +221,7 @@ func (cc *CleanupCommand) setTitle(c store.Comment) error { //nolint:dupl // not
 
 	client := http.Client{}
 	defer client.CloseIdleConnections()
-	r, err := client.Do(req)
+	r, err := client.Do(req) //nolint:gosec // RemarkURL comes from operator CLI flag, not user input
 	if err != nil {
 		return fmt.Errorf("title request failed for comment %s, %s: %w", c.ID, c.Locator.URL, err)
 	}
--- a/backend/app/cmd/import.go
+++ b/backend/app/cmd/import.go
@@ -42,7 +42,7 @@ func (ic *ImportCommand) Execute(_ []string) error {
 	}
 	req.SetBasicAuth("admin", ic.AdminPasswd)
 
-	resp, err := client.Do(req.WithContext(ctx)) // closes request's reader
+	resp, err := client.Do(req.WithContext(ctx)) //nolint:gosec // importURL built from operator CLI flags, not user input; closes request's reader
 	if err != nil {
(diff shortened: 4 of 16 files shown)


umputun (dismissed):
lgtm. the split into `safehttp` is clean, CIDR list + dial-time IP reuse correctly defeats DNS rebinding, and the control-case test in `title_test.go` is load-bearing — it'll catch a future refactor that strips the transport wiring.

audited other `http.Client` uses across `backend/app/` — nothing else fetches user-supplied URLs (cmd/* are operator flags, webhooks/telegram are admin-configured, internal RPC is internal). coverage is complete.

`MaxBytesReader` additions in `rest_private.savePictureCtrl` and `migrator.importFormCtrl` are correctly placed before `ParseMultipartForm` — real DoS fixes, not cosmetic.

couple minor nits, not blockers:

1. `safehttp_test.go:65-78` has a dead `dialer := ...` / `_ = dialer` that can go.
2. `safehttp.Transport()` allocates a fresh transport per call; `proxy/image.go:downloadImage` calls it per request so keepalive is defeated. this is pre-existing behaviour you preserved exactly, so not this PR's problem — worth a followup to either cache via `sync.Once` or rename to `NewTransport()` to make the cost explicit.
3. CGNAT `100.64.0.0/10` blocking is correct for SSRF but could surprise operators running remark42 inside a CGNAT environment. fine as-is, just flagging.

paskal:
Pushed fixup commit `e6a58075` addressing review:

- **safehttp.Transport** now clones `http.DefaultTransport` so Proxy/HTTP2/idle/TLS-handshake/expect-continue defaults are inherited (covered by new `TestTransport_PreservesDefaultTransportSettings`).
- `TestTransport_AllowsPublic` no longer dials TEST-NET-3 unbounded — uses a 100ms context to keep it deterministic; dead `dialer` var dropped.
- `proxy/image.go`: documented the `Image.Transport` contract on the struct field (nil → safehttp; caller-supplied → caller owns SSRF safety) and rewrote the `//nolint:gosec` comments to point at that contract instead of the misleading "SSRF mitigated by safehttp.Transport" claim.

Deferring the per-call `Transport()` allocation point to a follow-up as suggested. CGNAT/0.0.0.0/8/multicast scope-of-blocklist discussion is also a separate change — happy to take that as a follow-up if you want it widened.

umputun (approved):
lgtm, thx

PR #2045, opened 2026-04-17
fix(api): reject path traversal in /picture/{user}/{id}

paskal:
Unauthenticated path traversal in `GET /api/v1/picture/{user}/{id}`.

## Reproduction on `demo.remark42.com` (master, before this PR)

```console
$ curl -sS --path-as-is -i "https://demo.remark42.com/api/v1/picture/../remark.db" | tail -2
content-length: 191

{"code":18,"details":"can't get image ../remark.db","error":"can't get image file for ../remark.db: can't get image stats for ../remark.db: stat var/24/remark.db: no such file or directory"}
```

The 400 body leaks `var/24/remark.db` — confirming `..` reaches `path.Join` unsanitised. Default `Partitions=100` makes the file at the resolved path not exist in practice, but with `Partitions=0` (a documented option) the same probe returns the BoltDB. The error response also leaks the internal filesystem path in either configuration.

## After this PR

Same three probes run against a local handler built from this branch:

```console
$ curl -sS --path-as-is -i "http://localhost:xxxx/api/v1/picture/../remark.db"
HTTP 400
{"code":18,"details":"invalid picture id","error":"invalid picture id"}

$ curl -sS --path-as-is -i "http://localhost:xxxx/api/v1/picture/dev_user/..%2Fremark.db"
HTTP 400
{"code":18,"details":"invalid picture id","error":"invalid picture id"}

$ curl -sS --path-as-is -i "http://localhost:xxxx/api/v1/picture/%2E%2E/remark.db"
HTTP 400
{"code":18,"details":"can't get image","error":"image not found"}
```

`..` and `..%2F` are rejected at the handler entry with a generic `invalid picture id` (input-validation trip). `%2E%2E` is preserved by chi as a literal segment, falls through to the regular image load, and returns the sanitised `image not found`. In all cases: no attacker-supplied string is echoed, no internal filesystem path leaks, the original error is logged server-side for operators.

## Fix

`safePictureSegment` validates both URL params (rejects `..`, `/`, `\\`, NUL) at the handler entry. The raw filesystem error is replaced with a generic `image not found` and the original is logged for operators.

Reproduction test in `rest_public_test.go` covers `..`, `..` in id segment, and percent-encoded `..`.

## Verification

`go test ./rest/api/...` — green.

### DIFF
--- a/backend/app/rest/api/rest_public.go
+++ b/backend/app/rest/api/rest_public.go
@@ -373,13 +373,13 @@ func (s *public) listCtrl(w http.ResponseWriter, r *http.Request) {
 // etc.) also closes a log-injection vector since the rejected segment is
 // echoed into the access log.
 func safePictureSegment(seg string) bool {
-	if seg == "" || seg == "." || seg == ".." {
+	if seg == "" || seg == "." {
 		return false
 	}
 	if strings.ContainsAny(seg, "/\\") {
 		return false
 	}
-	if strings.Contains(seg, "..") {
+	if strings.Contains(seg, "..") { // also covers seg == ".."
 		return false
 	}
 	for _, r := range seg {
@@ -464,7 +464,7 @@ func (s *public) telegramQrCtrl(w http.ResponseWriter, r *http.Request) {
 	}
 
 	w.Header().Set("Content-Type", "image/png")
-	if _, err = w.Write(png); err != nil { //nolint:gosec // png bytes from go-qrcode, not HTML
+	if _, err = w.Write(png); err != nil {
 		log.Printf("[WARN] can't render qr, %v", err)
 	}
 }
--- a/backend/app/rest/api/rest_public.go
+++ b/backend/app/rest/api/rest_public.go
@@ -364,12 +364,36 @@ func (s *public) listCtrl(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// safePictureSegment reports whether seg is acceptable as a path segment in
+// the picture URL (no traversal markers, no path separators, no NULs). Picture
+// IDs are server-generated hashes plus a known extension, so any value carrying
+// these characters is hostile and must be rejected before reaching the store.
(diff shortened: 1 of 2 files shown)


umputun (dismissed):
lgtm on the fix — `safePictureSegment` + generic error + server-side log with `%q` escaping is the right shape, and the three test cases cover what matters.

one scope thing: the `//nolint:gosec` on `telegramQrCtrl` in `rest_public.go:458` is byte-identical to the same line in #2044 (gosec-rule restoration PR). so "merge in any order" isn't quite true — whichever lands second will conflict on that line. I'd drop it from this PR and let #2044 own it.

minor nit, not a blocker: `seg == ".."` in `safePictureSegment` is already covered by the `strings.Contains(seg, "..")` check two lines down, so it's dead. leaving defense-in-depth intentional is fine — just flagging in case you want to trim.

paskal:
Pushed fixup commit `65677830` addressing the log-injection finding (Copilot #2045-1):

`safePictureSegment` now rejects any `unicode.IsControl` rune (CR, LF, TAB, plus the NUL that was already caught). New `TestRest_LoadPictureRejectsControlCharsInSegment` exercises LF/CR/TAB/NUL across both URL segments.

On the QR PNG `//nolint:gosec` comment in this PR's commit history — that line lives on the SSRF/lint sister PR (#2044). I'll address there if it's worth doing.

paskal:
Pushed fixup `3b98a0cf` addressing both review points:

* Dropped the duplicate `//nolint:gosec` on `telegramQrCtrl`'s `w.Write(png)` — #2044 owns it now (otherwise the two PRs collide on that exact line).
* Trimmed `seg == ".."` in `safePictureSegment` — already covered by the `strings.Contains(seg, "..")` check below; left an inline comment naming the cover-by-superset.

umputun (approved):
lgtm, thx

PR #2046, opened 2026-04-17
fix(api): require explicit ?site= in matchSiteID middleware

paskal:
Multi-tenant isolation gap. The middleware short-circuits when `?site=` is omitted, allowing authenticated users to bypass cross-site checks on every protected and admin route.

This is one of four PRs split out from the original audit work; the others are #2044 (SSRF + lint), #2045 (path traversal), #2047 (TZ-agnostic store tests). They are independent — merge in any order.

## The bug

`backend/app/rest/api/rest.go:586`

```go
siteID := r.URL.Query().Get("site")
if siteID != "" && user.SiteID != siteID {
    http.Error(w, "Access denied", http.StatusForbidden)
    return
}
```

Drop the `?site=` query parameter and the check is skipped. On endpoints that read site from the JSON body (email/telegram subscribe, confirm) a user authenticated to siteA can perform the action against siteB without the cross-site guard ever firing.

## Fix

Require `?site=` to be present and to match `user.SiteID`. Body-only flows still work as long as the URL also carries the matching `?site=` — both sources must agree.

## Compatibility — official frontend is safe

The bundled embed widget always includes the param. From `frontend/apps/remark42/app/common/fetcher.ts:66`:

```js
const queryString = new URLSearchParams({ site: siteId, ...query });
```

Every call made by `apiFetcher` / `adminFetcher` — comments, votes, email and telegram subscribe + confirm, deleteme, admin actions — automatically gets `site=siteId` prepended, and the same value is injected into JSON bodies. The matchSiteID middleware sees it and the comparison works. No frontend change needed.

**Who could be affected:** custom API consumers or shell scripts that omit `?site=`. Such clients were already malfunctioning under multi-site deployments (their requests landed on whatever site with no isolation), but they would need to add the param to keep working after this fix. Worth a one-line CHANGELOG note.

## Reproduction

`TestRest_matchSiteID` in `rest_test.go` enumerates four cases (matching, mismatched, missing, empty `?site=`) — all exercise the middleware directly.

Existing test fixtures that relied on the implicit pass had to add `?site=remark42` to URLs:
- `addCommentGetCreatedTime` helper now derives the param from `c.Locator.SiteID`
- picture upload URL gets the param explicitly
- email/telegram subscribe table adds it to every endpoint
- the negative cases that previously asserted `StatusBadRequest` from the handler now correctly assert `StatusForbidden` from the middleware

## Verification

`go test ./rest/api/...` — green.

### DIFF
--- a/backend/app/cmd/server_test.go
+++ b/backend/app/cmd/server_test.go
@@ -48,7 +48,7 @@ func TestServerApp(t *testing.T) {
 	// add comment
 	client := http.Client{Timeout: 10 * time.Second}
 	defer client.CloseIdleConnections()
-	req, err := http.NewRequest("POST", fmt.Sprintf("http://localhost:%d/api/v1/comment", port),
+	req, err := http.NewRequest("POST", fmt.Sprintf("http://localhost:%d/api/v1/comment?site=remark", port),
 		strings.NewReader(`{"text": "test 123", "locator":{"url": "https://radio-t.com/blah1", "site": "remark"}}`))
 	require.NoError(t, err)
 	req.SetBasicAuth("admin", "password")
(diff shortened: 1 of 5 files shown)


umputun (approved):
lgtm. condition flip is right, the inline comment captures the "why" clearly, and the new `TestRest_matchSiteID` nails all four cases including the previously-bypassable missing-param path. existing test updates are mechanical additions of `?site=remark42` where middleware now requires it — fine.

heads-up for release notes: this is a behaviour change for any non-bundled API consumer that was omitting `?site=`. the bundled frontend and admin UI always send it so no user impact there, but worth a line in the release notes when this goes out.

PR #2047, opened 2026-04-17
test(store): use time.UTC in test fixtures to be timezone-agnostic

paskal:
Pure test hygiene, no production-code change.

Store tests stored timestamps with `time.Local` in their fixtures and asserted equality against engine-returned values that round-trip through UTC. `assert.Equal` compares zone identity, so on UTC runners (CI, most cloud envs) `Local==UTC` and the tests passed; on a developer machine in any other timezone (here BST, UTC+1) `TestService_Put`, `TestService_List`, `TestBoltDB_InfoPost`, `TestBoltDB_InfoList` and several others failed with the same wall-clock numbers but mismatched zones.

Switched all test fixtures to `time.UTC` across:
- `store/comment_test.go`
- `store/formatter_test.go`
- `store/service/service_test.go`
- `store/engine/bolt_test.go`
- `store/engine/engine_test.go`

Production code unchanged.

## Verification

`go test ./store/...` — green on a non-UTC laptop (BST, UTC+1).

### DIFF
--- a/backend/app/store/comment_test.go
+++ b/backend/app/store/comment_test.go
@@ -120,7 +120,7 @@ func TestComment_PrepareUntrusted(t *testing.T) {
 		Score:       10,
 		Pin:         true,
 		Deleted:     true,
-		Timestamp:   time.Date(2018, 1, 1, 9, 30, 0, 0, time.Local),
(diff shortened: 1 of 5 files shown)


umputun (approved):
lgtm, thx

PR #2048, opened 2026-04-17
test: use testing/synctest to eliminate wall-clock sleeps

paskal:
Go 1.25's `testing/synctest` package (GA) provides a fake-clock bubble for deterministic goroutine and timer testing. This converts tests that waited on real-time durations to use synctest, removing most wall-clock `time.Sleep` workarounds.

## Converted (11 tests across 9 files)

- `notify/notify_test.go` — all tests, replaced 17× `time.Sleep(110ms)` with `synctest.Wait()`
- `store/service/service_test.go` — `VoteSameIPWithDuration`, `UserReplies`, `submitImages`, `ResubmitStagingImages`, `deleteImagesOnCommentDelete`
- `store/image/{image,bolt_store}_test.go` — `Cleanup`, `Submit`, `SubmitDelay`
- `store/engine/bolt_test.go` — `FlagListBlocked`
- `providers/telegram_test.go` — `DispatchTelegramUpdates`
- `migrator/backup_test.go` — `TestBackup_Do`
- `_example/memory_store/accessor/data_test.go` — `FlagListBlocked`

## Simplifications

- `notify/notify_mock.go`: dropped the 10ms `time.After` delay and `ctx.Done` select in `MockDest`. The artificial I/O simulation is pointless and actively blocked `synctest.Wait` from draining the queue.
- Removed three dead-code `time.Sleep(1s)` calls in `EditCommentDurationFailed`, `EditCommentAdmin`, and `Info` tests — prepopulated comments from 2017 already exceed any `EditDuration`/`ReadOnlyAge` under real clock, so the sleeps were meaningless.
- `UserReplies`: replaced the `Eventually` + `Sleep` + mutex polling with a direct `time.Sleep` under the fake clock and a single equality check.

## Skipped (incompatible with synctest)

- `fs_store_test.go`: relies on OS file mtime (real wall clock)
- `rss_test.go`: needs real wall-clock second boundary for `pubDate`
- `admin_test.go`, `rest_private_test.go`, `rest_public_test.go`: `httptest` network I/O
- `cmd/server_test.go`: real HTTP server startup polling

## Quirks worth documenting

- `synctest.Wait()` does **not** advance fake time, despite what one might expect. It only returns once all other bubble goroutines are durably blocked. To advance the fake clock, the test goroutine must itself call `time.Sleep`.
- BoltDB keys the `last` bucket by `comment.Timestamp` nanosecond string. Rapid `b.Create` calls under frozen fake time produce identical keys and overwrite each other. `TestService_UserReplies` adds `time.Sleep(time.Nanosecond)` between `Create`s to advance the clock.
- `Bolt.Cleanup` uses strict `age > ttl`. Under fake time the age/ttl delta is exactly zero at the boundary, so `TestBoltStore_Cleanup` subtracts 1ms from the passed ttl to stay strictly under.

## Not in scope

The four pre-existing test failures on master (`TestBoltDB_InfoPost`, `TestBoltDB_InfoList`, `TestService_Put`, `TestService_List` — all GMT/UTC timezone mismatches) are unchanged and outside this refactor.

### DIFF
--- a/backend/_example/memory_store/accessor/data_test.go
+++ b/backend/_example/memory_store/accessor/data_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"sort"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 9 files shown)


umputun (commented):
lgtm on the refactor itself — tests pass, race-clean, no flakiness in 50+ iterations of the drop tests.

two things before merge:

1. **rebase needed.** branch is behind master by 2 commits (#2046 and #2047 landed after you opened this). `service_test.go` will conflict with #2047's UTC fixtures since both PRs touch the same fixture lines. also please drop the "pre-existing test failures (TestBoltDB_InfoPost etc)" paragraph from the description — #2047 just fixed exactly those.

2. **Copilot's drop-test concern is a false alarm** — `Submit()` at `notify.go:96-100` is a non-blocking send on a buffered channel with `default:` branch, so overflow is independent of MockDest latency. under synctest, three back-to-back non-blocking sends execute before the consumer is scheduled, and drops happen deterministically. no action needed, just noting in case you were weighing it.

minor, not a blocker: the `time.Sleep(time.Nanosecond)` workaround in `TestService_UserReplies` works but it points at a real latent fragility in bolt's `last` bucket — keys collide on identical ns timestamps and silently overwrite. production can't hit it (<1ns apart HTTP requests), but worth filing as a followup to key that bucket on something stronger than ns format.

umputun:
@copilot resolve the merge conflicts in this pull request

umputun (approved):
lgtm, thx

PR #2049, opened 2026-04-21
fix(auth): close OAuth open-redirect by wiring AllowedRedirectHosts

paskal:
Closes the OAuth open-redirect audit finding by bumping `go-pkgz/auth/v2` to the version carrying [go-pkgz/auth#275](https://github.com/go-pkgz/auth/pull/275) and wiring the new `AllowedRedirectHosts` opt-in.

Fifth and final PR from the security audit. Sister PRs: #2044 (SSRF + lint), #2045 (path traversal), #2046 (matchSiteID), #2047 (TZ tests) — all merged.

## Fix

`cmd/server.go:getAuthenticator` now passes an `AllowedRedirectHosts` that returns the operator's existing `--allowed-hosts` config (minus the CSP `self` sentinel, which is not a real hostname). Per the library contract, `Opts.URL`'s host is always implicit, so a default single-site deployment gains the protection with no config change. Multi-host embeds keep working as soon as their hosts are in `AllowedHosts` — they already have to be there for the CSP `frame-ancestors` header.

## Reproduction — local dev-auth instance

Before the fix (dep bumped, `AllowedRedirectHosts` nil):

```
GET /auth/dev/login?from=https://evil.example.com/phish&site=remark
[redirect 1] -> http://localhost:8084/login/oauth/authorize?...
POST .../authorize (username=testuser)
[redirect 1] -> http://localhost:18080/auth/dev/callback?code=...
[redirect 2] -> https://evil.example.com/phish        <-- 307 to attacker host
```

After the fix:

```
GET /auth/dev/login?from=https://evil.example.com/phish&site=remark
[redirect 1] -> http://localhost:8084/login/oauth/authorize?...
POST .../authorize (username=testuser)
[redirect 1] -> http://localhost:18080/auth/dev/callback?code=...
-> final status: 200  (user-info JSON, no Location header)
```

Server log on rejection — host only, no path/query leak:

```
[WARN] rejected redirect to disallowed host: evil.example.com
```

## Diff

| File | Change |
|---|---|
| `backend/app/cmd/server.go` | Set `AllowedRedirectHosts` in `auth.Opts` |
| `backend/go.mod` / `go.sum` | Bump `go-pkgz/auth/v2` to `v2.1.2-0.20260421203319-686683f19cf7` |
| `backend/vendor/github.com/go-pkgz/auth/v2/**` | Vendor refresh (includes new `provider/redirect.go`) |

## Verification

- `cd backend/app && go test -race ./...` — all 15 packages pass
- `cd backend/_example/memory_store && go build -race ./... && go test -race ./...` — builds and passes
- Local attack reproduction above


paskal:
Pushed fixup `808806f0` addressing both Copilot inline points:

- **server.go:1364 — entry normalisation.** Extracted `getAllowedRedirectHosts` next to the existing `getAllowedDomains`. It strips schemes, paths and ports via `url.Parse`, drops the `'self'` / `"self"` CSP sentinels, and skips wildcard entries (`*`, `*.cdn.example.com`) that are valid CSP source expressions but not valid hostnames. Parse failures log a warning rather than poisoning the allowlist.
- **server.go:1366 — unit test.** New `Test_getAllowedRedirectHosts` table covers bare hostnames, scheme stripping (http/https), paths, host:port, scheme:port, all three `self` spellings, wildcards, empty entries, and a mixed real-world combination.

The `getAuthenticator` closure now just calls the helper, so the wiring stays one line and the logic is testable.

paskal:
Pushed fixup `e5f3cc24` addressing the three new Copilot points:

* **server.go:823 — port preservation.** `getAllowedRedirectHosts` now keeps the explicit port: `example.com:8080` and `https://example.com:8443` are emitted as `example.com:8080` / `example.com:8443`. Bare hostnames stay bare (any-port match). The auth validator checks both `Hostname()` and `Host`, so this gives operators the choice: list the host alone for any-port, or `host:port` to lock it down.
* **server_test.go:922 — adjusted.** Three test cases updated; the mixed-real-world case now asserts `admin.example.com:8443` is preserved verbatim.
* **fs_store.go:45 — nolint rationale fixed.** Save is reached via `image.Service.Save` (id from `CachedImgID` hash) and `SaveWithID` (image proxy cache, same hash flow). Neither is HTTP validation — the comment now names the actual invariant: id is a server-generated hash, never user input.

umputun (approved):
lgtm. vendor refresh is clean, tests/lint/race all pass, Copilot points all addressed, and the port-preservation matches the upstream validator's dual host/host:port match.

Couple of optional nits, none blocking:

1. `server_test.go:913` and `918` — `"https scheme stripped"` and `"scheme without port stays bare host"` have the same input/output (`https://example.com` → `example.com`). One is dead weight.

2. No coverage for the warn/skip branch at `server.go:820` (`u.Hostname() == ""`, e.g. `https://`). Also no IPv6 cases — relevant because you emit `u.Host` verbatim for port-ful entries, so bracketed `[::1]:8443` is worth a row in the table.

3. `server.go:820-821` — the warn fires on `err != nil || u.Hostname() == ""`, so when the hostname is empty but parse succeeded the `%v` logs `err=<nil>`. Minor cosmetic.

4. `getAllowedRedirectHosts` and `getAllowedDomains` now share the same `https://`-prefix + `url.Parse` + `'self'`-drop pattern. Could be factored into a shared normalizer in a follow-up, not worth doing here.

fs_store.go nolint tweak is fine in this PR — it was an inaccurate rationale from the earlier audit commits, fits the series.

PR #2071, opened 2026-05-22
ci(site): skip BuildKit cache on release events so the new version reaches the live site

paskal:
## Summary

The site embeds the latest remark42 version at build time — `site/src/data/github.js` fetches it from the GitHub releases API and Eleventy bakes it into every page (the version badge in the header). The ci-site.yml workflow already triggers on `release: published`, but the build step uses `cache-from: type=gha,scope=site-<platform>`, so on a release-triggered run — where no `site/**` file has changed since the last master build — Buildx reuses the cached `yarn build` layer and ships the previous tag's version.

v1.16.0 is the concrete case: the release workflow ran and succeeded at 2026-05-22T18:29:56Z, but the deployed site kept showing v1.15.0 because the cached layer was hit.

## Fix

Drop `cache-from` only for `release` events; `cache-to` keeps populating so the next non-release master build still benefits from the cache.

```yaml
cache-from: ${{ github.event_name == 'release' && '' || format('type=gha,scope=site-{0}', matrix.platform) }}
```

Targeted at the smallest blast radius — every other trigger path (master push, PR) behaves exactly as before.

## Why not Option B (build-arg)

A more surgical alternative is to pass the release tag as a build arg and have the data file prefer it over the GitHub API call. That avoids both the cache hit AND any GitHub API propagation race (the workflow fired 2 seconds after release publication; even with a fresh build, the API might still have returned the previous tag at that exact moment). I went with the simple cache-disable here because release events are rare (handful per year), the cold rebuild adds a couple of minutes, and the change has obvious correctness — no extra coupling between Dockerfile, workflow inputs, and data file. If the cold-build cost ever bites, we can revisit and switch to the build-arg approach.

## Verification

- Diff is a single line under a 9-line explanatory comment.
- Non-release branches: `cache-from` resolves to the same `type=gha,scope=site-<platform>` value as before.
- Release events: `cache-from` resolves to empty string → Buildx skips the import step.
- `cache-to` unchanged, so the cache layer continues to populate.

### DIFF
--- a/.github/workflows/ci-site.yml
+++ b/.github/workflows/ci-site.yml
@@ -53,13 +53,22 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.PKG_TOKEN }}
 
+      # cache-from is skipped on release events: the site embeds the latest
+      # remark42 version at build time (site/src/data/github.js fetches it from
+      # the GitHub releases API and Eleventy bakes it into the HTML). On a
+      # release-triggered run no site/** file has changed since the last
+      # master build, so Buildx would reuse the cached "yarn build" layer and
+      # ship the previous tag's version. Discarding the cache only for release
+      # events forces a fresh build that picks up the new tag; non-release
+      # runs keep the cache for fast iteration. cache-to is still populated
+      # so the next non-release run reuses what we just built.
       - name: build and push by digest
         id: build
         uses: docker/build-push-action@v7
         with:
           context: ./site
           platforms: ${{ matrix.platform }}
-          cache-from: type=gha,scope=site-${{ matrix.platform }}
+          cache-from: ${{ github.event_name == 'release' && '' || format('type=gha,scope=site-{0}', matrix.platform) }}
           cache-to: type=gha,scope=site-${{ matrix.platform }},mode=max
           outputs: type=image,name=ghcr.io/umputun/remark42-site,push-by-digest=true,name-canonical=true,push=true
 


umputun:
two things before this merges.

**1. the cache-skip expression is a no-op (Copilot is right).**

`github.event_name == 'release' && '' || format(...)` falls into the classic GHA ternary trap: on a release event it evaluates `true && ''` -> `''`, then `'' || format(...)` -> since `''` is falsy, `||` returns `format(...)`. So `cache-from` resolves to the formatted cache string on release events too, and the cache is never actually skipped. Invert it as Copilot suggests:

```yaml
cache-from: ${{ github.event_name != 'release' && format('type=gha,scope=site-{0}', matrix.platform) || '' }}
```

**2. even fixed, this won't reach the live site.**

remark42.com runs `ghcr.io/umputun/remark42-site:master` (per the server compose). Release events build `:v1.16.0` and `:latest` only, never `:master` (the manifest step keys off `refs/tags/*`). So a corrected #2071 gives a fresh `:latest`, but the deployed `:master` stays stale.

verified on the v1.16.0 case: ran `ghcr.io/umputun/remark42-site:latest` locally, version badge still reads v1.15.0 (cache hit, as described) - and that image isn't what the site serves anyway.

to actually get a release version onto remark42.com, either point the `remark42-site` compose service at `:latest`, or have release builds also push `:master`. The cache fix alone isn't enough.

umputun:
follow-up - how v1.16.0 got onto remark42.com in the meantime, and a step back on the approach.

**the immediate fix.** pushed `e8b9d700` to master - a real doc fix (the kubernetes manual still pinned `remark42:v1.14.0`, two releases stale). That change lives under `site/`, so `COPY ./ /site` in the site Dockerfile got a different checksum, the `yarn build` layer cache-missed, `github.js` re-fetched, and v1.16.0 baked into a fresh `:master`. Master push -> deploy job fired -> updater pulled `:master`. Site now shows v1.16.0.

worth noting why that worked where a release rebuild didn't: any genuine `site/` content change invalidates the `COPY` layer naturally. The cache only goes stale when nothing in the build context changed - exactly the release-event case this PR targets, and also a push that touches only `ci-site.yml` (not in the `./site` context).

**stepping back.** even with the expression fixed, this only de-caches `release`-event builds, which produce `:latest`/`:v1.16.0` - not `:master`, which is what the site serves. So the cache fix alone still wouldn't move the live site.

the deeper thing: embedding "latest release" into static HTML at build time is fragile by design. It depends on every release reliably forcing a no-cache rebuild of the *deployed* tag, and on winning a race with GitHub API propagation (the workflow fired ~2s after publish). The build-arg option mentioned in the description removes the API race but still needs the rebuild. The durable fix is to drop build-time embedding entirely - fetch the version client-side on page load. No rebuild, no cache, no race, always current. Only cost is an anonymous GitHub API call per visit, and if traffic makes rate limits a concern, a tiny cached endpoint or worker covers it.

so my suggestion: either take this PR further (fix the expression, and make release builds also tag/deploy `:master`, or point the site at `:latest`), or skip the CI patch and move the version lookup client-side. The one-line cache toggle on its own won't fix what the title promises.

paskal:
Closing in favor of the client-side fetch approach per review feedback.

The cache-skip expression was a no-op (GHA ternary trap — Copilot was right); even with the expression inverted, release builds tag `:v1.16.0`/`:latest` but the site serves `:master`, so the version still wouldn't reach the live site without further changes to the manifest step. And the underlying design — embedding the latest release into static HTML at build time — is the source of the fragility: it races GitHub API propagation, depends on every release reliably forcing a no-cache rebuild of the deployed tag, and recreates the same problem on any push that doesn't touch `./site` (workflow-only changes, etc.).

Replacing with a small client-side fetch in #NEXT — drops `site/src/data/github.js`, replaces `{{ github.latestVersion }}` in the header with `<span data-remark42-version>` populated by an inline-bootstrap fetch of `/releases/latest`. No rebuild, no cache, no API race, no tag mismatch. GitHub's response already carries `Cache-Control: public, max-age=60`, so the per-visitor cost is bounded.

PR #2072, opened 2026-05-23
site: fetch latest version client-side instead of embedding at build time

paskal:
## Why

The version badge in the header (`site/src/includes/components/header.njk`) was filled in by `site/src/data/github.js` calling the GitHub releases API at Eleventy build time and baking `data[0].tag_name` into every page. Three failure modes followed from that design:

1. **Layered cache.** Buildx caches the `yarn build` layer in `ci-site.yml`; on a release-triggered workflow nothing under `./site` changes, so the cached HTML (with the previous tag baked in) gets shipped. v1.16.0 went out and remark42.com kept showing v1.15.0 until a separate `site/` commit landed and naturally invalidated the `COPY` layer.
2. **Tag mismatch.** The deploy pulls `ghcr.io/umputun/remark42-site:master`, but release events build `:v1.16.0` and `:latest` only. A cache-skip workflow tweak wouldn't even reach the served image.
3. **API propagation race.** The workflow fires ~2s after release publish, so even with cache disabled the API might still return the previous tag from a stale read replica.

All three vanish if the version is fetched in the browser instead. Per #2071 review feedback (cache-skip approach was incorrect on both expression and routing). Closes the design problem at its root rather than patching one symptom.

## What

`fetch('{{ site.githubApiUrl }}/releases/latest')` in `inline.njk` (which already runs in the `<head>` for theme detection), updates any `[data-remark42-version]` element with `tag_name`. The placeholder span is `hidden` in the template so a failed/blocked fetch leaves no stray gap.

Cached in `sessionStorage` with a 1h TTL — repeated navigations within a tab don't re-hit the GitHub API, and a stale display tops out at one hour for very long-lived tabs.

DOM update is gated on `DOMContentLoaded` (script loads synchronously in `<head>`, so a cache-hit fetch can resolve before `<body>` is parsed). `console.warn` on fetch failure (rate-limit / offline / blocked) — same signal the old build-time `github.js` produced.

## Cost

GitHub serves `/releases/latest` with `Cache-Control: public, max-age=60`, so the per-visitor browser cache alone bounds the rate. On top of that, sessionStorage means at most one fetch per browser tab per hour. Unauthenticated rate limit is 60/hour/IP — well within normal traffic from any single visitor IP.

## Changes

- `header.njk` — `{{ github.latestVersion }}` → `<span data-remark42-version hidden></span>` placeholder. aria-label of the anchor changed from "Repository" → "releases" to match the link target.
- `inline.njk` (renamed from `inline.js` for nunjucks templating) — adds the fetch + sessionStorage cache + DOM-update logic alongside the existing theme detection. API endpoint templated from `site.githubApiUrl`.
- `site/src/data/site.json` — adds `githubApiUrl` so the API endpoint is in one place, derived/changeable alongside `githubUrl`.
- `site/src/data/github.js` — deleted (no longer used).
- `package.json` + `yarn.lock` — drop `node-fetch` from devDependencies (was only used by `github.js`).

## Verification

- `yarn install` succeeds; lockfile cleaned.
- `yarn build` succeeds; output `build/*.html` contains the placeholder span and the inline fetch on all pages spot-checked (`/`, `/demo/`, `/docs/getting-started/installation/`).
- Templated URL renders as the literal `fetch('https://api.github.com/repos/umputun/remark42/releases/latest')` in the built output — verified.
- No orphan `{{ github.* }}` template refs remain.


umputun (commented):
both copilot findings are real:

1. **`inline.js:17` race**: `head.njk` includes `inline.js` synchronously (no `defer`/`async`), so the fetch fires mid-head-parse. Cache hits or very-fast responses can resolve before `[data-remark42-version]` is in the DOM, then the badge stays empty. Gate the DOM update on `document.readyState === 'loading' ? DOMContentLoaded : run-now`.

2. **`header.njk:13` gap**: `mr-2` is unconditional, so the 0.5rem gap before the github icon is visible during fetch and on any failure (rate limit, offline, blocked). Either hide the span by default (`hidden` class) and unhide on success, or drop `mr-2` and have JS add it together with `textContent`.

minor: silent `.catch(() => {})` leaves no console signal on rate-limit/block. The old `github.js` logged on error. Trade-off, your call.

otherwise clean.

paskal:
All three Copilot points addressed in 7697dcf3:

1. **Hard-coded API URL** → renamed `inline.js` → `inline.njk` so nunjucks evaluates the include; added `githubApiUrl` to `site.json`; the fetch URL is now `{{ site.githubApiUrl }}/releases/latest`. One place to update if the repo ever moves.
2. **aria-label / link-target mismatch** → label changed to `"{{ site.name }} releases on GitHub"` so screen readers describe the actual destination.
3. **Silent fallback vs `console.warn`** → kept `console.warn` (intentional, after your earlier note about the old `github.js` logging on error). Updated the PR description to match.

paskal:
You're right — my earlier commit added `githubApiUrl` and renamed the file but never actually changed the `fetch()` line. The build output looked right because the hard-coded URL happens to match what `{{ site.githubApiUrl }}` expands to, so I didn't catch it. Fixed in c93a5175.

paskal:
Fair — fixed in cef18b9d.

umputun (approved):
all three round-1 items addressed - `readyState` gate on the DOM update, span `hidden` by default and unhidden on success, `console.warn` kept. The fetch reads from `site.githubApiUrl` now, and `github.js` + `node-fetch` are fully removed, lgtm

PR #2075, opened 2026-05-26
docs: close backtick in smtp.login_auth default cell

### LINKED ISSUE #2074: The doc page layout is broken
The layout is broken at the bottom of the table with parameters on https://remark42.com/docs/configuration/parameters/

paskal:
## Why

The parameters table on https://remark42.com/docs/configuration/parameters/ stopped rendering correctly partway through the SMTP section — roughly 40 rows from `smtp.login_auth` down to `dbg/DEBUG` appeared as raw pipe-delimited text instead of HTML table cells.

Cause: the `Default` column for `smtp.login_auth` had `` `false `` with an opening backtick but no closing one. kramdown saw an unclosed code span, gave up on the cell, and from that point on treated the rest of the table as literal text.

Introduced in cebe9291 — first shipped in **v1.11.0** (Oct 2022), so it's been broken for ~3.5 years.

Closes #2074

## What

One-character fix — add the missing backtick to close the code span.


umputun (approved):
lgtm, thx

PR #2077, opened 2026-05-26
fix: parameter docs + --help text inconsistencies (audit)

paskal:
## Why

Doc audit against the backend Go flag tags surfaced a small batch of inconsistencies — wrong defaults, a typo that broke a table column, a paragraph wedged inside a table that terminated rendering of the second half, and shuffled `--help` strings on deprecated flags. Fixing them as a single pass.

## Docs (`site/src/docs/configuration/parameters/index.md`)

- **`image.bolt.file`** default — was `` `/var/pictures.db` `` (looked like an absolute system path), actual default is `./var/pictures.db` (relative to the working dir, per `server.go:202`).
- **`notify.webhook.template`** default — the docs row was both wrong (claimed default `` `{"text": "{{.Text}}"}` ``) and broken (the unescaped `|` inside the cell collapsed the Description column). Updated to the correct effective default `{"text": {{.Text | escapeJSONString}}}` with `\|` to escape the table column separator.
- **"Custom OAuth2 integration currently supports only one custom provider"** — was a free-standing paragraph wedged between two table rows. kramdown terminated the main parameters table on that paragraph and restarted a new headerless table for the remaining 60+ rows. Moved the paragraph to its own `### Custom OAuth2 integration` subsection below the table so the table stays contiguous.

## Backend `--help` text (`backend/app/cmd/server.go`)

- **`allowed-hosts`** — description ended with a stray double apostrophe: `CSP 'frame-ancestors''`. Dropped the trailing `'`.
- **Deprecated `auth.email.{port,passwd,user,tls}` descriptions** were shuffled:
  - `port` said *"SMTP password"*
  - `passwd` said *"SMTP port"*
  - `user` said *"enable TLS"*
  - `tls` said *"SMTP TCP connection timeout"*

  Each now matches the flag it's actually describing. (`auth.email.timeout` was the only one already correct.) The docs table already had the correct descriptions for these deprecated flags — only `--help` was wrong.
- **`notify.webhook.template` description and headers description** — said "webhook *authentication* template / authentication headers" but the template is the payload body (not auth-related). Reworded to "payload template" / "webhook headers".

## Behaviour change ⚠️

**`notify.webhook.template` default switches from the literal `{"text": "{{.Text}}"}` to `{"text": {{.Text | escapeJSONString}}}`.**

`server.go:286` had `default:"{\"text\": \"{{.Text}}\"}"`. go-flags applies the default tag at parse time, so `Notify.Webhook.Template` was never empty when the user omitted the flag — and that meant the safer `webhookDefaultTemplate` fallback in `backend/app/notify/webhook.go:50` (which uses `escapeJSONString`) was dead code. The unescaped literal default produces malformed JSON whenever a comment contains a `"`, newline, or backslash.

This PR drops the `default:` tag so the package-level fallback now takes effect. For comments without special characters the payload is byte-identical; for comments with special characters the new payload is valid JSON where the old one was malformed. Webhook receivers that previously failed to parse such payloads will now succeed.

If a deployment explicitly *relied* on the unescaped form (e.g. their receiver tolerated malformed JSON or did string-matching), they can restore the prior behaviour by setting `NOTIFY_WEBHOOK_TEMPLATE='{"text": "{{.Text}}"}'`.

## Verification

- `go test ./app/cmd/ ./app/notify/` — passes.
- `golangci-lint run ./app/cmd/` — 0 issues.
- `go build ./...` in `backend/_example/memory_store` — clean.
- `yarn build` in `site/` — clean; rebuilt HTML shows the smtp.login_auth row + every subsequent row rendering as proper `<td>` cells (157 `<tr>` total across the two tables, up from the truncated render before). The `notify.webhook.template` cell renders with the pipe inside `<code>` as expected.

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -90,7 +90,7 @@ type ServerCommand struct {
 	EnableEmoji                bool          `long:"emoji" env:"EMOJI" description:"enable emoji"`
 	SimpleView                 bool          `long:"simple-view" env:"SIMPLE_VIEW" description:"minimal comment editor mode"`
 	ProxyCORS                  bool          `long:"proxy-cors" env:"PROXY_CORS" description:"disable internal CORS and delegate it to proxy"`
-	AllowedHosts               []string      `long:"allowed-hosts" env:"ALLOWED_HOSTS" description:"limit hosts/sources allowed to embed comments via CSP 'frame-ancestors''" env-delim:","`
+	AllowedHosts               []string      `long:"allowed-hosts" env:"ALLOWED_HOSTS" description:"limit hosts/sources allowed to embed comments via CSP 'frame-ancestors'" env-delim:","`
 	SubscribersOnly            bool          `long:"subscribers-only" env:"SUBSCRIBERS_ONLY" description:"enable commenting only for Patreon subscribers"`
 	DisableSignature           bool          `long:"disable-signature" env:"DISABLE_SIGNATURE" description:"disable server signature in headers"`
 	DisableFancyTextFormatting bool          `long:"disable-fancy-text-formatting" env:"DISABLE_FANCY_TEXT_FORMATTING" description:"disable fancy comments text formatting (replacement of quotes, dashes, fractions, etc)"`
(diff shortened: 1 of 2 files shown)


paskal:
Good catch — you were right, my audit was wrong. Addressed in 9e572fec:

server.go:286 had `default:"{\"text\": \"{{.Text}}\"}"` (JSON-unsafe literal). That overrode the safer `webhookDefaultTemplate` fallback in `webhook.go:50`, which uses `escapeJSONString` — so the flag's default tag was effectively hiding the safe fallback the notify package intended.

Fix: dropped the unsafe default tag so the `webhook.go` fallback now takes effect at runtime. The actual effective default is `{"text": {{.Text | escapeJSONString}}}`. Updated `parameters/index.md` to document that (with `\|` to escape the kramdown table column separator), aligned the `--help` description, and adjusted the headers description to drop the misleading "authentication" wording.

`go test ./app/cmd/ ./app/notify/` — passes.

paskal:
You're right that it's a user-visible behaviour change, not a pure doc/help fix — calling that out explicitly was missing from the PR description. Updated the description with a dedicated "Behaviour change ⚠️" section explaining the switch, why the prior tag was effectively a bug (it shadowed the package's own safer fallback), what payload differences to expect, and the one-line escape hatch for anyone who relied on the old literal: `NOTIFY_WEBHOOK_TEMPLATE='{"text": "{{.Text}}"}'`.

Intent is to fix; happy to revert if @umputun prefers preserving the literal default for backward-compat instead. Resolving for now — please flip if you want it the other way.

umputun (approved):
lgtm. Verified the webhook default change myself. The `default:` tag at server.go:286 was shadowing the `escapeJSONString` fallback in webhook.go:47-49, so dropping it is the right call. Plain comments stay byte-identical, comments with quotes/newlines now produce valid JSON instead of malformed. Fine to switch the default.

help/doc fixes all match the flag defs now, and the parameters table lines up with notifications/index.md:49 and the runtime fallback, so the cross-doc nit copilot raised is resolved too.

PR #2106, opened 2026-07-03
Fix dropped notification errors and switch to errors.Join

paskal:
Rescoped to a single logical change. The CI and compose cleanups that were originally bundled here now live in #2107 and #2108.

## Bug: only the last failing recipient's error was reported

`notify/email.go` accumulated errors with `multierror.Append(fmt.Errorf(...))`, omitting the `result` accumulator as the first argument. Each iteration overwrote the accumulator with a fresh single-error value, so when several recipient emails failed only the last failure surfaced; earlier ones were silently dropped. The telegram notifier did it correctly.

## Fix

Rather than just correcting the argument, this replaces `hashicorp/go-multierror` with the stdlib `errors.Join` (Go 1.20+) everywhere it was used: `notify/email.go`, `notify/telegram.go`, `rest/api/rest_private.go`, `store/service/service.go`, `store/image/image.go` and `store/engine/bolt.go` (the last already used `errors.Join`, so the codebase was mid-migration). This fixes the bug as a side effect and drops the direct dependency; it stays indirect because `go-pkgz/lcw/v2` still imports it upstream.

A regression test in `email_test.go` sends two failing recipients and asserts both errors are reported.

The user-visible error string format changes from go-multierror's bullet list to newline-joined messages; the affected tests are updated and no code depends on the old format.

### DIFF
--- a/backend/_example/memory_store/go.mod
+++ b/backend/_example/memory_store/go.mod
@@ -21,8 +21,6 @@ require (
 	github.com/go-pkgz/rest v1.22.0 // indirect
 	github.com/go-pkgz/routegroup v1.6.0 // indirect
 	github.com/gorilla/css v1.0.1 // indirect
-	github.com/hashicorp/errwrap v1.1.0 // indirect
-	github.com/hashicorp/go-multierror v1.1.1 // indirect
 	github.com/kr/text v0.2.0 // indirect
 	github.com/microcosm-cc/bluemonday v1.0.27 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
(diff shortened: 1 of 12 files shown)


umputun (approved):
reviewed the errors.Join migration and the email multi-recipient fix.

the accumulator bug is fixed - the old `multierror.Append(fmt.Errorf(...))` dropped the `result` accumulator each iteration, so only the last failing recipient surfaced. Now errors are accumulated across both loops and `errors.Join` reports all of them, with a regression test covering two failing recipients.

every `errors.Join` swap keeps the prior nil/aggregation semantics (nil args dropped, nil returned when all nil), the message-format change is contained to the updated tests, and `go-multierror` stays indirect via `lcw/v2`. Build, tests, lint and race pass on backend and the example module.

PR #2107, opened 2026-07-03
Run backend CI on PR updates and freeze frontend lockfile installs

paskal:
Two CI hygiene changes, split out of #2106.

## Backend CI skipped pushed PR updates

`ci-backend.yml` had `pull_request: types: [opened, reopened]`, which excludes `synchronize`. Pushes to an open PR branch therefore did not re-run backend tests, lint or coverage, so a broken follow-up commit could land after the first green run. Removing the `types` filter restores the default activity set (`opened`, `synchronize`, `reopened`) without broadening to every PR event; the workflow stays on `pull_request` with `contents: read`.

## Frontend installs weren't frozen

`ci-frontend.yml` and `ci-frontend-api.yml` ran `pnpm i`, which can mutate the lockfile and produce non-reproducible installs. They now use `pnpm install --frozen-lockfile`, matching `release.yml` and making the intended CI behaviour explicit.

### DIFF
--- a/.github/workflows/ci-backend.yml
+++ b/.github/workflows/ci-backend.yml
@@ -10,7 +10,6 @@ on:
       - "!backend/scripts/**"
       - "!**.md"
   pull_request:
-    types: [opened, reopened]
     paths:
       - ".github/workflows/ci-backend.yml"
       - "backend/**"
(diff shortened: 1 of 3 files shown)


umputun (approved):
reviewed both CI hygiene changes.

dropping `types: [opened, reopened]` from ci-backend.yml restores the default pull_request set (opened, synchronize, reopened), so pushes to an open PR re-run backend tests, lint and coverage. The `paths:` filter is kept, so it still only fires on backend-touching PRs, not on every PR event.

switching `pnpm i` to `pnpm install --frozen-lockfile` across the three frontend workflows makes installs reproducible and matches release.yml. A drifted lockfile now fails CI instead of being silently rewritten.

PR #2108, opened 2026-07-03
Remove obsolete version key from compose files

paskal:
Split out of #2106.

The top-level `version: "2"` field is obsolete in the current Compose file format; Docker Compose ignores it and prints a warning on every invocation. Removed it from `docker-compose.yml`, `compose-dev-backend.yml`, `compose-dev-frontend.yml` and `compose-e2e-test.yml`.

`docker compose config` output is byte-identical to master for all four files, so there is no functional change.

### DIFF
--- a/compose-dev-backend.yml
+++ b/compose-dev-backend.yml
@@ -4,7 +4,6 @@
 #
 # build remark42 docker image - docker compose -f compose-dev-backend.yml build
 # start remark42 service - docker compose -f compose-dev-backend.yml up
-version: "2"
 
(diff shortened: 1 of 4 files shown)


umputun (approved):
checked the four files for any v2-format-specific keys (mem_limit, cpu_shares, volumes_from, extends, top-level networks/volumes) - none present, so removing `version: "2"` can't reinterpret anything that's left. Matches the byte-identical `docker compose config` claim. Safe cleanup.

PR #2109, opened 2026-07-03
Cache per-user flag lookups in comment listings

paskal:
## Problem

`alterComment` issued two `engine.Flag` calls (Blocked, then Verified) for every comment. Because `Find`, `FindSince`, `User` and `Last` all funnel through it, a listing of N comments triggered up to 2N BoltDB read transactions even when many comments shared the same author, which is the common case on a busy thread.

## Change

Added a `userFlagCache` that memoises blocked/verified results keyed by `{siteID, userID}` for the duration of a single listing, so a repeated author is looked up once. `alterComment` keeps its signature for single-comment callers (`Get`) by using a fresh cache; the batch paths (`FindSince`, `alterComments`) share one.

Flag-read errors are deliberately not cached: the current comment is treated as not blocked/verified (matching the previous per-comment behaviour) and the next comment for that user retries, so a transient read failure can't pin stale state across the whole listing.

## Tests

`TestService_alterCommentsFlagCaching` covers three cases: repeated user collapses 5 comments to 2 lookups, distinct users are looked up per user, and an errored lookup is retried rather than cached. The existing `TestService_alterComment` (single-comment path) is unchanged and still passes.

### DIFF
--- a/backend/app/store/service/service.go
+++ b/backend/app/store/service/service.go
@@ -128,6 +128,7 @@ func (s *DataStore) FindSince(locator store.Locator, sortMethod string, user sto
 	}
 
 	changedSort := false
+	flags := s.newUserFlagCache()
 	// sets votes controversy for comments added prior to #274
 	// also sanitizes locator.URL for comments added prior to #927
 	for i, c := range comments {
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm. One thing worth confirming since the description doesn't mention it: a user's blocked/verified state is now a per-listing snapshot - if an admin blocks or verifies mid-listing, every comment by that user renders one consistent value instead of old-for-earlier / new-for-later. Reads as more correct, just want to confirm it's intended.

paskal:
It's intentional, changes mid-listing are unlikely.

PR #2112, opened 2026-07-03
Fix flaky TestPublic_FindCommentsCtrl_ConsistentCount

paskal:
`TestPublic_FindCommentsCtrl_ConsistentCount` chose the expected HTTP status with `strings.Contains(tc.params, "=bad")`. Comment IDs in the fixture are random UUIDs, so when one happened to start with `bad` (e.g. `offset_id=bad49e60-...`), the params string contained `=bad` and the case was wrongly expected to return 400 while the handler correctly returned 200. That failed the whole suite roughly 0.2% of the time, independent of the code under test.

Identify bad-request cases by their error response body (`"error":`) instead, which is deterministic per table row and independent of the generated IDs. This matches exactly the intended invalid-parameter rows (`limit=bad`, `offset_id=bad`, in both plain and tree formats) and no valid row.

Surfaced when it failed the validate check on an unrelated PR.


umputun (approved):
lgtm. `"error":` in `expectedBody` matches exactly the four invalid rows and no valid row in the table, so the status check is deterministic now instead of hostage to a UUID that happens to start with `bad`.

PR #2116, opened 2026-07-04
fix: add --trusted-proxy to gate client-IP forwarding headers

paskal:
## Problem

Per-IP rate limiting and (with `--votes-ip`) vote de-duplication key on the client IP, which is recovered from forwarding headers (`X-Real-IP` / `X-Forwarded-For` / `CF-Connecting-IP`) when Remark42 runs behind a reverse proxy. Those headers were accepted from **any** client, so a caller could set them to change its apparent IP and evade per-IP controls.

## Change

- New `--trusted-proxy` / `TRUSTED_PROXY` (comma-separated CIDR/IP). The forwarding headers are honored **only** when the direct TCP peer is a trusted proxy; requests from any other peer keep their real socket address.
- When unset, headers are trusted from any client as before (so existing deployments don't break on upgrade) and a **startup WARN** is logged recommending the flag.
- Wired as a global middleware before the rate limiter / IP consumers; a bad CIDR fails startup.

## Docs

A "Trusted proxies and client IP" section in the parameters reference with a security note, per-topology recommendations (incl. `172.16.0.0/12` for a same-compose Docker proxy), the requirement that the proxy overwrite `X-Real-IP`, plus a note in the nginx manual.

## Tests

`realIPMiddleware` (trusted peer → header honored; **untrusted peer → spoofed header ignored**) and `ParseTrustedProxies` (CIDR, bare v4/v6, blanks, malformed → error). `go test -race`, gofmt, vet, golangci-lint clean. Reviewed with codex (no bypass found; two follow-ups folded in).

## Credit

Reported privately by [@de3erve-hunter](https://github.com/de3erve-hunter) via GitHub Security Advisory GHSA-56x6-q882-mf27. Thanks a lot for the detailed report and proof of concept.

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -85,6 +85,7 @@ type ServerCommand struct {
 	Address                    string        `long:"address" env:"REMARK_ADDRESS" default:"" description:"listening address"`
 	WebRoot                    string        `long:"web-root" env:"REMARK_WEB_ROOT" default:"./web" description:"web root directory"`
 	UpdateLimit                float64       `long:"update-limit" env:"UPDATE_LIMIT" default:"0.5" description:"updates/sec limit"`
+	TrustedProxies             []string      `long:"trusted-proxy" env:"TRUSTED_PROXY" description:"reverse-proxy networks (CIDR or IP) whose X-Real-IP/X-Forwarded-For headers set the client IP; if unset, these headers are trusted from any client (see docs)" env-delim:","`
 	RestrictedWords            []string      `long:"restricted-words" env:"RESTRICTED_WORDS" description:"words prohibited to use in comments" env-delim:","`
 	RestrictedNames            []string      `long:"restricted-names" env:"RESTRICTED_NAMES" description:"names prohibited to use by user" env-delim:","`
 	EnableEmoji                bool          `long:"emoji" env:"EMOJI" description:"enable emoji"`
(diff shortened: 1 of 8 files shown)


paskal:
Copilot's comments were against the pre-review-round commit. Addressed in the latest push: the untrusted path now strips the forwarding headers and pins `RemoteAddr` to the bare socket IP (so anon-id / vote-dedup / logs can't be fooled and stay port-stable), and the doc comment now says `R.RealIP` to match the code. Went through a codex + subagent review round — findings (untrusted-path normalization, fail-fast placement, v4-mapped-IPv6 parse, proxy-hygiene docs) all folded in.

umputun (approved):
lgtm. A few optional notes, none blocking:

1. `ParseTrustedProxies` accepts `0.0.0.0/0` and `::/0`, which trusts every peer and silently re-opens the bypass, but unlike the unset case it logs no WARN. Worth mirroring the unset-case warning there.

2. the trusted path leans on `R.RealIP` (leftmost public XFF), so it's only safe because the operator sets `X-Real-IP` per the docs. That's correct and well documented. If you ever want it enforced in-code regardless of proxy config, parsing XFF from the right against the trusted CIDRs would do it. Fine as a follow-up, not now.

3. test gaps: the `peer == nil` branch (unparseable RemoteAddr) is never exercised, and there's no trusted-peer + empty-header fallback case. Also `seenAddr`/`seenHdr` are shared closure locals, so the tests would race if anyone adds `t.Parallel()`.

4. nit: the flag description is the longest in the struct and drops `CF-Connecting-IP` even though the code strips it. Could trim it and lean on the docs link, same for the 3-line startup WARN vs the terse `[WARN]` style elsewhere.

PR #2119, opened 2026-07-05
Make notify drop tests deterministic

paskal:
`TestService_WithDrops` and `TestService_SubmitVerificationWithDrops` submitted three items into a size-1 queue and asserted at least one was dropped, relying on the single consumer not draining the queue between submits. `synctest` doesn't fully pin this because `MockDest`'s send path does real logging I/O outside the bubble, so under CI's `-race` scheduling the consumer occasionally drained all three and delivered everything, failing the `<= 2` assertion (~0.2% of CI runs).

Add an optional gate channel to `MockDest` so a destination blocks in `Send`/`SendVerification` until released. The tests now submit one item (the consumer picks it up and blocks on the gate), fill the size-1 queue, submit an overflow item that is deterministically dropped, then release the gate. Assert exactly two delivered instead of "at most two".

Verified with `go test -race -count 100` on both tests (previously the pattern that flaked) plus the full package.

### DIFF
--- a/backend/app/notify/notify_mock.go
+++ b/backend/app/notify/notify_mock.go
@@ -15,10 +15,14 @@ type MockDest struct {
 	id               int
 	closed           bool
 	lock             sync.Mutex
+	block            chan struct{} // if non-nil, Send/SendVerification wait on it before recording, letting tests pin the consumer
 }
 
 // Send mock
 func (m *MockDest) Send(ctx context.Context, r Request) error {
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm. Gating the consumer and closing the gate to release all sends turns the probabilistic `<=2` into an exact `Len==2` plus order, so the drop is deterministic instead of scheduling-dependent.

PR #2120, opened 2026-07-05
Drop armv7 build target and remove dead golangci settings

paskal:
Two small build/tooling cleanups.

## Drop armv7

The published multi-arch Docker manifest is `amd64`+`arm64` only, but GoReleaser and the Makefile `dockerx` target still produced `linux/arm/v7` artifacts with no matching image. Remove the armv7 target (`goarch: arm` + `goarm: 7`, plus the now-redundant `goarch: arm` ignore entries) from `.goreleaser.yml`, and `linux/arm/v7` from the Makefile. The only net effect is linux no longer shipping armv7 binaries; all other platforms are unchanged.

## Remove dead golangci config

`goconst` and `lll` have `settings:` blocks in `backend/.golangci.yml`, but neither linter is in the `enable:` list (config uses `default: none`), so the settings were inert. Removed them; the remaining `gosec`/`gocritic`/`govet`/`misspell` settings are untouched (`golangci-lint config verify` passes).

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -26,26 +26,17 @@ builds:
       - amd64
       - arm64
       - "386"
-      - arm
-    goarm:
-      - "7"
     ignore:
       - goos: darwin
         goarch: "386"
-      - goos: darwin
-        goarch: arm
       - goos: freebsd
         goarch: arm64
       - goos: freebsd
         goarch: "386"
-      - goos: freebsd
-        goarch: arm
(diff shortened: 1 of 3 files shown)


umputun (approved):
lgtm. Confirmed `goconst`/`lll` aren't in the enable list (`default: none`), so those settings were inert and removing them is a no-op; the armv7 drop is consistent across `.goreleaser.yml` and the Makefile.

PR #2122, opened 2026-07-05
Modernise Docker build workflows

paskal:
## ci-build.yml — buildx cache modernisation

Replaced the legacy `actions/cache` + `/tmp/.buildx-cache` local cache (with a manual rotate step) with buildx `type=gha` cache, separate scopes for the main and example images. The two build steps are now `docker/build-push-action@v7` (`load: true`, no push) rather than raw `docker buildx build` — required because `type=gha` needs the GitHub Actions cache URL/token that the action injects automatically but a raw shell `buildx` call does not. This workflow runs on PRs, so the change is exercised by this PR's own CI.

## docker.yml — build each image once

The publish job built each image twice per platform (one `build-push-action` call per registry). It now builds once and pushes the same content-addressed image to both ghcr.io and DockerHub via two `type=image` outputs; the single build digest is identical for both registries, so digest export is simplified. The multi-arch manifest merge is unchanged.

**Validation note:** `docker.yml` only runs on `workflow_run` (master / release tags), so it is **not** exercised by this PR's CI. The single-build/two-output digest handling was verified against the build-push-action and BuildKit docs (BuildKit's image exporter returns one root manifest digest, identical across registries for identical content), and the downstream "verify all digests present" step fails loudly if a digest were missing. Worth watching the first master run after merge.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -30,14 +30,6 @@ jobs:
       - name: set up Docker Buildx
         uses: docker/setup-buildx-action@v4
 
-      - name: expose GitHub Actions cache
-        uses: actions/cache@v6
-        with:
-          path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ github.sha }}
-          restore-keys: |
-            ${{ runner.os }}-buildx-
-
       - name: free disk space
         run: |
           sudo rm -rf /usr/share/dotnet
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm. Traced the digest path: one build pushed by digest to both registries yields the same manifest digest, so collapsing to a single `steps.build.outputs.digest` is correct and the verify-digests step still guards it. The `ci-build` `type=gha` cache swap is already green here.

PR #2124, opened 2026-07-06
Return 400 for export of an unknown site

paskal:
`exportCtrl` mapped **every** export failure to `500 Internal Server Error`, so requesting a backup for a non-existent site (e.g. a wrong `-s`/`--site`) came back as a misleading 500 instead of a client error — inconsistent with the rest of the admin/public API, which returns `400` + `ErrSiteNotFound` for site-lookup failures (`admin.go:188`, `rest_public.go`, `rss.go`).

## Change
- Add an `engine.ErrSiteNotFound` sentinel, wrapped at the bolt db-lookup as `fmt.Errorf("site %q %w", siteID, ErrSiteNotFound)` — chosen so the rendered message stays exactly `site "x" not found` and the existing engine tests are unchanged.
- `exportCtrl` maps it to `400` + `rest.ErrSiteNotFound` via an `exportErrStatus` helper (covering both `mode=file` and `mode=stream`); genuine internal failures (gzip close/write) still return 500.
- The `rpc` store drops typed errors over jrpc, so the helper also matches the `"not found"` message as a fallback — safe because the export path only ever does a site-level lookup (an internal bolt error there is `no bucket … in store`, not `not found`).

## Tests
`TestMigrator_Export` now asserts an unknown site returns `400` + `"code":6` in both file and stream modes; the successful-export and bolt `site "x" not found` message assertions are unchanged.

### DIFF
--- a/backend/app/rest/api/migrator.go
+++ b/backend/app/rest/api/migrator.go
@@ -4,11 +4,13 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -18,6 +20,7 @@ import (
 
 	"github.com/umputun/remark42/backend/app/migrator"
 	"github.com/umputun/remark42/backend/app/rest"
+	"github.com/umputun/remark42/backend/app/store/engine"
 )
 
 // Migrator rest with import and export controllers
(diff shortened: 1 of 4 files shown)


umputun (approved):
lgtm. Confirmed the `%w` wrap renders exactly `site "x" not found`, so only `exportCtrl`'s status mapping changes and no other `db()` caller regresses; the `errors.Is` path is the real guard with the string match as a scoped jrpc backstop.

Issue #1989, opened 2026-01-08
Allow disabling avatar functionality?


geerlingguy:
I'm setting up Remark42 on my personal site (see https://github.com/geerlingguy/jeffgeerling-com/issues/167), and one thing I'm considering to maintain consistency with my old Drupal-based comments, as well as to reduce visual noise, is disabling avatars from displaying.

<img width="532" height="333" alt="Image" src="https://github.com/user-attachments/assets/f8e27187-f861-4c67-954b-3e56c7be4114" />

For anonymous auth especially, I find those default Gravatar icons a bit distracting. The quickest way would be to just use CSS, but is there any way to disable Gravatar/avatar caching entirely? I don't want to waste resources on something I wouldn't display.

paskal:
The round shape comes from CSS in the frontend (`border-radius` on the avatar element), not from Gravatar — Gravatar always returns a square image.

`UseGravatar: true` is currently hardcoded in `backend/app/cmd/server.go:1262` and there's no configuration parameter to disable it. Remark42 caches avatar images locally via the go-pkgz/auth avatar proxy, so Gravatar is only hit once per user rather than on every page load.

For hiding avatars visually, CSS is the only option right now, but since the widget renders in an iframe, you can't target it from the host page's stylesheet. You'd need to either:

1. Fork the frontend and remove the avatar rendering, or
2. Use a custom build with the avatar component hidden

Making Gravatar/avatar support configurable (e.g. an `AVATAR_ENABLED=false` parameter) is a possibility. I need to think about it, maybe we should provide easier frontend alterations than yet another backend flag.

PR #2125, opened 2026-07-07
Simplification and added customization of the web interface

dittnamn:
This is mainly a simplification of the interface with some added options for customization so that you're able to hide features you don't want to use. These options have been added the same way as Simple view is implemented.

The following changes have been made:
- Added dashes and dots to permitted usernames. This allows users to be named Dr. Evil or Britt-Marie.
- Added options NAME_MINLENGTH, NAME_MAXLENGTH for username length as the defaults 3 and 64 may seem extreme for some cases.
- Added option NAME_CHARACTERS to be able to disallow numbers from names to be able to avoid users named like "dude123". This option can be expanded with more options in the future.
- Added option to hide the voting features (HIDE_VOTING), as voting might feel overkill for sites without many comments. The votes aren't removed, so if it's reactivated, old votes will still be there. When hiding is activated, sorting features based on votes are also hidden.
- Added option to hide the hide button from comments (HIDE_HIDE), as visitors might be confused by the feature and hide users by mistake when they intended to hide a single comment. Previously hidden users can still be restored.
- Added option to hide the avatars (HIDE_AVATARS) which makes the interface look cleaner, also see bug #1989.
- Added option to hide the user ID from the profile information to make the interface look cleaner and that the ID is probably of no use for most visitors.
- When both e-mail and anonymous authentication is activated, these are combined into one single form with the e-mail field marked as non-required.
- Removed the header from the login form when only one form provider is active to make the interface look cleaner.
- When anonymous authentication is activated, the Sign in button is changed to 'Sign in or pick name', to make new users feel more welcome to add their own comments without the need to go through a lengthy process.
- Changed the edit countdown to show the edit time in minutes until there's 90 seconds left to reduce the distraction from it. Also added parenthesis around the time to show it belongs to the edit button.
- A Swedish translation has been added, including the newly added lines. Other translations will need updating.

Combined login

Swedish translation - user logged in

Swedish translation - admin logged in

Swedish translation - disabled avatars


paskal:
We'll take a look, but a quick thought - it's always better start with the issue and discuss the problem you're having rather than starting with a PR, as a chance that it will be merged as-is are very thin. Some stuff is low-risk and better be extracted into separate PRs, like SV translation.

dittnamn:
> We'll take a look, but a quick thought - it's always better start with the issue and discuss the problem you're having rather than starting with a PR, as a chance that it will be merged as-is are very thin. Some stuff is low-risk and better be extracted into separate PRs, like SV translation.

Well, this started as some hard-coded changes that I did for myself, but after a while it grew a bit so I figured that others could be helped as well with more customization options. I would say that these changes are pretty small and would benefit others. Because they're mostly optional, most existing users wouldn't even notice them. The combined login could be made optional as well if you feel the need for it, but I think a simpler user interface is an advantage for everyone.

paskal:
Hi there, to be clear - I want to ship 1.17.0 with bugfixes and new translations, improve e2e tests, and then before releasing 1.18.0 switch interface from preact to static html + css + js, https://github.com/umputun/remark42/issues/2173.

I think of porting js roughly as-is first but simplify and clarify CSS right away, and then separately figure out how to make the result as simple as possible.

From this PR I dislike hides frontend options exposed on backend, it feels wrong to pass if frontend hides avatars or not from backend. It should be a line in a frontend. We'll need it after the switch, but as frontend-only options, and I am thinking of revising current options to move simple view to frontend-only option as well (in the docs, backend passing it will continue working for the time being).

One small separate PR I would love you to do before 1.17.0: permit . and - in usernames, no new options for backend. That is three character classes that have to agree, and master already has them disagreeing: the anonymous check in server.go:1193 allows letters, digits, underscore and a literal space with a minimum of 3; the email check in middleware.go:307 allows \s with a minimum of 4; the form in auth.tsx:332 says minLength={3}. So an email user typing a three-letter name passes the form and is refused by the server. While you are in there, please make the minimum 3 everywhere and count characters instead of bytes: the anonymous check measures len(user), so a two-letter Cyrillic name is four bytes and passes while a thirty-three-letter one is sixty-six and is refused; utf8.RuneCountInString is what it should be, and the email regex already counts runes. The whitespace differs too, \s on the email side against a literal space on the anonymous side, so one shared class settles that as well. Nothing else from the backend half.

The rest of the changes we'll better separate into smaller chunk and discuss after the switch, as the code would look different. Most of what you propose make sense, but I would like to separate all hide options from combined login for example, and I haven't yet thought properly if we want this particular implementation but generally I'm not against it.

Issue #777, opened 2020-09-14
Without Preact 


Textras:
Is anyone using this without Preact?

umputun:
not sure what you meant. API can be used with or without preact, even without js. If you mean alternative implementation with a different framework - we don't have it and not planning to do it.

Textras:
Sorry, the latter. Got it, thank you.

Mavrin:
Hi, @Textras! You can use remark42 with any js framework. Remark42 js client is standalone and don't need any external dependencies. Preact is used underhood, but it can be replaced by another library, it  doesn't matter. For example, it can be integrated with vue https://github.com/umputun/remark42/blob/master/docs/spa.md

Issue #333, opened 2019-05-17
Errors "Unhandled promise rejection" in logs with Ghostry addon


vblz:
Should the app supports work with Ghostery? There are repeated the same errors in console while browsing page with remark widget, like https://radio-t.com/online/, using Ghostery.
![изображение](https://user-images.githubusercontent.com/2076171/57949803-e590a900-78fe-11e9-8980-745357dd460f.png)
Error in logs does not affect work of remark. All functions are available and working well.

FF 66.0.5 (64-bit)
Ghostery 8.3.4

umputun:
I don't think it has anything to do with remark42. To me, it looks like a site-specific issue. 

@Reeywhaar - agree?

Reeywhaar:
Ghostery claims that it blocked "/advertising", facebook, twitter, ya.metrika and google.analytics, I don't know how it related to remark, even it works fine as you say. I think if you install Ghostery — be ready for troubles. I can check how it works with Ghostery, but it will be purely enthusiastic move. I think issue can be closed.

PR #2128, opened 2026-07-12
Cleanup of CSS colors

dittnamn:
The intention of this cleanup is to make the interface more consistent with fewer different colors, to make it easier to make your own themes without changing a lot of colors. There's no intention of actually changing the look of the interface noticeably through this, and most users would hopefully not notice some slight color changes that has been made.

Before, there were many different colors used across the interface, but a lot of these were quite similar to each other, so colors close to each other have been changed into a slightly different one of the colors with a named feature instead of the numbered colors. A few extra named colors were added from the previous numbered colors, so now there's no numbered colors left. All variables are now in rgb or rgba values to make it consistent.

Some things have changed color more than slightly, like the indicator next to the comment actions and the comment collapse lines. This is to make the coloring more consistent across various elements, but there shouldn't be any extreme changes that stand out.

As there are now several things that are based on the named variables instead of color names, there now seems to be some pointless double definitions for light mode and dark mode. However, these have been left in the code for now to avoid changing too much at once. In the future, there's room for some further cleanups.

The usernames in dark mode were gray before, but are now using the primary color, as they are in light mode to make them stand out a bit more.

The coloring of verificationIconInactive in light mode was removed, so it's now the same gray color as in dark mode. This should show the difference between unverified and verified users more clearly.

There were different shadow colors where these were used, but it's now the same color everywhere. This makes the shadow slightly more visible around the login dropdown.

The tooltips were black in both light and dark mode, but are now inverted colors of the text vs. background colors.

--white-color is still being used in some places, and there might be more suitable options to change these into.

--inverse-text-color was added to be used for text on buttons etc. using the primary color as background, instead of just using white text. Before there were some inconsistencies with different colors on different buttons, but this should be fixed now.

--secondary-darker-text-color was renamed to --lighter-text-color, as that's what it's actually used for.

The ::selection color was transfered to a named selection-color without change, but the default color should probably be changed, at least in light mode, as it's close to other colors used, like the pinned comment background.

The border around the ghost avatar had a unique color before, but now the border is removed instead as this looks cleaner.

The table header now has a background color to make it stand out better.

The documentation is updated to describe how you change the colors using __colors__.

Along with __colors__, a new variable named __font__ has been added, to allow customization of the font used.

Interface using default colors:
Using default colors

Interface using default colors in dark mode:
Using default colors - dark mode

Interface using customized colors and font, as in the example in the updated frontend documentation:
Customized colors and font
Customized colors and font - logged in

### DIFF
--- a/frontend/apps/remark42/app/components/auth/auth.module.css
+++ b/frontend/apps/remark42/app/components/auth/auth.module.css
@@ -26,7 +26,7 @@
   min-width: 240px;
   padding: 16px;
   background-color: rgb(var(--primary-background-color));
-  box-shadow: 0 10px 15px rgba(var(--black-color), 0.1), 0 -1px 6px rgba(var(--black-color), 0.05);
+  box-shadow: 0 10px 15px rgba(var(--shadow-color)), 0 -1px 6px rgba(var(--shadow-color));
   border-radius: 6px;
 }
 
@@ -54,7 +54,7 @@
     width: 100%;
     height: 0;
     content: '';
-    border-top: 1px solid var(--line-color);
+    border-top: 1px solid rgb(var(--line-color));
   }
 
   &::after {
@@ -72,7 +72,7 @@
 
 :global(.dark) .divider {
   &::after {
-    background-color: var(--color8);
+    background-color: rgb(var(--primary-background-color));
   }
 }
 
@@ -111,7 +111,7 @@
 }
 
 :global(.dark) .provider {
-  color: var(--color32);
+  color: rgb(var(--lighter-text-color));
 }
 
 .radio:checked + .provider {
@@ -122,7 +122,7 @@
 }
 
 :global(.dark) .radio:checked + .provider {
-  color: rgb(var(--white-color));
+  color: rgb(var(--inverse-text-color));
   background-color: rgba(var(--primary-color), 0.4);
 }
 
@@ -191,7 +191,7 @@
 .error {
   margin-bottom: 12px;
   padding: 6px 8px;
-  background-color: var(--error-background);
-  color: var(--error-color);
+  background-color: rgb(var(--error-background));
+  color: rgb(var(--error-color));
   line-height: 1.2;
 }
--- a/frontend/apps/remark42/app/components/auth/components/button.module.css
+++ b/frontend/apps/remark42/app/components/auth/components/button.module.css
@@ -16,7 +16,7 @@
   cursor: pointer;
   white-space: nowrap;
   background-color: rgb(var(--primary-color));
-  color: rgb(var(--white-color));
+  color: rgb(var(--inverse-text-color));
 
   &:hover {
     background-color: rgb(var(--primary-brighter-color));
@@ -30,7 +30,7 @@
   &:disabled {
     opacity: 0.6;
     cursor: default;
-    background-color: var(--color9);
+    background-color: rgb(var(--primary-color));
   }
 }
 
--- a/frontend/apps/remark42/app/components/auth/components/oauth.module.css
+++ b/frontend/apps/remark42/app/components/auth/components/oauth.module.css
@@ -22,7 +22,7 @@
   cursor: pointer;
   align-items: center;
   justify-content: center;
-  border: 1px solid var(--line-color);
+  border: 1px solid rgb(var(--line-color));
   border-radius: 50%;
 
   &::after {
@@ -32,7 +32,7 @@
 
   &:hover {
     transform: scale(1.05);
-    border-color: var(--line-brighter-color);
+    border-color: rgb(var(--line-brighter-color));
   }
 
   &:active {
@@ -40,8 +40,8 @@
   }
 
   &:focus {
-    border-color: var(--line-brighter-color);
-    box-shadow: 0 0 0 2px var(--line-color);
+    border-color: rgb(var(--line-brighter-color));
+    box-shadow: 0 0 0 2px rgb(var(--line-color));
   }
 }
 
--- a/frontend/apps/remark42/app/components/avatar/avatar.module.css
+++ b/frontend/apps/remark42/app/components/avatar/avatar.module.css
@@ -6,5 +6,4 @@
 
 .avatarGhost {
   box-sizing: border-box;
-  border: 1px solid var(--color40);
 }
--- a/frontend/apps/remark42/app/components/avatar/avatar.module.css
+++ b/frontend/apps/remark42/app/components/avatar/avatar.module.css
@@ -6,4 +6,5 @@
 
 .avatarGhost {
   box-sizing: border-box;
+  border: 1px solid rgb(var(--ghost-color));
 }
--- a/frontend/apps/remark42/app/components/button/button.module.css
+++ b/frontend/apps/remark42/app/components/button/button.module.css
@@ -10,7 +10,7 @@
   white-space: nowrap;
 
   &:focus {
-    box-shadow: 0 0 0 2px var(--color47);
+    box-shadow: 0 0 0 2px rgb(var(--primary-brighter-color));
     outline: none;
   }
 
@@ -23,37 +23,37 @@
 .kindLink {
   background: transparent;
   font-weight: 600;
-  color: var(--color9);
+  color: rgb(var(--primary-color));
 
   &:hover {
-    color: var(--color33);
+    color: rgb(var(--primary-brighter-color));
   }
 
   &:disabled,
   &:hover:disabled {
-    color: var(--color9);
+    color: rgb(var(--primary-color));
   }
 }
 
 .kindPrimary {
-  background: var(--color15);
-  color: var(--color6);
+  background: rgb(var(--primary-color));
+  color: rgb(var(--inverse-text-color));
 
   &:hover {
-    background: var(--color33);
+    background: rgb(var(--primary-brighter-color));
   }
 
   &:hover:disabled {
-    background: var(--color15);
+    background: rgb(var(--primary-brighter-color));
   }
 }
 
 .kindSecondary {
-  background: var(--color6);
-  color: inherit;
+  background: rgb(var(--inverse-text-color));
+  color: rgb(var(--primary-color));
 
   &:hover {
-    box-shadow: inset 0 0 0 2px var(--color33);
+    box-shadow: inset 0 0 0 2px rgb(var(--primary-brighter-color));
   }
 }
 
@@ -70,14 +70,14 @@
 
 .themeDark {
   &.kindSecondary {
-    background: var(--color8);
-    color: var(--color20);
+    background: rgb(var(--primary-background-color));
+    color: rgb(var(--primary-text-color));
   }
 
   &.kindLink {
     &:disabled,
     &:hover:disabled {
-      color: var(--color6);
+      color: rgb(var(--inverse-text-color));
     }
   }
 }
--- a/frontend/apps/remark42/app/components/button/button.module.css
+++ b/frontend/apps/remark42/app/components/button/button.module.css
@@ -10,7 +10,7 @@
   white-space: nowrap;
 
   &:focus {
-    box-shadow: 0 0 0 2px rgb(var(--primary-brighter-color));
+    box-shadow: 0 0 0 2px rgba(var(--primary-brighter-color), 0.4);
     outline: none;
   }
 
@@ -44,7 +44,7 @@
   }
 
   &:hover:disabled {
-    background: rgb(var(--primary-brighter-color));
+    background: rgb(var(--primary-color));
   }
 }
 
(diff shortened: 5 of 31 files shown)


paskal:
Thanks for taking this on, and sorry for the slow response. I have rebased the branch onto current master (one conflict, in `global.css`, where master's `padding: 0` met your `font-family: var(--font-family, "system-ui")`; I kept both). CI is green on the rebase, but a review turned up a few things that CI cannot see.

### Two declarations are invalid CSS and get dropped

`app/styles/global.css:95`

```css
color: rgb(var(--deactivated-color), (136, 136, 136));
```

The fallback has to live inside `var()`, otherwise `rgb()` receives six arguments and the whole declaration is discarded at parse time, leaving the iframe preloader with no colour. It should be:

```css
color: rgb(var(--deactivated-color, 136, 136, 136));
```

`app/styles/global.css:71`

```css
color: var(--primary-background-color, #fff);
```

Since the token now holds a bare `r, g, b` triplet, this computes to `color: 255, 255, 255`, which is invalid, so the declaration is dropped here too. It needs `rgb(...)`, and the `#fff` fallback is unreachable because the variable is always defined in `:root`.

Neither is caught by anything: `pnpm lint:styles` passes and no test renders `.preloader` or `.preloader_view_iframe`.

### Two elements become invisible

`app/components/spinner/spinner.module.css` changes the grey spinner's visible arc from `--color37` (`#586069`) to `--line-color` (`#e2efef`), while its track becomes `rgba(var(--shadow-color))`. On a white background the arc is now lighter than the track it is supposed to stand out against, so that spinner effectively disappears.

`app/components/comment-form/comment-form.module.css:180` gives `.preview` a `border-color` of `rgb(var(--primary-background-color))` and, on the next line, the same value for `background`. The `1px dashed` preview outline is therefore no longer visible.

### Several contrast ratios drop below WCAG AA

| surface | before | after |
|---|---|---|
| auth button hover, white text | 3.49:1 | 2.14:1 |
| primary button label | 3.5:1 | 2.87:1 |
| profile counter badge | 4.68:1 | 2.87:1 |
| new markdown table header | n/a | 2.86:1 |

The counter badge was compliant for large text before this change and is not afterwards. Most of these follow from `--primary-brighter-color` moving from `0, 153, 153` to `0, 197, 197` and from surfaces moving to `--primary-color` (`#0aa`).

### The numbered variables need to stay as aliases

Deleting `--color0` through `--color46` silently resets any custom theme built against them. `templates/iframe.ejs` still accepts those keys, since it only checks that a key starts with `--`, so an existing override keeps applying and simply does nothing, which is close to undiagnosable from the outside.

These are a public interface, and we keep public interfaces around indefinitely. Please keep every numbered variable defined, mapped to whichever named variable replaced it, so existing themes keep working. Marking them as superseded in a comment is welcome; removing them is not. New code should of course use the named ones.

### Appearance changes beyond the rename

The description says the look should not change noticeably, and for most of the diff that holds. A few changes do alter it, though, and we would like those reverted so the PR does what it says:

- `comment.module.css:213,253,307,361` turns usernames teal across the whole thread; they were a neutral grey.
- `tooltip.module.css:10,45` swaps `--black-color`/`--white-color` for `--primary-text-color`/`--primary-background-color`, which are theme-inverted, so the tooltip flips from black-on-white to light-on-dark in the dark theme.
- `avatar.module.css:7` drops `border: 1px solid var(--color40)` from `.avatarGhost` with no replacement, removing the outline (and shifting layout, since the element is `border-box`).
- `comment.module.css:23` drops `color` from `.verificationIconInactive`, so it inherits full-contrast text and no longer reads as inactive.
- `button.module.css:53` changes `.kindSecondary` text from `inherit` to teal, and uses `--inverse-text-color`, a text token, as a background fill.
- `button.module.css:46` makes `:hover:disabled` set the same value as `:hover`, so disabled primary buttons now brighten on hover; it previously reverted to the base colour.
- `input.module.css:26` moves the disabled input fill from `#f1f5f9` to `--line-brighter-color` (`#cbd5e1`), a visibly darker grey; `--highlight-color` looks like the intended match.
- `thread.module.css:33,94` leaves the collapse-line hover state nearly indistinguishable from its resting colour in both themes, losing the affordance.

The one appearance change we would like to keep is the new markdown table header styling in `raw-content.css`, which is a genuine improvement. It needs a darker background than `--primary-color` to clear the contrast floor against white text, though.

### `__colors__` cannot reach the dark theme

Values passed through `__colors__` are set on `:root`, while every dark value is declared on the `:root .dark` descendant, which wins. So overriding any of the eleven variables redefined for dark has no effect in that theme. The documentation notes this in passing, but since easier theming is the goal of the change it is worth addressing directly.

### A dark-theme legibility regression

`app/components/comment-form/text-expander.module.css:37` changes `.suggesterItem:hover` from white to `rgb(var(--primary-background-color))`, which resolves to `38, 38, 38` under `.dark`. That puts near-black text on the teal highlight. `--inverse-text-color` looks like the intended token.

### Documentation

In `site/src/docs/configuration/frontend/index.md`, the `__colors__` signature is not copyable as written: it is shown as an array while the value is an object, several entries carry a stray colon inside the key string, one is missing its colon separator, and `--shadow.color` should be `--shadow-color`. `app/typings/global.d.ts:44` describes the same value as "an array of different rgb values". The page also documents `--primary-darker-color`, which no stylesheet reads, so setting it does nothing.

### Smaller things

- `--inverse-text-color` and `--white-color` both hold `255, 255, 255`, and 14 call sites still use the latter, so one colour now has two names.
- `--shadow-color` bakes alpha into the token, unlike every other colour, so `rgba(var(--shadow-color))` cannot vary opacity. In `auth.module.css` two shadows that were `0.1` and `0.05` have become one value.
- `input.module.css:41` still uses `rgb(var(--white-color))` in the autofill `:focus` rule while the base rule moved to `--primary-background-color`, so a focused autofilled input goes white in the dark theme.
- Several `.themeDark` / `.themeLight` blocks now resolve to exactly their base rules and can be deleted: `thread.module.css:87-97`, `dropdown.module.css:40-46`, `settings.module.css:65-79`, `subscribe-by-email.module.css:54-63`.
- Missing trailing semicolons at `comment.module.css:337` and `:387`.

dittnamn:
Thanks for looking at it. When I get the time to look through your comments, I will check to see what I can fix.

When it comes to keeping the numbered variables, I would like you to reconsider to see if it would be acceptable to actually drop the backwards compatibility. I don't really see a good method of implementing it without unnecessary code and the combined colors also makes it difficult to know which intended change a potential user did. Creating new named variables to keep all slightly different colors goes against the intention of the change. There was also a mix between RGB colors and hexadecimal colors, and this cleanup uses RGB/RGBA colors all around to make it simpler.

The feature was undocumented before, and the only descriptions I can find on how to use it online are from people who don't even seem to use it themselves anymore. As the fallback would be to the default colors, I don't really see any major issues coming from simply removing them. Anyone using undocumented features should always be prepared that they can break with future updates. The new theming gets documented, and I assume those who went through theming through the undocumented feature won't have any issues converting their settings.

paskal:
Okay, it's reasonable - let's drop then.

umputun:
second pass over the same commit (013fa2f), done independently of the review above. Most of it came back the same, so this only covers where the two disagree and what hasn't been raised yet.

**where I read it differently**

the profile counter badge was compliant for normal text before, not only for large text. `#0e7e9d` on white is 4.68:1 and the badge is 14px/700, which is under the 18.66px-bold threshold, so 4.5:1 was the bar it was already clearing. It's 2.86:1 now.

`text-expander.module.css:37` isn't a dark-theme legibility regression. `rgb(var(--primary-background-color))` resolves to 38,38,38 there, and against the teal highlight that measures 5.28:1, better than the light theme's white-on-teal at 2.14:1. It's an inconsistency between the two themes, not something unreadable.

on keeping the numbered variables: they were never documented. `__colors__` doesn't appear anywhere under `site/` at the base commit, and the deprecation policy in `site/src/docs/configuration/parameters/index.md` is written for backend command-line options. The upgrade break is real, but keeping 46 alias variables alive works against the point of the PR. One sentence in the new `__colors__` section saying the `--colorN` variables were replaced by the named ones covers it.

`--shadow-color` taking four components is documented by this PR, in the bullet and in the example.

`__colors__` and the dark theme is sharper than "cannot reach it". It can't reach the 11 tokens redefined under `:root .dark`, and it does reach the other 10. The second half is the dangerous one, because the new doc line tells readers dark is unaffected.

**not raised yet**

`comment.module.css:360` - the PR body says dark-mode usernames now use the primary color "as they are in light mode". The base doesn't have that: both themes used `--color10` (#777). So light-mode usernames change too, on every comment byline, and contrast goes from 4.48:1 to 2.87:1.

`button.module.css:13` - the focus ring went from `--color47` (rgba(37,156,154,0.4)) to a solid `rgb(var(--primary-brighter-color))`. The same token was translated as `rgba(var(--primary-color), 0.4)` at the other six sites in this PR, so this is the one that got different treatment, and the two button components now disagree on their ring.

`comment-form.module.css:134` - `.themeDark .error` border-top moved to `--line-color` (#575757) while the light side at :169 kept `--secondary-background-color`. `--color7` maps to `--secondary-background-color` at :125, :129, :151 and :155 in the same file, so this looks like a slip. That 8px band used to blend into the form's own #333 frame.

`raw-content.css:53,61` - cell borders and the even-row background are both `rgb(var(--line-color))` now, in both themes, so borders disappear on every alternate row. They were distinct before (`--color16` vs `--color21` in light).

`global.css:6` - `var(--font-family, "system-ui")` quotes a generic font-family keyword, which turns it into a family name nothing matches. It can't fire anyway, since `:root` always defines `--font-family`, so dropping the fallback is cleaner than fixing it.

`comment-form__subscribe-by-telegram.module.css:12` - downstream of the `global.css:95` fix. That rule does `filter: invert(1)` assuming the preloader dots are white. With the `color` declaration dropped they inherit `--primary-text-color` and invert to near-white on the light dropdown.

on the invalid declarations generally: `c26f45e5` fixed the same class of bug, a `var()` result that isn't a valid color falling back to inherited, and that commit deliberately kept the inline fallbacks on these two preloader rules.

going the other way, one thing to strike: the Subscribe button's spinner isn't affected. `.kindPrimary` sets `color: rgb(var(--inverse-text-color))` explicitly, so the dots stay white there.

**docs**

the light-theme sentence is the one worth changing on the branch, since it's wrong in the direction that breaks things.

the rest is ours to clean up rather than something to ask for here, recording it so it doesn't get lost: the `__colors__` signature line is written as an array, with trailing colons inside the keys and `--shadow.color` instead of `--shadow-color`; the "an array" wording in the bullet and in the new `global.d.ts` comment contradicts `Record<string, string>`, and an actual array makes the loop log "Key 0 must start with `--`" twice and apply nothing; `--primary-darker-color` is in the documented example but no stylesheet reads it; and neither `__colors__` nor `__font__` reaches `last-comments` or `counter`, which render into the host page, while the example enables `last-comments`.

**for later, not for this branch**

`--selection-color` isn't in the `:root .dark` block, so selected text is near-white on pale teal in dark. One line, and this PR is what makes it cheap.

`create-iframe.test.ts:88` already asserts the destructure-out-of-params shape for `onReveal`, so `__font__` could use the same three-line test.

the `.themeDark`/`.themeLight` pairs that now resolve identically in root, dropdown, settings, subscribe-by-email and text-expander. Leaving them alone on this branch is the right call, as the PR body says. One exception, `markdown-toolbar`: its `:hover` genuinely differs, so it can't be collapsed the same way.

**__font__**

it's a new permanent public config key arriving in a PR about colors, and #1633 has an open proposal covering the same ground, `styling.colors.primary.light/dark` plus fonts. Unlike the color tokens it does work in both themes, since it's set inline on documentElement and nothing redeclares it. Worth saying why it belongs here rather than on its own.

dittnamn:
Alright, I've made a few changes and I hope I've covered it without adding any more errors. A few parts from the code review haven't changed (at least not yet):

The intention behind having both `--white-color` and `--inverse-text-color` with the same default color is that there are some cases where the color always should be white, like the QR code background. I think all occurrences where the color doesn't actually need to be white while retheming should use an alternative variable as the color source.

When it comes to the `tooltip.module.css`, this was one of those places where I didn't think there always need to be actual white on actual black, and that it would be better to have it varying by the theme. I know it's a larger change appearance-wise to invert them in the dark theme, but can this be an acceptable change to be able to move away from using `--black-color`?

I might be missing something, but the change in `spinner.module.css` should only be affecting the dark theme, where `--line-color` is equivalent to `#575757`. Compared to the old `--color37` (`#586069`), this should just be a minor change.

The hover colors for the collapse line are similar due to using the `--line-color` and `--line-brighter-color` for consistency with other lines in the system without introducing another named variable. The default colors of lines could be changed a bit for this contrast to increase, but I think this is a decision better left for you to decide on as it would change all lines.

Similar to this, the lower contrasts for the buttons were partly caused by trying to combine multiple colors with each other and that there were some inconsistencies between different buttons. Any further changes here might cause changes in the design that I think it's better for you to decide on. It's possible to redefine the default values, or switch the button hovers to `--primary-darker-color` instead. I would avoid adding new named colors, but that's also an option.

Regarding the added `__font__` setting along with the color changes, this was due to it using the same mechanism as the `__colors__` setting, and it felt like it belonged with these changes as it's also about theming and in line with the added customization/documentation. On its own, it would be a pretty small change that shouldn't need a separate PR.

PR #2146, opened 2026-08-17
feat: add configurable SMTP HELO hostname

### LINKED ISSUE #2145: feat: Add SMTP_HELO_HOST configuration to fix EHLO localhost on strict mail servers
### Description
I run Remark42 in a rootless Podman container behind a strict production Postfix mail server. The mail server enforces `reject_non_fqdn_helo_hostname`, which means it flat out rejects any SMTP connection that greets it with `EHLO localhost`.

Right now, Remark42 fails to send notifications with this error:
`504 5.5.2 <localhost>: Helo command rejected: need fully-qualified hostname`

Even if I set the container's hostname to a valid FQDN (like `comments.mydomain.de`) or tweak `/etc/hosts`, the handshake is still sent as `localhost`. This happens because Remark42 uses `go-pkgz/email` under the hood. Go's standard `net/smtp` library defaults to a hardcoded `"localhost"` string if an explicit hostname isn't passed directly to the client handshake function. 

Since I cannot change the production mail server settings, the only clean fix is to let Remark42 tell the Go SMTP client what hostname to use. 

I'd like to propose adding a new environment variable, for example `SMTP_HELO_HOST` (and a matching CLI flag `--smtp.helo-host`), so we can explicitly set the EHLO/HELO domain.

### Suggested Implementation
In `go-pkgz/email`, the standard `smtp.SendMail` helper isn't flexible enough because it handles the connection and hardcodes the `Hello("localhost")` call internally. To fix this, the dialer logic needs to use `smtp.NewClient` manually so we can pass a custom HELO string.

It would look something like this where the SMTP client is initialized:

```go
// Assuming heloHost is populated from the new SMTP_HELO_HOST env var.
// If empty, it can fallback to "localhost".

client, err := smtp.Dial(fmt.Sprintf("%s:%d", smtpHost, smtpPort))
if err != nil {
    return err
}
defer client.Close()

heloStr := "localhost"
if heloHost != "" {
    heloStr = heloHost
}

if err := client.Hello(heloStr); err != nil {
    return fmt.Errorf("failed to send HELO: %w", err)
}

if useTLS {
    if err := client.StartTLS(tlsConfig); err != nil {
        return err
    }
}
```

Could we get this added? It would make deploying Remark42 in hardened container environments a lot easier without having to route traffic through an extra SMTP relay sidecar container just to fix a header string.

damnms:
Allow the SMTP HELO/EHLO hostname to be configured separately from the SMTP server hostname, as `--smtp.helo_host` / `SMTP_HELO_HOST`.

Remark42 has always greeted the mail server as `localhost`. A relay that enforces `reject_non_fqdn_helo_hostname` or `reject_unknown_helo_hostname`, which is the default posture on many hardened Postfix installs, refuses that greeting and the message never leaves. Left unset the setting keeps the previous `localhost` greeting, so existing deployments are unaffected.

The field landed upstream in [go-pkgz/notify v1.4.0](https://github.com/go-pkgz/notify/releases/tag/v1.4.0) and [go-pkgz/auth v2.2.0](https://github.com/go-pkgz/auth/releases/tag/v2.2.0), so this needs both bumps to compile: the versions in master have no `HELOHost`. They also carry `go-pkgz/email` to v0.8.0, and the example module is tidied alongside as any change to `backend/go.mod` requires.

Documented in two places: the parameters table, and the email setup page, where it gets its own section covering what it does, that leaving it unset changes nothing, and the failure it exists to fix.

### Covers both mail paths

Remark42 sends mail through two clients: `go-pkgz/notify` for notifications and `go-pkgz/auth`'s `sender.EmailParams` for email-authentication verification. Setting only the first would have left a strict relay accepting notifications while still rejecting sign-in emails, which is the half-fixed state the reporter would have hit. `go-pkgz/auth` [gained the same field](https://github.com/go-pkgz/auth/pull/313) for this, and both are now wired to `SMTP_HELO_HOST`.

Resolves #2145.


damnms:
tested on my server, seems to work

paskal (approved):

damnms:
looks to me as if something went wrong. it shows me 120 files are changed. shouldnt this be maybe 4 or 5?

paskal:
It's vendor dir update, can be avoided only by separate PR but with rebase I think it's okay to keep it here. @umputun, feel free to review and merge the current version.

umputun (approved):

PR #2155, opened 2026-08-19
chore(deps): bump the go-modules-updates group, tidy the example module

paskal:
#2152 fails its `test examples` step with:

```
go: updates to go.mod needed; to update it:
	go mod tidy
```

The example module replaces `github.com/umputun/remark42/backend` with `../../`, so `backend/_example/memory_store/go.mod` carries the backend's dependencies as indirect entries. The group bump moved those in `backend/go.mod` without touching the example, leaving the two out of step.

This branch is #2152 rebased onto current master with `go mod tidy` and `go mod vendor` run in the example module. The tidy brings `go-pkgz/rest` to v1.23.2, `golang.org/x/crypto` to v0.55.0, `x/image` to v0.45.0, `x/net` to v0.57.0 and `x/sys` to v0.47.0, promotes `testify` to v1.12.0, and drops `go-spew` and `go-difflib`, which are no longer required. The vendored tree needed no changes and already matched.

Verified after the rebase: the backend builds and its store and API suites pass, and in the example module `go build -race ./...`, `go test -race ./...`, `go build -mod=vendor ./...` and `golangci-lint run --config ../../.golangci.yml` all pass with a re-run `go mod tidy` producing no further changes.

This supersedes #2152, which can be closed once this lands.

PR #2156, opened 2026-08-19
ci: track the latest Go 1.25 patch instead of pinning one

paskal:
`govulncheck` fails on master, and on every open backend PR, against the pinned `go-version: "1.25.12"`. Seven stdlib advisories are reported, all with `Fixed in: go1.25.13`:

`GO-2026-5026`, `GO-2026-5972`, `GO-2026-6088`, `GO-2026-6089`, `GO-2026-6090`, `GO-2026-6091` and `GO-2026-6218`, across `crypto/tls`, `encoding/asn1`, `encoding/xml`, `html/template`, `net/http` and `net/url`.

None of them needs a source change; they are toolchain-only reports that clear with the patch release. But pinning 1.25.13 would only move the problem to the next advisory, which is what happened with `76d0cc2c`, where 1.25.12 was pinned for `GO-2026-5856` and is now itself out of date.

`setup-go` accepts a minor-only spec, so `"1.25"` resolves to the newest patch on that line at run time and this stops recurring. Staying on `1.25` rather than `stable` keeps a move to a new minor a deliberate change, and matches the `go 1.25.0` directive in both `backend/go.mod` and the example module. All four pins are in `ci-backend.yml` and `release.yml`; nothing else in the repo fixes a Go version.

Worth noting that this is also what is red on #2155, whose own changes are unrelated to these reports.

### DIFF
--- a/.github/workflows/ci-backend.yml
+++ b/.github/workflows/ci-backend.yml
@@ -42,7 +42,8 @@ jobs:
       - name: install go
         uses: actions/setup-go@v6
         with:
-          go-version: "1.25.12"
+          go-version: "1.25"
+          check-latest: true
           cache-dependency-path: backend
 
       - name: test and build backend
@@ -97,7 +98,8 @@ jobs:
       - name: install go
         uses: actions/setup-go@v6
         with:
-          go-version: "1.25.12"
+          go-version: "1.25"
+          check-latest: true
           # both go.sum files so the cache key covers the main and example modules scanned below
           cache-dependency-path: |
             backend/go.sum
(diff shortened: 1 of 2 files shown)


umputun:
the vulnerability scan here was red for a reason unrelated to this PR: GO-2026-6222 in `golang.org/x/image` v0.43.0, reachable from `app/store/image/image.go:333` where `image.Decode` runs on uploaded data. I've bumped it to v0.45.0 on master in 35a389cb, with the example tidied in the same commit per your #2154.

can you rebase this onto master? It should go green then, and I'd rather judge the pin-vs-float question on a clean run than on a red one that was never about your change.

no objection to the change itself so far. Tracking the latest patch does mean stdlib CVE fixes arrive without a manual bump, which is the failure mode we've hit before.

umputun (approved):

PR #2162, opened 2026-08-19
Add backlog item for the blocked preact upgrade

paskal:
`preact` sits at an exact `10.6.2` pin, 93 releases behind the current 10.x, and looks like ordinary drift that a lockfile refresh would clear. It is not, so this records what a bump actually costs before someone tries it again.

Type errors scale with the version, from 0 at 10.10.6 to 78 at 10.29.8, all of them `TS2786` and `TS2322` around the `react` alias to `@preact/compat@^17.1.1`. The real blocker sits below that: 10.10.6 type-checks, lints, builds and passes translation and size checks, and still fails 10 tests across the Auth and Select suites. Both readable failures are a state change that no longer produces a DOM update, so it is a reconciliation difference rather than a typing artefact. The identical tree on 10.6.2 passes all 320 tests.

No dependency is changed here; the file records the measurements and points at settling the `@preact/compat` alias first.

### DIFF
--- a/docs/backlog/preact-upgrade-blocked.md
+++ b/docs/backlog/preact-upgrade-blocked.md
@@ -1,0 +1,42 @@
+---
+worth: yes
+where: frontend/apps/remark42/package.json:38
+added: 2026-08-19
+---
+# preact cannot be bumped past 10.6.2 without code changes
+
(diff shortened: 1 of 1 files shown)


paskal:
Superseded: doing the upgrade rather than recording it.

PR #2163, opened 2026-08-19
Update preact to 10.29.8

paskal:
`preact` was pinned exactly at `10.6.2` in two places, `frontend/apps/remark42/package.json` and the `pnpm.overrides` block, so no lockfile refresh ever moved it. It is now on 10.29.8, 93 releases ahead.

| package | from | to |
|---|---|---|
| `preact` | 10.6.2 | 10.29.8 |
| `@preact/compat` (the `react`/`react-dom` alias) | ^17.1.1 | ^18.3.2 |
| `@testing-library/preact` | 3.2.2 | 3.2.4 |
| `typescript` | ^4.7.4 | ^5.9.3 |

### TypeScript had to move first

Current preact types a function component's return as `ComponentChildren`, which only satisfies a JSX component check on TypeScript 5.1 or newer through `JSX.ElementType`. On 4.9 every component in the tree fails with `TS2786: 'X' cannot be used as a JSX component`. 5.9.3 is what `frontend/e2e` already uses, so this aligns the app with the rest of the workspace rather than introducing a new version.

`tsconfig.json` moves to `"jsx": "react-jsx"` with `"jsxImportSource": "preact"`, because the classic `jsxFactory` setup makes TypeScript read a global `JSX` namespace while modern preact scopes it to `preact/jsx-runtime`. The bundle is deliberately unaffected: `ts-loader` now overrides `jsx` back to `preserve`, so JSX still reaches babel intact and `babel-plugin-jsx-remove-data-test-id` keeps working. Jest already ran the automatic runtime with `importSource: preact`.

### Component typings

Preact now splits `HTMLAttributes` into element-specific interfaces and wraps attribute values in `Signalish`. `Button`, the auth `Button`, `Input` and `TextareaAutosize` move to `ButtonHTMLAttributes`, `InputHTMLAttributes` and `TextareaHTMLAttributes`, and `type` narrows from `string` to the literal union the DOM accepts. `getHandleClickProps` returns `role` as a literal so it satisfies `AriaRole`, and `useSessionStorage` returns `Dispatch<StateUpdater<T>>`, which is what `useState` has returned since the setter and updater types were separated.

`TextareaAutosize` also forwards its ref properly. It previously did `externalRef || localRef` and then read `.current`, which is not valid for a callback ref; it now keeps one internal ref and syncs the forwarded one, so both object and callback refs work.

### Tests

Three dispatches had to change, and the reason is worth recording. `preact/compat` maps `onFocus` onto `focusin` and `onBlur` onto `focusout` (`preact/compat/src/render.js`), so once compat is loaded a plain `focus` event never reaches the handler: dispatching `focus` natively gives 0 calls and `focusin` gives 1. The `Select` and `Auth` specs now dispatch the events the handlers are actually bound to.

`sort-picker` is the same class of problem from the other direction. There the handler is correctly bound to `change`, and a native dispatch fires it, but `fireEvent.change` does not reach elements rendered inside the compat providers, so the spec sets the value and dispatches directly.

### Bundle size

preact's growth over 93 releases pushes three bundles past their limits. Measured from this branch:

| bundle | measured | limit before | limit now |
|---|---|---|---|
| `remark.mjs` | 78,024 B | 76 KB | 80 KB |
| `last-comments.mjs` | 38,966 B | 39 KB | 40 KB |
| `deleteme.mjs` | 14,526 B | 14 KB | 15 KB |

`remark.mjs` first went to 78 KB and failed CI by 24 bytes, because size-limit reads `78 KB` as 78,000. 80 KB leaves about 2 kB of headroom rather than repeating that. `last-comments.mjs` was not pushed over by this change; it sat at 38,966 against a 39,000 limit, so any change at all would have tripped it.

### Verified

`pnpm lint`, `pnpm type-check`, `pnpm coverage` (320 of 320 passing across 39 suites), `pnpm build`, `pnpm translation-check` and `pnpm size-check` all pass, and `pnpm audit` reports no known vulnerabilities.

Rebased on current master, which now includes #2160 and #2172. The lockfile conflicted with the latter and was regenerated rather than hand-merged; lint, type-check, 320 tests, build and size-check were re-run afterwards.


paskal:
Good to be reviewed and merged now.

umputun (commented):
nothing blocking, all minor. Six things worth a look before merge, and three separate items at the bottom that aren't yours to fix.

both Copilot points are dead, don't act on them: the forwarded-ref one you already fixed in 11a26c15 (preact's `useImperativeHandle` is a layout effect with deps `args.concat(ref)`, and it nulls object refs and calls callback refs with `null` on unmount), and the `engines.node` one is pre-existing from #2160 - `undici@7.29.0` and `brace-expansion@5.0.9` are byte-identical in the merge base.

**findings**

`frontend/apps/remark42/package.json:73-74` - the `@typescript-eslint/eslint-plugin` bump to 8.62.1 never loads. `eslint-config-react-app` is the only config declaring the `@typescript-eslint` plugin, and `@rushstack/eslint-patch` resolves it relative to that config file, whose sibling `node_modules` symlinks to 5.62.0, so lint runs parser 8 against v5 rule implementations. Nothing regressed, the enforced rule set is identical to master, but the manifest declares a plugin ESLint never loads. Fix: force it through `pnpm.overrides` (`eslint-config-react-app>@typescript-eslint/eslint-plugin`), or drop the plugin from devDependencies and keep only the parser. Adding `plugins: ['@typescript-eslint']` to `.eslintrc.js` throws `PluginConflictError`, so that route is out.

`frontend/CLAUDE.md:42-45` and `frontend/apps/remark42/webpack.config.js:120-122` - the stated reason for the ts-loader `jsx: 'preserve'` override isn't true. `babel-plugin-jsx-remove-data-test-id@3.0.0` has an unconditional `ObjectProperty` visitor next to its `JSXOpeningElement` one, and TS `react-jsx` emit produces exactly the string-literal key it matches, so test ids get stripped either way. Fix: the override is still worth keeping, but for a different reason, it leaves `@babel/preset-react` as the single JSX transform so `.babelrc.js` stays the one place `importSource` is set. Say that in both places.

`frontend/apps/remark42/.size-limit.js:8,16` - the bundle size paragraph in the description is stale. It documents the limits from 27171fea (remark.mjs 78 KB) while c158f644 committed 80, and doesn't mention the `last-comments.mjs` 39 -> 40 raise at all. Fix: update the paragraph to all three with the CI-measured bytes. The 40 KB is right, don't lower it, 36 bytes of headroom would trip on the next change.

`frontend/CLAUDE.md:44-45` - `grep -c data-testid public/*.mjs` covers only half the output. The legacy config emits `public/*.js` off the top-level `.babelrc.js` plugin list while the modern one merges an inline list on top, so dropping the plugin from the top-level list breaks the legacy bundle and the check still reports 0. Fix: `grep -c data-testid public/*.js public/*.mjs`.

`frontend/apps/remark42/package.json:73-74` - the `@typescript-eslint` 5 -> 8 jump isn't in the description's table, and it's the biggest bump in the PR. Fix: add the row with the reason (5.62 caps its TS peer below 5.1, and `packages/api` was already on 8.62.1). Given the first item, say the parser moved rather than implying both halves did.

`frontend/CLAUDE.md:21-24` - the `preact` and `@testing-library/preact` bullets went away but both pins are still there, preact in two places that have to agree. That two-place pin is what froze 10.6.2 for 93 releases and it was never written down. Fix: not in that section, its stated cause (pnpm 10 hoisting) is what this PR disproved. A standalone line saying the version is duplicated in the app manifest and `pnpm.overrides`, and that a bump has to change both.

**separate, not for this PR**

1. `import { h }` is dead in all 56 `.tsx` files under the automatic runtime. Nothing flags it, no `noUnusedLocals`, and `no-unused-vars` is off for `.ts?(x)`. Worth a mechanical pass with `@typescript-eslint/no-unused-vars` turned on so it can't come back.
1. four components stayed on the generic `JSX.HTMLAttributes` (`select.tsx`, `icon-button.tsx`, `dropdown__item.tsx`, `arrow.tsx`) while four moved to element-specific ones. `IconButton` renders a `<button>` but its props now reject `type` and `disabled` where `Button` takes both.
1. `engines.node` is still `>=18`, the only pin in the repo still saying 18. `.nvmrc`, all CI workflows and both Dockerfiles are on 20.

paskal:
The `engines.node` point Copilot raised is addressed in #2168 rather than here, since it grew past a one-line change: four documentation references were still telling contributors to install Node 16 and PNPM 8, and the version turns out to be pinned in fifteen places with nothing linking them. That PR raises the floor to 20, fixes the stale docs and records the full enumeration in `CLAUDE.md`.

Keeping it separate because this PR introduces none of the Node-20-only dependencies; `undici` and `brace-expansion` arrived with #2160, which is already merged.

umputun:
the bundle limits don't match what the description says. Body and commit message both say two limits, going 76 to 78 with a measurement of 77.02 kB, but `.size-limit.js` raises three:

- `remark.mjs` 76 to 80 KB
- `last-comments.mjs` 39 to 40 KB
- `deleteme.mjs` 14 to 15 KB

`remark.mjs` at 80 leaves about 3 KB of slack over the 77.02 you measured, and `last-comments.mjs` isn't mentioned anywhere. Could you set each one just above what it actually measures and say so in the body? A limit with unexplained headroom stops catching the next regression, which is the thing it exists for.

no other concerns. The preact and TypeScript moves look right, and I particularly like that the focus tests ended up calling the element's real `.focus()` rather than dispatching the `focusin` that compat happens to listen for. Same for the sort-picker one dispatching a genuine `change` event with the reason written down. Those read as tests made more honest rather than adjusted to pass, which is not how these usually go.

umputun:
correcting myself on the limits above: I took the 77.02 kB from the description instead of measuring, and the description is what's stale. Built from this branch the numbers are 78,024 bytes for `remark.mjs`, 38,966 for `last-comments.mjs` and 14,526 for `deleteme.mjs`, so all three genuinely exceed their old limits, and `last-comments.mjs` had only 34 bytes of room before. The new 80/40/15 leave 1,976, 1,034 and 474 bytes, which is modest rather than the slack I claimed. So the limits are right and only the body needs updating to say three bundles with the real figures.

one thing that did come out of measuring: restoring `fireEvent.focus` and `fireEvent.blur` in the Select and both Auth cases passes 42 of 42 against the upgraded Preact graph, so those three did not have to change. Calling the element's real `.focus()` is still the better test and I would keep it, but the reason given in the description is not what was happening. Sort-picker is the real one, and I confirmed it: putting `fireEvent.change` back gives `updateSorting` zero calls.

separately, this and #2172 conflict in `frontend/pnpm-lock.yaml`, and #2172 changes the workspace graph. Worth merging that one first and then rebasing this so the lockfile is regenerated against the final three-project graph, which is also the only way this branch's CI tests the state that actually ships.

paskal:
You are right, and the description was the stale part. Corrected in the body: three limits move, not two, and the measured values are the ones you built, 78,024 / 38,966 / 14,526.

Two details worth recording. `remark.mjs` first went to 78 KB and failed CI by 24 bytes, because size-limit parses `78 KB` as 78,000; 80 KB leaves roughly 2 kB rather than repeating that. And `last-comments.mjs` was not pushed over by this change at all: it was already sitting at 38,966 against a 39,000 limit, so the next change of any kind would have tripped it.

paskal:
@umputun, rebased - feel free to merge.

UPD: fixing CI...

umputun (approved):

umputun (approved):

Issue #825, opened 2020-11-30
Switch to backend-generated UI


umputun:
Just an idea I'd like to explore and will apricate feedbacks from both backend and frontend devs.

Currently, we have a decent separation between UI and the backend with REST-like API. This is all fine and standard; however, in multiple places the separation bites us, and we are forced to do odd things, like infamous sed preparing HTML. Another issue is two separate places to configure things. And the last problem - we have a constant backlog of UI tasks, and the situation doesn't get any better. One more thing - with an endless number of various npm modules involved in the build/deploy/ui-functionality we are constantly racing against security issues and have to update things way too often just to address new vulnerabilities.

From another hand, our UI is not that advanced — all we need to do just 3 "widgets" with a minimal level of interactivity. Counter widget is trivial, last comments kind of static thingy and the primary comments tree already delegates some complicated parts (like preview generation) to server. The ideal solution could be fully static HTML generated by the server. This is very tempting and, theoretically, can be extremely fast, CDN friendly, and all other goodies we get from the static content. The issue with this solution - it may require a significant amount of extra generation, i.e. we will need all HTML for all sorts, all pages, and different pages/views for the different access levels. It could also be challenging from auth/security POV.

I think entirely static rendering will be hard to do right and may limit us significantly. It is why something in between, namely [htmx](https://htmx.org). We could render UI components dynamically on the server-side with this library without the need for the second language (js/ts) and a dedicated UI framework. Well, htmx is a kind of js framework but sort of а different beast. The idea of such a "dynamic server-side rendering" is something unusual these days, but not necessarily a bad thing. I'm not sure how all of this can be integrated, and if we could use it as-is or write some extensions, but generally speaking to me, this direction looks very promising.

cc: @akellbl4 @paskal

umputun:
I have very limited practical experience with htmx. All I did was a small project with a relatively simple, but dynamic UI for a monitoring/orchestration system. My impression after this project was mostly positive. It can be used and allowed me to make the frontend (bootstrap based) without much headache. 

Some problems I run to:
- Initially, it feels very unusual to write component controllers returning HTML snippets and rendering from templates
- The right structure still somewhat unclear, I mean how to organize things, where to put controllers, templates, and so on
- Sometimes it is not easy to pick the right scope for the component 
- Stateful parts of UI need to keep the state on the server-side somehow

paskal:
As we won't get rid of our API anyway, I think it's worth trying to implement that on a separate endpoint as an experiment.

[...]

PR #2164, opened 2026-08-19
Move eslint to 9 with flat config

paskal:
`eslint` moves from 8 to 9, `.eslintrc.js` and `.eslintignore` are replaced by `eslint.config.mjs`, and `prettier` moves from 2 to 3.

The chain is forced: `eslint-config-react-app` pins eslint to `^8` and has not had a release since 7.0.1, so it has to go; `eslint-plugin-prettier` 4 does not support eslint 9, and 5 requires prettier 3.

## Rule parity

Rule coverage does not drop. Comparing `eslint --print-config` before and after gives 172 active rules on an application file and 171 on a spec file either way, and 172 against the old 171 on a plain JS file.

What is gone:

- `flowtype/define-flow-type`, `flowtype/require-valid-file-annotation` and `flowtype/use-flow-type` — there is no Flow in this repository
- `no-new-object` and `no-new-symbol` — renamed upstream, and their replacements `no-object-constructor` and `no-new-native-nonconstructor` are on
- `react/forbid-foreign-prop-types` — there are no propTypes anywhere in `app/`
- `no-useless-constructor`, on TS files only — the `@typescript-eslint` version of it is on, at `error`

Eslint 9 adds `no-constant-binary-expression`, `no-empty-static-block` and `no-unused-private-class-members` through its recommended set, plus the two renames above. `prettier/prettier` and `unicode-bom` are on as well.

Several rules that `eslint-config-preact` and `typescript-eslint` do not provide are declared explicitly in the last entry of the config array, which is where they have to be: `eslint-config-prettier`, pulled in by `prettierRecommended`, switches some of them off, and in flat config the later entry wins. Turning all of them on produces no violations across the 214 files eslint now processes.

## Dependencies

Direct devDependencies go from 78 to 73, and the lockfile from 1701 resolved packages to 1651.

Removed: `eslint-config-react-app`, `eslint-plugin-flowtype`, `eslint-plugin-react`, `eslint-plugin-react-hooks`, `eslint-plugin-testing-library` and `@types/eslint`. The first four came in through `eslint-config-react-app`; the react rules that matter are still applied, through `eslint-config-preact`. `eslint-plugin-testing-library` was a dependency that no configuration ever referenced. `@types/eslint` is a stub package that has been superseded by eslint's own bundled types.

Added: `confusing-browser-globals`, which supplies the list behind `no-restricted-globals` and was previously reached through `eslint-config-react-app`.

Everything touched is on its current release: `eslint` 9.39.1, `prettier` 3.9.4, `eslint-config-preact` 2.0.0, `eslint-config-prettier` 10.1.8, `eslint-plugin-prettier` 5.5.6, `eslint-plugin-import` 2.32.0, `eslint-plugin-jsx-a11y` 6.10.2.

Eslint 10 is not reachable yet: `eslint-config-preact` 2 and `eslint-plugin-prettier` 5 both declare a `^9` peer.

## Working directory

Flat config does not cascade the way `.eslintrc` did. Eslint loads the config next to the directory it runs in, not the one nearest the file being linted, so anything invoking eslint has to run from `frontend/apps/remark42`. The pre-commit hook now does, and the documented commands are corrected to match; `pnpm lint`, `pnpm test` and `pnpm build` never worked from `frontend/` and the docs said they did.

## Formatting

Prettier 3 changes two defaults that affect this repository: `trailingComma` becomes `all`, and markdown and HTML formatting differ slightly. The reformatting it produced is in the diff, and no rule was disabled to accommodate it.


umputun (changes requested):
two blocking things.

**`frontend/apps/remark42/eslint.config.mjs:16-24`, the rule set shrank.** Dropping `eslint-config-react-app` removed 37 enforced rules on `.ts`/`.tsx` and 35 on `.js`/`.cjs`, not the four import rules plus `no-unused-vars` the description accounts for. `lint:scripts` runs at `--max-warnings=0`, so react-app's `warn` rules were hard gates too. Gone: `no-restricted-globals` (error, the 60-name `confusing-browser-globals` list, and with `no-undef` off for TS it was the only guard against a bare `location`/`event`/`name` hitting a DOM global), `eqeqeq`, `array-callback-return`, `@typescript-eslint/no-unused-expressions`, `no-throw-literal`, `no-loop-func`, `no-eval`, `no-implied-eval`, `no-script-url`, `react/no-direct-mutation-state`, `react/no-typos`. Nothing in the tree violates them today, re-enabling all 37 over the 210 files gives zero errors, so it's a gate loss rather than a regression. Fix: re-declare the ones worth keeping, or say in the description that the set was narrowed and list what went. Note `confusing-browser-globals` came in via react-app and is gone from the lockfile, so `no-restricted-globals` needs it back as an explicit devDependency or the list inlined.

**`frontend/CLAUDE.md:55` still lists `eslint` 8 under held-back majors**, so that line is false once this merges. `stylelint` 14, `babel` 7 and `jest` 28 still hold. Fix: drop the eslint entry, same as ca313203 did for `typescript` 4.7 and the preact pins.

three smaller ones, all in files already open here:

- `frontend/apps/remark42/package.json:92-98`, the `eslint-plugin-react-hooks` 4.6 -> 7.1.1 bump has no effect. Nothing imports it, so `...preact` supplies the rules and `eslint-config-preact@2.0.0` resolves its own 5.2.0. `eslint-plugin-react` and `eslint-plugin-testing-library` are orphaned the same way now that flat config imports plugins by module. Fix: drop all three and let `eslint-config-preact` own them, or register react-hooks explicitly after `...preact`.
- `frontend/apps/remark42/eslint.config.mjs:8`, `eslint --max-warnings=0 eslint.config.mjs` fails on `import/no-anonymous-default-export`, and the `lint:scripts` glob `**/*.{ts?(x),?(c)js}` skips `.mjs` so CI never sees it. The old `.eslintignore` had `!.eslintrc.js`, so the config file used to be linted. Fix: `const config = [...]; export default config;`, and widen the globs to `?(c|m)js` in `lint:scripts` plus `mjs`/`cjs` in `format`.
- `frontend/apps/remark42/eslint.config.mjs:33-40`, `caughtErrors: 'none'` and the `^_` patterns sit inside the ts/tsx block, so the twelve `.js`/`.cjs` files in the lint set get eslint 9's new `caughtErrors: 'all'` default with no escape hatch. An unused `catch (e)` in `webpack.config.js` errors where the same code in TypeScript doesn't. Fix: add a matching `{ files: ['**/*.js','**/*.cjs'], rules: { 'no-unused-vars': [...] } }` block.

one question: `packages/api` stays on eslint 8 with `.eslintrc.cjs`, and its lint script still passes `--ext`, which eslint 9 removed. Migrate it here, or track it separately? Worth saying either way, since this reads as "eslint 9 done" once merged.

paskal:
Closing rather than merging, completing the set with #2169, #2170 and #2171. eslint 8 works; this is modernisation whose payoff depends on the widget outliving the #825 decision.

Findings kept:

- `eslint-config-react-app` declares `peerDependencies.eslint: ^8.0.0`, has not moved past 7.0.1, and CRA is retired, so it blocks eslint 9 outright with no release to wait for. Everything else in the chain already supports 9.
- Dropping it silently removes rules. The ones it was actually contributing and that had to be restored explicitly: the four import rules (`import/first`, `import/no-amd`, `import/no-anonymous-default-export`, `import/no-webpack-loader-syntax`), and its `no-unused-vars` behaviour (`ignoreRestSiblings`, `caughtErrors: 'none'`, `^_`). Without the latter, seven existing sites fail, including props destructured only to keep them out of a `...props` spread.
- prettier has to go to 3: `eslint-plugin-prettier` 4 supports only eslint 7 and 8, and 5 requires `prettier >= 3` as a peer, so a pnpm override cannot hold it back. The reformat is small (16 files, 45 lines) because `.prettierrc.js` already sets `trailingComma: 'es5'`.
- eslint 9 reports unused disable directives by default, which surfaced eight `eslint-disable ... no-console` comments guarding a rule nothing had enabled.

Branch `chore/eslint-flat-config` is kept.

Note this leaves `@typescript-eslint` at 5.30 against TypeScript 5.9 once #2163 merges, whose `SUPPORTED_TYPESCRIPT_VERSIONS` is `>=3.3.1 <5.2.0`. #2163 carries that bump itself, so it is covered.

PR #2168, opened 2026-08-19
Require node 20 and record every place the version is pinned

paskal:
`frontend/apps/remark42/package.json` declared `engines.node: ">=18"` while everything that actually runs has been on 20 since the pnpm 8 to 10 migration. Raised to `>=20`.

Deliberately a plain major rather than a patch floor. Some dev dependencies are stricter within Node 20 (`undici` wants `>=20.18.1`, `brace-expansion` `20 || >=22`), and any current Node 20 satisfies them. Following those patch floors into `engines` and four documentation files would turn every lockfile refresh into a documentation change; `frontend/CLAUDE.md` records that reasoning so it does not get re-litigated.

### The documentation had drifted much further

Four references were stale, all of them what a new contributor reads first:

- `site/src/docs/contributing/frontend/index.md` told people to install Node.js 16 and PNPM 8
- `site/src/docs/getting-started/installation/index.md` said "Go 1.25, Node 16+, PNPM 8, and Perl", twice
- `CLAUDE.md` repeated the same line for local release builds

All now say Node 20 and PNPM 10, matching `.nvmrc`, the CI matrices and the container images.

### Why the version keeps drifting

It is pinned across five categories with nothing linking them: `engines` and `packageManager` fields in several manifests, both `.nvmrc` files, the `node:` matrices and `pnpm/action-setup` blocks in `ci-frontend.yml` and `release.yml`, four container images including `frontend/Dockerfile.e2e`, and the documentation above.

`frontend/CLAUDE.md` already carried a checklist for this, so the entries it was missing are folded into it rather than starting a second list. An earlier revision of this branch did start one in the root `CLAUDE.md`; that duplicate is removed, because two lists of the same thing drift apart. The additions are the `node:` CI matrices, `site/package.json`'s `engines`, and the rule about keeping the floor at the major.

Two traps are called out there: `frontend/package.json` carries `packageManager` but no `engines`, so grepping for `engines` misses it, and `site` uses yarn rather than pnpm so its `packageManager` moves independently.

### Verified

Manifests parse and the site builds clean (`yarn build`, 35 pages), which matters here because `ci-site.yml` gates its only build job to master and tags, so site changes get no pull request validation.

Raised by Copilot on #2163. Kept separate from that PR because it grew past the one-line `engines` change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,38 +23,6 @@
   - The example module replaces `github.com/umputun/remark42/backend` with `../../`, so it carries the backend's dependencies as indirect entries. Leaving them stale fails the `test examples` CI step with `go: updates to go.mod needed; to update it: go mod tidy`.
   - This applies to Dependabot pull requests too: the bot updates `backend/` only, so its Go module PRs need the example tidied before they can go green.
 
-## Node and pnpm versions live in many places
-
-Changing the required Node or pnpm version means changing **every** entry below in the same commit.
-They drift otherwise: the docs sat on "Node 16+" and "PNPM 8" while CI, Docker and `.nvmrc` had all
-moved to Node 20 and pnpm 10.
-
-Version declarations:
-- `frontend/apps/remark42/package.json` - `engines.node`, `engines.pnpm`, `packageManager`
-- `site/package.json` - `engines.node`, `engines.yarn`, `packageManager` (site uses yarn, not pnpm)
-- `frontend/package.json` and `frontend/packages/api/package.json` - `packageManager` only, and they
-  are easy to miss because neither declares `engines`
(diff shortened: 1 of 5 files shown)


paskal:
Reverted to a plain `>=20` floor, with the docs matching.

Copilot's five comments pushed toward `>=20.18.1` because `undici` declares that. Not taking it: `engines.node` states the major we support, and individual dev dependencies being stricter within that major is normal and satisfied by any current Node 20. Chasing those patch floors into `engines` and four documentation files would turn every lockfile refresh into a documentation change.

`frontend/CLAUDE.md` records the reasoning next to the pinning checklist so the next person does not re-litigate it from the same Copilot suggestion.

umputun (approved):

PR #2169, opened 2026-08-19
Move stylelint to 17

paskal:
Second of the held-back tooling majors. `stylelint` goes from 14 to 17, along with the config and plugins that had to move with it.

### Two packages had to be dropped

`stylelint-config-prettier` is deprecated: stylelint 15 removed the stylistic rules it existed to switch off, so it is now a no-op that only blocks upgrades.

`@mavrin/stylelint-declaration-use-css-custom-properties` is the harder one. Its latest release declares `peerDependencies.stylelint: ^14 || ^15`, so it caps the upgrade outright. It is replaced by `stylelint-declaration-strict-value`, which is maintained and supports 16 and 17, and enforces the same thing: colour values must come from a custom property rather than a literal.

`max-empty-lines` is also gone from the config, since stylelint 15 removed the rule.

### Matching the previous scope rather than widening it

The replacement rule is easy to configure more strictly than the one it replaces, and doing so turns a version bump into a CSS rewrite. Two adjustments were needed to keep the scope honest:

- `ignoreFunctions: true`, because otherwise `rgb(var(--primary-text-color))` is reported for not using a variable, which it plainly does. Without this the rule fires 82 times on values that are already correct.
- the rule is switched off for `templates/**`. Those are standalone pages (`demo`, `deleteme`, `comments`, `markdown-help`) rather than the themeable widget surface, and the previous plugin never linted them, so every literal colour there would have been a new error.

### Two real defects surfaced

`app/components/spinner/spinner.module.css` had a `:global(.dark) &` block at the top level. `&` has no scoping root there, so the rule never applied to anything, and its two declarations were byte-identical to the `.root` rule above it. Removed as dead code; nothing rendered differently.

`templates/markdown-help.html` used the deprecated `clip: rect(0 0 0 0)` visually-hidden pattern, now `clip-path: inset(50%)` with `clip-path: none` for the focusable state.

### Auto-fixed

49 findings from new rules in `stylelint-config-standard`, applied with `--fix`: 34 `color-function-alias-notation` (`rgba(...)` to `rgb(...)`, the CSS Color 4 form, semantically identical for our browser targets) and 15 `media-feature-range-notation`.

### On the postcss `importFrom` deprecation

`csstools/value-no-unknown-custom-properties` still resolves `importFrom` correctly after the bump to v6; I checked by linting a file referencing an undefined custom property and confirming it is still reported.

The 116 deprecation warnings in build logs come from the other call site, `postcss-custom-properties` in `webpack.config.js`. That one is deliberately not touched here: `importFrom` is what generates the literal fallback declarations that precede every `var()` usage, so removing it changes the emitted CSS rather than just the configuration. It belongs with the postcss and bundler work, not in a stylelint bump.

### Verified

`pnpm lint`, `pnpm type-check`, `pnpm coverage`, `pnpm build` and `pnpm size-check` all pass.

Stacked on #2164, which is itself on #2163; review those first. Worth noting that #2128 edits many of the same `rgba()` call sites, so whichever lands second will need a rebase.


paskal:
Closing rather than merging, same reasoning as #2170 and #2171: stylelint 14 works, and this is tooling investment in a widget whose future is undecided.

Findings kept:

- `@mavrin/stylelint-declaration-use-css-custom-properties` declares `peerDependencies.stylelint: ^14 || ^15`, so it caps the upgrade outright. `stylelint-declaration-strict-value` replaces it and supports 16 and 17.
- `stylelint-config-prettier` is a no-op since stylelint 15 removed the stylistic rules it disabled, and only blocks upgrades.
- The replacement rule needs `ignoreFunctions: true`, or it reports `rgb(var(--primary-text-color))` for not using a variable, firing 82 times on values that are already correct. It also has to be switched off for `templates/**`, which are standalone pages the previous plugin never linted.
- Two real defects surfaced and are worth fixing independently of this PR: `spinner.module.css` has a `:global(.dark) &` block at top level where `&` has no scoping root, so it never applied, and its declarations duplicate `.root`; `templates/markdown-help.html` uses the deprecated `clip: rect(...)` visually-hidden pattern.
- Separately, the `postcss-custom-properties` `importFrom` deprecation (116 build warnings) comes from `webpack.config.js`, not the stylelint config, and cannot be removed without changing emitted CSS: it generates the literal fallbacks preceding every `var()`.

Branch `chore/stylelint-16` is kept.

PR #2170, opened 2026-08-19
Move jest to 30

paskal:
Third of the held-back tooling majors. `jest` goes from 28 to 30, with `jest-environment-jsdom`, `@types/jest`, `@swc/jest`, `@testing-library/jest-dom` and `jest-fetch-mock` moving with it. All 320 tests pass.

### `@swc/core` deliberately stays at 1.2.205

This is the part worth reviewing carefully. Bumping it to current (1.16.1) fails 13 suites and 57 tests, all with the same error:

```
TypeError: Cannot redefine property: getCookie
TypeError: Cannot redefine property: telegramSubscribe
TypeError: Cannot redefine property: parseQuery
```

Newer swc emits module exports as non-configurable getters, so `jest.spyOn(someModule, 'someExport')` throws. Setting `module.type: 'commonjs'` in the transform options does not change it.

I confirmed the attribution by bisecting rather than guessing: jest 30 with swc 1.2.205 fails only on its own removed aliases, while jest 29 with swc 1.16.1 fails on 57 redefinition errors. The version of swc is the cause, not the version of jest.

Fixing it properly means migrating every `jest.spyOn` on an imported binding to `jest.mock` with a factory, which is a test-mocking migration rather than a version bump, so it is left for its own change. Note that this pins us to a 2022 build of swc until that happens.

### What jest 30 itself needed

**Removed matcher aliases**, 36 call sites across 7 files: `toBeCalledWith`, `toBeCalledTimes`, `toBeCalled` and `toThrowError` are gone in favour of the `toHaveBeen...` forms and `toThrow`.

**`fetchMock.enableMocks()` moved to module scope** in `app/__mocks__/fetch.ts`. It was called inside `beforeAll`, and jest 30's circus runner rejects hooks registered from within another hook: `Cannot add a hook after tests have started running`. Top level is the documented usage and it registers its own `beforeEach`.

**`window.location` can no longer be redefined.** `constants.config.test.ts` used `Object.defineProperty(window, 'location', ...)` to force an https protocol; jsdom now seals it. Replaced with a `@jest-environment-options {"url": "https://test.com"}` docblock, which gives the same protocol without touching the property.

### Other moves

`@testing-library/jest-dom` goes from 5 to 7, and `@types/testing-library__jest-dom` is removed because the package has shipped its own types since v6. `jest-fetch-mock` goes from 3 to 4 and `@swc/jest` from 0.2.21 to 0.2.39.

### Verified

`pnpm lint`, `pnpm type-check`, `pnpm coverage` (320 of 320 across 39 suites), `pnpm build`, `pnpm size-check` and `pnpm translation-check` all pass.

Stacked on #2169, which is on #2164, which is on #2163.


paskal:
Closing rather than merging, for the same reason as #2171: jest 28 works, so this is modernisation with a payoff that only accrues if the widget outlives the #825 decision.

The finding worth keeping, because it cost a bisect to establish:

- The failures were **not jest**. Bumping jest and `@swc/core` together failed 57 tests with `Cannot redefine property` on module exports. Jest 29 with the old swc passes all 320; jest 29 with swc 1.16 fails. Newer `@swc/core` emits non-configurable export getters, so `jest.spyOn` on an imported binding throws. Setting `module.type: 'commonjs'` in the transform does not change it.
- That is what pins `@swc/core` to its 2022 build (1.2.205). Unpinning it needs every `jest.spyOn` on a module export migrated to `jest.mock` with a factory, which is a test-mocking migration rather than a version bump.
- Jest 30 itself needed only 36 removed-alias renames (`toBeCalledWith` and friends), moving `fetchMock.enableMocks()` to module scope because circus rejects hooks registered inside hooks, and replacing a `window.location` redefinition with a `@jest-environment-options` URL since jsdom now seals it.

Branch `chore/jest-30` is kept.

Issue #2166, opened 2026-08-19
Remove the React compatibility alias and collapse the frontend toolchain


paskal:
The frontend carries a lot of machinery for a fairly small interface, and most of it traces to one decision that has stopped paying for itself. This is worth doing whatever we decide in #825, because it makes that decision much easier either way.

## Measurements

- 157 non-test source files, 8,745 lines
- 29 CSS module files
- a Redux store of 1,052 lines
- **15 runtime dependencies against 83 devDependencies**

The last line is the finding.

## The expensive part is the React impersonation, not Preact

Preact appears in around 50 non-test files, but overwhelmingly for `h` and JSX types; those files are indifferent to what renders them.

The cost is that Preact has to pretend to be React. `react` and `react-dom` in `package.json` are npm aliases to `@preact/compat`, and that alias exists for exactly two consumers: `react-redux` (13 files) and `react-intl` (32). **No file in the app imports `react` or `react-dom` directly.** Two libraries hold the entire compatibility layer in place, and `react-redux` additionally pulls a real `react` resolution into the lockfile through `use-sync-external-store`.

What that layer actually costs, all observed during the preact 10.6 to 10.29 upgrade:

- `preact/compat` silently remaps `onFocus` to `focusin` and `onBlur` to `focusout`. A test dispatching a plain `focus` event stops reaching its handler, with no error and no warning.
- `@testing-library/preact` rewrites `change` to `input` for every element the moment it detects a compat vnode. Correct for `<input>`, wrong for `<select>`, and it broke a test that had nothing to do with the upgrade.
- the third-party `@preact/compat` package was pinned at 17.x while preact was on 10.29.
- all 78 TypeScript errors in that upgrade clustered on the alias.

None of these are Preact problems.

### Removing it

**`react-redux`** is the contained half. The coupling is concentrated in `app/store/index.ts`, which imports `useDispatch`, `useSelector` and `TypedUseSelectorHook` and re-exports typed versions, plus the call sites in components. The store's own logic is plain Redux. A replacement is a small binding over Preact context.

**`react-intl`** is the real work: 32 files and five API shapes, `FormattedMessage` (53 occurrences), `useIntl` (43), `defineMessages` (30), `IntlShape` (20) and `IntlProvider` (20). The catalogues are already extracted JSON and the extraction tooling is ours in `tasks/`. Message lookup is straightforward; plural rules and date formatting need actual design rather than mechanical replacement.

[...]

PR #2171, opened 2026-08-19
Move babel to 8

paskal:
Last of the four held-back tooling majors. `@babel/core`, `preset-env`, `preset-react` and `plugin-transform-react-jsx` go from 7 to 8, and `babel-loader` from 8 to 10.

### An unmaintained plugin had to be replaced, not upgraded

`babel-plugin-jsx-remove-data-test-id` is what keeps `data-testid` out of production bundles. Its latest release is 3.0.0, it declares `peerDependencies: {"@babel/core": "^7.0.0"}`, and it does not work with babel 8. That is not inferred from the peer range; the build fails with:

```
TypeError: t.jSXOpeningElement is not a function
```

Babel 8 renamed that builder, and the plugin has not been updated.

Rather than add another third-party dependency, it is replaced by `tasks/babel-plugin-remove-test-id.js`, eight lines that remove `data-testid` JSX attributes. That is a dependency removed rather than swapped, which is the direction #2166 argues for.

Verified the replacement actually works: `grep -c data-testid` over the production bundles returns 0, the same check the previous plugin satisfied.

### preset-env dropped three options

Babel 8 removed all of these, each surfacing only once the previous one was fixed:

- `loose`, in favour of granular `assumptions`. Removed rather than translated, so output is now spec-compliant.
- `bugfixes`, which is unconditional now.
- `useBuiltIns` and `corejs`, which move to a separate plugin.

That last one needed measuring rather than translating. Wiring up `babel-plugin-polyfill-corejs3` with the documented equivalent of `useBuiltIns: 'usage', corejs: 3` blew `embed.mjs` from 2.2 kB to 11.04 kB, past its 2.5 kB limit, and `remark.mjs` to 90.61 kB against an 80 kB limit. Passing it explicit targets, and hoisting `targets` to the config root where both it and preset-env read them, changed nothing.

So I measured what the old injection was actually contributing: without any core-js plugin, `embed.mjs` is 2.21 kB against master's 2.16 kB, and every other bundle lands within a few hundred bytes of master. At `defaults, not IE 11, not samsung 12`, preset-env was injecting roughly 50 bytes. The polyfill plugin is therefore not wired in, and the config records why.

Worth a follow-up: `core-js` is still a declared runtime dependency and now has nothing injecting it.

### Also removed

`@babel/plugin-syntax-flow`, which arrived as a dependency of `eslint-config-react-app` (dropped in #2164) and has no purpose here, since there is no Flow in this codebase.

### Verified

`pnpm lint`, `pnpm type-check`, `pnpm coverage`, `pnpm build`, `pnpm size-check` and `pnpm translation-check` all pass, and the production bundles contain no `data-testid`.

Stacked on #2170, which is on #2169, on #2164, on #2163.

### Still on the list

`module:fast-async` remains in the plugin list. It dates from 2022 and every browser in `targets` supports native async, so it is likely dead weight, but removing it changes the emitted output and belongs in its own change rather than at the end of a major upgrade.


paskal:
Closing rather than merging. babel 7 is not broken, so this is modernisation whose payoff only accrues if the widget lives for years, and that is exactly what is undecided while #825 is open. Preserving the findings here so they are not rediscovered:

- `babel-plugin-jsx-remove-data-test-id` pins `@babel/core: ^7.0.0` and calls `t.jSXOpeningElement`, which babel 8 renamed, so it fails at build. It is also what keeps `data-testid` out of production bundles, so it cannot simply be dropped. This branch replaces it with an eight-line local plugin.
- `@babel/preset-env` in babel 8 removes `loose`, `bugfixes` and `useBuiltIns`/`corejs`, each surfacing only after the previous is fixed.
- The documented `useBuiltIns` replacement, `babel-plugin-polyfill-corejs3`, over-injects regardless of targets: it took `embed.mjs` from 2.2 kB to 11.04 kB past its 2.5 kB limit, and explicit targets plus root-level `targets` changed nothing. Measuring the old injection showed it was worth roughly 50 bytes at our browser targets, so the plugin is not worth wiring in.

Branch `chore/babel-8` is kept if this is revisited.

Issue #240, opened 2019-01-05
Publish client library to npm


webkonstantin:
It would be nice to be able to use remark client library directly instead of dropping script tags on the page

Something like this

```js
const remark = require('remark')

remark.embed(node, options)
remark.counter(node, options)
remark.lastComments(node, options)
```

igoradamenko:
Yep, we'll do it after switching from iframe widget to iframe-less (https://github.com/umputun/remark/issues/218#issuecomment-438005245). Right now it's overengineering to try to extract things from current widget to the package and to the iframe widget. 

I'm leaving this issue open as a reminder. Thanks.

akellbl4:
Possible Requirements:

- pack code by widget with ability to lazy load the module out of the box
- provide all of the necessary methods for controlling the widgets
- pack file with API methods separately and expose API methods for outside use

Ideas:

- create wrappers for JS frameworks (for example remark42-react) for easier integration with them
- provide theme context which provides ability to modify theme of comments
- provide ability to replace components with user's own implementation

paskal:
Closing this after seven years, as we have decided against publishing the widgets as an npm package.

The reason is authentication rather than packaging. The widget runs in an iframe served from the Remark42 origin, and the OAuth login flow depends on that: the popup authenticates against Remark42 first-party, and the cookie it sets is only usable by a widget running on that origin. Imported into a host page's bundle, the widget would run in the host's origin and its auth cookie would be third-party, which Safari blocks today and Chrome is moving away from. Login would break, and break silently.

The iframe-less direction mentioned in the first reply here did not happen, and the auth flow built since is the reason it will not.

What the request was actually after, control over when and how widgets load, is available today through `window.REMARK42.createInstance(config)` and the matching `destroy()`, which cover lazy loading, single-page app route changes and mounting on intersection. The contract is written out in #1714. For a fully custom frontend the supported path is the REST API.

Discussed further in #1715.

PR #2172, opened 2026-08-19
Drop the @remark42/api package

paskal:
`@remark42/api` is removed, along with its CI workflow, workspace entry, dependabot entry and the four `*:api` scripts. 2,845 lines deleted.

### It cannot log anyone in

This is the part that decides it. `clients/auth.ts` exposes exactly `anonymous`, `email` and `telegram`. There is no OAuth method, and there cannot be one: the popup flow lives in the widget and works only because the widget is same-origin inside the iframe. Separately, `credentials` appears nowhere in the package, so its cookie auth cannot work cross-origin as written, and its in-memory token is only ever populated from the `AUTH_SEND_JWT_HEADER` response header, which #1929 fixed for non-OAuth providers only.

So even fully packaged, published and typed, it would enable anonymous and email-link commenting and nothing else. That is not the integration surface anyone was asking for.

### Nobody uses it

npm shows 60 downloads last month and 501 in the last year, at roughly two a day, which is mirrors and scanners rather than people. All 13 published versions are alphas; twelve went out on 2022-06-19 and the last on 2022-07-11, and `dist-tags.latest` still points at `0.6.0-alpha.12`, a version this repo is *behind*.

GitHub code search finds four repositories referencing it and none is a real consumer: this repo, a wholesale copy of it, a catalogue page, and one project whose import resolves through a tsconfig path alias to its own source. The one genuine third-party custom UI, `lauralesteves/remark42-admin-ui`, calls the REST API directly with axios.

Nothing in this repository consumes it either: the widget has its own separate client at `app/common/api.ts`.

### This matches what maintainers have already said

In #1383 umputun's answer was "If you want an alternative frontend, API is here and you can implement anything you like", and "the token is a fully server-side concern. No SDK/API/whatever on the frontend can change this logic." akellbl4 asked what an SDK adds over the bare API. That position stands; this change stops the repository implying otherwise.

Every third-party integration found in the wild builds against the iframe embed script rather than any package: the Quartz plugin, the Django template tag, the PHP/OPS3 plugin, the Docusaurus wrapper. The public contract is `remark_config` and `window.REMARK42`, and it is untouched here.

### Follow-ups this does not do

The npm package is left published rather than deprecated, since that needs registry credentials. Deprecating `@remark42/api` with a pointer to the REST API docs would be the tidy finish.

#240 ("Publish client library to npm", open since 2019) and #1383 want a real answer rather than staying open. The REST reference at `site/src/docs/contributing/api` is also filed under Contributing, which is the wrong place if it is the supported integration path.

### Verified

`pnpm install`, `pnpm lint`, `pnpm build` and all 320 tests pass in the remaining workspace.

Note for sequencing: this removes the second coverage upload, so the split that #2167 works around stops existing. If this lands first, #2167 can be closed instead of merged.


paskal:
Copilot's point is fixed: `frontend/CLAUDE.md` still counted `ci-frontend-api.yml` among the places the pnpm version is pinned, and asserted a fixed total that no longer holds. Both corrected, and a sweep for `packages/api`, `@remark42/api` and `ci-frontend-api` across md, yml and json now returns nothing.

### Deprecating the npm package

Removing it here does not remove it from the registry: `@remark42/api` stays installable, `dist-tags.latest` still points at `0.6.0-alpha.12` from 2022-07-11, and anyone who finds it will get a client that cannot authenticate. Whoever owns the npm package can close that off with one command, which needs no publish and no new code:

```
npm deprecate "@remark42/api@*" "Unmaintained and removed from remark42. There is no client SDK; use the REST API directly: https://remark42.com/docs/contributing/api/"
```

The `@*` range covers all 13 published versions rather than only the latest. npm then shows that message on every install and on the package page, which is the redirect we want.

Worth noting the message points at a page currently filed under *Contributing*. If the REST API is the supported integration path, and #1383 says it is, that page belongs in the integration docs, and the deprecation text should be updated to the new URL when it moves.

umputun (approved):

PR #2175, opened 2026-08-20
Replace react-redux with a preact context binding

paskal:
First step of #2166. `react-redux` is replaced by a small binding over Preact context, in `app/store/context.tsx`.

No behavioural change: the store, the reducers, the thunks and every selector stay exactly as they were. This only changes what connects components to the store.

It is not free, though, it is a saving. Measured against current master, `remark.mjs` goes from 77.64 kB to **76.15 kB** gzipped and `last-comments.mjs` from 38.81 kB to **37.97 kB**, so roughly 1.5 kB and 0.8 kB. The size limits are left where they are, so this shows up as headroom rather than a tighter gate.

### Why this one first

`react-redux` is one of exactly two packages holding the `@preact/compat` alias in place, and it is by far the more contained of the pair. The store's own logic is plain `redux` plus `redux-thunk`; the only react-redux import inside `app/store` was a single line in `index.ts` re-exporting typed hooks. Everything else is 9 files across components and hooks. `@types/react-redux` goes with it, having no remaining source usage.

It also carries a real react into the tree: `react-redux` pulls `use-sync-external-store`, which resolves an actual `react` in the lockfile rather than the compat shim.

### What the binding provides

Only what the widget uses, which is less than react-redux offers:

- `Provider`
- `useDispatch`
- `useSelector(selector)` and `useSelector(selector, equalityFn)`, the second form used once, in `thread.tsx`
- `shallowEqual`
- the `TypedUseSelectorHook` type that `store/index.ts` re-exports as `useAppSelector`

The subscription keeps the selector, the equality function and the last selected value in refs, so the subscription itself stays stable across renders while always comparing against the current render's selector. A changed selector therefore cannot resurrect a stale result.

It subscribes in `useLayoutEffect` rather than `useEffect`, and runs the comparison once immediately on subscribing. Both are needed for the same reason react-redux does them: an effect that runs after paint leaves a window in which a dispatch is not delivered, because the listener does not exist yet, and the component then renders a stale value until some later unrelated dispatch happens to differ from it. `context.test.tsx` covers that case and fails without the immediate check.

### Tests

The four test files that imported `Provider` from `react-redux` now import it from the same place the app does, and `comment-form__subscribe-by-rss.test.tsx`, which mocked the `react-redux` module, mocks `store/context` instead. `redux-mock-store` is unaffected and stays, since it mocks the store rather than the bindings.

The binding also gets its own tests, in `app/store/context.test.tsx`, which it did not have when it was a dependency: initial value, re-render on dispatch, and the subscribe-window case described above.

323 tests across 40 suites pass. The 320 pre-existing ones are unchanged and exercise the connected components heavily, which is the point: no test was adjusted to accommodate the new binding.

### Verified

Rebased onto current master, then verified from a wiped `node_modules`: `pnpm install --frozen-lockfile`, `pnpm lint`, `pnpm type-check`, `pnpm test` (323 of 323), `pnpm build`, `pnpm size-check` and `pnpm translation-check`.

### Next

`react-intl` is the other half, and the larger one at 27 non-test files. Once both are gone the `react` and `react-dom` aliases, the `@preact/compat` dependency and the `paths` entries in `tsconfig.json` can be deleted, which is what #2166 is actually after.

### DIFF
--- a/frontend/apps/remark42/app/components/auth-panel/auth-panel.test.tsx
+++ b/frontend/apps/remark42/app/components/auth-panel/auth-panel.test.tsx
@@ -1,7 +1,7 @@
 import { mount } from 'enzyme';
 import createMockStore from 'redux-mock-store';
 import { Middleware } from 'redux';
-import { Provider } from 'react-redux';
+import { Provider } from 'store/context';
 import { IntlProvider } from 'react-intl';
 
 import type { User } from 'common/types';
--- a/frontend/apps/remark42/app/components/auth/auth.tsx
+++ b/frontend/apps/remark42/app/components/auth/auth.tsx
@@ -2,7 +2,7 @@ import clsx from 'clsx';
 import { h, Fragment, JSX } from 'preact';
 import { useState, useRef } from 'preact/hooks';
 import { useIntl } from 'react-intl';
-import { useDispatch } from 'react-redux';
+import { useDispatch } from 'store/context';
 
 import { setUser } from 'store/user/actions';
 import { Input } from 'components/input';
--- a/frontend/apps/remark42/app/components/comment-form/__subscribe-by-email/comment-form__subscribe-by-email.test.tsx
+++ b/frontend/apps/remark42/app/components/comment-form/__subscribe-by-email/comment-form__subscribe-by-email.test.tsx
@@ -1,6 +1,6 @@
 import { mount } from 'enzyme';
 import { act } from 'preact/test-utils';
-import { Provider } from 'react-redux';
+import { Provider } from 'store/context';
 import { Middleware } from 'redux';
 import createMockStore from 'redux-mock-store';
 import { IntlProvider } from 'react-intl';
--- a/frontend/apps/remark42/app/components/comment-form/__subscribe-by-email/comment-form__subscribe-by-email.tsx
+++ b/frontend/apps/remark42/app/components/comment-form/__subscribe-by-email/comment-form__subscribe-by-email.tsx
@@ -1,6 +1,6 @@
 import { h, FunctionComponent, Fragment } from 'preact';
 import { useState, useCallback, useRef } from 'preact/hooks';
-import { useSelector, useDispatch } from 'react-redux';
+import { useSelector, useDispatch } from 'store/context';
 import clsx from 'clsx';
 import { useIntl, defineMessages, IntlShape, FormattedMessage } from 'react-intl';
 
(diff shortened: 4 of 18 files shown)


paskal:
@umputun, this one is good to go, simplification PR. Next one will cut react out, then I'll start cutting code into static files.

umputun (dismissed):
approving. Read the binding and every call site, and had codex go over it independently as a second pass. We come out the same way: no missed update, no stale value, no crash path. CI green.

two things worth recording, neither blocking.

**one extra render at mount, per component whose selector builds an object**

the immediate `checkForUpdates()` after subscribing is the right idea, and react-redux does the same check (`use-sync-external-store-shim.development.js:172`). What differs is what it calls. React-redux's `getSnapshot` there is the memoized selector from `with-selector.js:101-106`, which returns the previous selection unchanged when the state reference is the same, so `Object.is` holds and nothing fires. This one calls the raw selector, and a selector building a fresh object fails `Object.is` every time.

so at mount: render 1 computes `O1`, the layout effect subscribes and immediately recomputes `O2`, `Object.is(O1, O2)` is false, `setTick`, render 2. Exactly one extra render per component, and the effect deps are `[store]` so it does not repeat. It hits `ConnectedRoot` and every `ConnectedComment`, both of which build a fresh object with no `equalityFn`, so a 200-comment thread mounts ~201 extra component renders.

guard is small and keeps the property the check is there for:

```ts
const state = store.getState();
const selected = selector(state);
const stateRef = useRef(state);
stateRef.current = state;

useLayoutEffect(() => {
  const checkForUpdates = () => { /* unchanged */ };
  const unsubscribe = store.subscribe(checkForUpdates);
  if (store.getState() !== stateRef.current) checkForUpdates();
  return unsubscribe;
}, [store]);
```

any dispatch that could have been missed in that window produced a new root state object under redux's immutability contract, so nothing real is dropped. Every reducer here holds to that: `comments` copies before changing (including `COMMENTS_PATCH`'s lazy copy), `childComments` builds a new map, `thread` and `user` copy, `cmpRef` returns either the untouched old value or the new input. React-redux misses the in-place-mutating-reducer case too, so this is parity rather than a step back. Your race test still exercises the guarded path, since `bump` returns a new state object.

worth one focused test next to it: object-returning selector plus a render counter, mount with no dispatch has to stay at a single render.

**selector exceptions now escape from `dispatch`**

`checkForUpdates` runs the selector inside the store listener. React-redux ran it there too, but through `checkIfSnapshotChanged` (`use-sync-external-store-shim.development.js:201-211`), whose `catch` returns `true` and converts a throw into a forced render. Here a throwing selector breaks whoever called `dispatch`.

not reachable today. `allComments` never deletes keys, since `COMMENTS_SET` rebuilds from `{...state}`, so `thread.tsx`'s `getThreadIsCollapsed(comment)` still gets a real comment even on a stale id, and `ConnectedComment` reads `props.data`. Recording it as a property that went away, not asking for anything.

PR #2176, opened 2026-08-21
Replace react-intl and remove React from the widget

paskal:
Second and final step of #2166. `react-intl` is replaced by `app/common/intl.tsx`, a small i18n binding over Preact context, and with `react-redux` already gone nothing holds the React compatibility alias.

React is now absent from the lockfile, the installed tree, the config and the bundles: `react`, `react-dom`, `react-intl`, `@types/react`, `@preact/compat`, `use-sync-external-store` and `intl-messageformat` are all gone, along with the `paths` entries in `tsconfig.json` and three babel-loader excludes. Runtime dependencies go from 15 to 10.

`preact/compat` goes too, which matters more than its 3.8 kB. Importing it anywhere installs hooks on preact's shared `options` that remap `onFocus`/`onBlur` to `focusin`/`focusout` for every element and make `@testing-library/preact` rewrite `change` to `input`, the two bugs behind #2166, still live until now. `Button` was wrapped in `forwardRef` with no caller passing one, and `TextareaAutosize` now takes its ref as an ordinary prop. The workaround in `sort-picker.spec.tsx` is gone with them, since `fireEvent.change` reaches a `<select>` again.

Gzipped, against master: `remark.mjs` 76.17 kB to 56.47, `last-comments.mjs` 37.97 to 18.26, `deleteme.mjs` 14.51 to 8.42. The limits move with them and keep more relative headroom than master shipped.

### The binding

`IntlProvider`, `useIntl`, `createIntl`, `defineMessages`, `FormattedMessage` and `IntlShape`. 32 files change only their import.

The export names copy react-intl's deliberately: `formatjs extract` finds messages by recognising `defineMessages`, `FormattedMessage` and `intl.formatMessage` in the AST rather than by import source, so renaming one silently empties the catalogue. `frontend/CLAUDE.md` records that, along with the destructive part: `translation:generate` would then strip the unextracted keys from all 24 catalogues and the check would pass.

A message the binding cannot parse falls back to the message in the source: a broken, unhandled or nested tag, a brace that is not a well-formed placeholder, and a placeholder naming a value the caller did not supply. `mk.json` and `th.json` carried broken markup and rendered in English; both are repaired, so a catalogue sweep over every locale can now require well-formed markup with no exceptions listed.

`translation:check` gained the validation that would have caught them when they were proposed: a translation's tags have to be well-formed pairs of the names the English string uses, with no attributes, and its placeholders have to be ones the English string provides. Leaving a tag or a placeholder out stays allowed. Run against master's catalogues it reports both.

### enzyme

`@types/enzyme` was the last thing pulling `@types/react`, so React could not leave while enzyme stayed. Its three test files move to `@testing-library/preact`, which now has no rival: `@testing-library/preact-hooks` had one import left and its own unmet peer warning. `intersection-observer` was a runtime dependency nothing imported, and the `cheerio` override lost its last dependent with enzyme.

Enzyme's `.find(X).prop()` threw unless exactly one node matched, so the converted tests assert node counts explicitly to keep that.

### Verified

All 181 message ids formatted across all 24 catalogues through both real react-intl and this binding: 4344 comparisons, no differences. From a wiped `node_modules`: `pnpm install --frozen-lockfile`, `pnpm lint`, `pnpm type-check`, `pnpm test` (392 tests, 42 suites), `pnpm build`, `pnpm size-check`, `pnpm translation-check`.


paskal:
All four Copilot findings were correct, and two of them were real defects in the checker this PR adds.

`checkTranslation.js` tested for any angle bracket in the residue, so ordinary translated text containing `<` or `>` was rejected: "Под < 10 комментариев" and "a > b" both failed. It also discarded a pair's contents wholesale, so a nested `<a>text <b>x</b></a>` passed even though the binding treats a nested tag as invalid and falls back to English. Both are fixed by mirroring `intl.tsx`'s rule: only `</` or `<` opening a tag name counts as markup, self-closing tags are text, and the pair contents are inspected alongside the text around them. Verified against six cases, three that must pass and three that must fail.

The two documentation findings were the same staleness in two places. Both files said nothing validates catalogue values, which was true when written and stopped being true in this PR. They now describe what each check covers, and name the gap that remains: unsupported ICU syntax is ordinary text to both the checker and the sweep, so neither can catch it.

umputun (dismissed):
approving. The checker fix is right. Seven values injected into `ru.json` and run through the real `checkTranslation.js`: nested, attribute and unclosed all reject; well-formed, bare `<`, bare `>` and a trailing `<br/>` all accept. Seven for seven, and it agrees with the binding on every case now.

two leftovers, both one line in `app/common/intl.test.tsx`, both the same thing you just fixed in the checker.

**`isWellFormed` still tests for any angle bracket**

`intl.test.tsx:276` is `!/[<>{}]/` over the stripped residue, which is the rule you removed from `checkTranslation.js:71`. So the two now disagree in the other direction: the checker accepts `"Под < 10 через <a>Remark42</a>"` and the sweep rejects it, and the same goes for a trailing `<br/>`. Reach is narrow, since `isWellFormed` runs on `root.powered-by` and `commentForm.notice-about-styling` only, but a valid future translation of either fails the required Jest job, and your own reason for the checker fix applies here word for word.

dropping the precheck is probably smaller than aligning it. The `expect(rendered).toBe(withoutTags(translated))` assertion below already catches broken markup for those two ids: on a nested value the binding falls back, so `rendered` is `Powered by Remark42` while `withoutTags(translated)` is `Работает на <b>Remark42</b>`, and they differ. Generic structural validation now lives in `translation:check`. Aligning `isWellFormed` to `TAG_START` works too if you want the assertion to stay independent of the binding, but that is a third copy of the same parser to keep in step.

**the comment above it is now false**

`intl.test.tsx:279` still says `checkTranslation.js compares key sets and never looks at a value`. Same staleness you fixed in `frontend/CLAUDE.md` and the translations guide, and this copy sits directly above the sweep that now overlaps the checker, which is where someone would go to work out which check owns what.

nothing else, and the rest of the review is unchanged since none of it moved between commits.

no XSS surface: a tag does something only when the caller supplies a function for that name, so catalogue text reaches the page as a text node or as an argument to a handler you wrote, and a translator cannot produce an element or an attribute. No ICU plural, select, selectordinal, number or date construct appears in any of the 24 catalogues or in any `defaultMessage`, and no comma sits inside braces anywhere, so the documented gap is not reachable today. Running the binding's own parse functions over every catalogue entry gives 4344 entries and zero fall back, which reproduces your 4344 from the other direction. Test case counts survive the enzyme conversions intact, 4, 10 and 2 respectively, and `useSessionState` gains one.

paskal:
Both fixed, and dropping the precheck was the right call over aligning it.

`isWellFormed` is gone rather than rewritten. The assertion below it already covers those two ids: on a broken value the binding falls back to English while the expectation is the locale's own words, so they differ. Verified across six shapes, three that must fail and three that must pass: nested, unclosed and unhandled-tag all still reject, and `"Под < 10 через <a>Remark42</a>"` and a trailing `<br/>` now pass where the old rule rejected them. That leaves two parsers to keep in step instead of three, with generic structural validation owned by `translation:check`.

The comment above it now says which check owns what, since that is the question someone lands there with.

umputun (approved):

PR #2177, opened 2026-08-21
Complete and correct every translation catalogue

paskal:
Every catalogue is now fully translated, and the defects an audit of all 24 of them turned up are fixed. 309 strings across 19 languages, no change to any English source string.

## Completing the catalogues

No locale carries English text any more, with a short list of deliberate exceptions below.

The Telegram subscription block, six keys, had never been translated in any of the 16 locales that ship it, and `auth.telegram-message-1/2/3` were English in nine more. Those three carried an older source wording rather than the current one, so a check for "value equals the English string" could not see them. Finnish was the largest single gap at 45 strings.

Two of the six Telegram subscription keys are word for word the English of their `subscribeByEmail` siblings, so each catalogue's own existing wording was reused rather than a second phrasing invented for the same sentence.

Left identical to English on purpose, because the word is the same in that language: `RSS` and `Telegram` everywhere, "Email" in be, it, pl, ro, ua and vi, "Site" in bp, fr and ro, and "Conflict." in ro. German and Turkish do not use bare "Site" and say "Website" and "Web sitesi".

## The Telegram authorisation paragraph

That paragraph is assembled in `telegram-link.tsx` from five keys, with the anchor and the QR clause in the middle, so a catalogue that translates each key in isolation produces word salad in any language whose verb does not sit where English puts it. Japanese rendered as `テレグラムを開く リンクで または QR コードをスキャン そこで「スタート」を...`, and Korean, Persian, Traditional Chinese and Czech broke the same way. Each of those now splits the sentence where its own grammar wants it, and every catalogue was checked by rendering both layouts, since the QR clause only appears above 768px.

Ten catalogues also had `auth.telegram-link`, which is the anchor text and reads "by the link", left as "Telegram bot" from an older source wording, so the paragraph named the bot twice and never said what the link was for.

Quotation marks in that paragraph now follow each language rather than the English source: «» for be, ua, ru, fa, fr and ar, „“ for bg, cs, de and mk, „” for pl and ro, ”” for fi, 「」 for ja and zh-tw. The button it tells the reader to press is Telegram's own, and Telegram ships no interface translation for Japanese, Thai or Vietnamese, so those three now name it the way Russian and Traditional Chinese already did, with the Latin label alongside the translated one.

## The error-code collision

`errors.8` is `ErrReadOnly` (`backend/app/rest/httperrors.go:29`), but 13 catalogues carried a copy of `errors.7`, which is `ErrUserBlocked`. A reader who simply hit a read-only thread was told they had been blocked, in Belarusian, Bulgarian, Brazilian Portuguese, German, Finnish, French, Japanese, Polish, Russian, Turkish, Ukrainian, Vietnamese and Simplified Chinese. Each now says the page is read-only, in the terminology that catalogue already uses for its own read-only badge.

Czech had an off-by-one: `errors.19`, restricted words, carried the text of `errors.18`, file not found, and `errors.18` was left in English. A comment caught by the word filter therefore reported a missing file. Both are rewritten, and Czech is the only locale with that shift.

## Other defects

- `de` `vote.downvote` had a leading space, and `mk`, `pl` and `vi` had double spaces
- `ua` `errors.9` was missing a space after a full stop
- `vi` wrote "Bằng đã huỷ đăng kí" for "Bạn", `es` used the obsolete accented "ó" for "o", `th` misspelled "คลิ๊ก" for "คลิก"
- `cs` opened the Telegram paragraph with the infinitive "Otevřít" where the rest of it is imperative, `de` alternated between tippen and klicken for the same action, and `mk` had its anchor text in caps
- `ar` `commentForm.upload-file-fail` had an unclosed quotation mark
- `fr` `auth.telegram-check` was "Sélectionner" on a button the surrounding text calls "Vérifier", and `bp` had "Verifique" against its own "Verificar"
- `mk` had dropped `{shortcut}` from the bold, italic and link tooltips, losing the keyboard hints
- `fi` `comment.pin` was "Sitoo", which means "it binds", and `comment.unpin` followed from it; `errors.forbidden` was misspelled "Kieletty."
- `be` `errors.failed-fetch` ended in a stray "r"
- `zh-tw` `authPanel.read-only` wrote 唯獨 for 唯讀
- `ja` `errors.conflict` used 相衝, which is not Japanese usage
- `it` `auth.symbols-restriction` misspelled "numberi" and used "username" where `auth.username` now says "Nome utente"
- `vi` `auth.telegram-message-3` misspelled "đaay"

## How these were found

An audit of all 24 catalogues for defects a key-set comparison cannot see: values shared between keys whose English differs, placeholders dropped or invented, stray whitespace, and links or self-references. That is what surfaced the `errors.7`/`errors.8` collision, and reading Czech for style before rewriting it surfaced the off-by-one.

The markup validation added to `translation:check` in #2176 would have caught none of them: a duplicated error string and a stray space are both well-formed data.

## Review

Independent passes over the translations themselves, judging register and terminology against each catalogue rather than grammar alone. They reworded fifteen strings, including a wrong Belarusian genitive plural, four cases where a subscribe button had drifted from its own catalogue's register, and an Italian anglicism, and they disagreed with each other on the Finnish upload message in a way that produced a better third version.

Verified with `pnpm translation-check` and the full test suite, plus a sweep confirming no locale still shares `errors.7` and `errors.8`, no placeholder is dropped or invented anywhere, no value carries stray whitespace or broken markup, and every catalogue still holds all 181 keys.


umputun (approved):
approving. Two things to raise, both pre-existing rather than introduced here, and one limit at the end.

**the catalogues themselves**

24 catalogues, 181 keys each, key sets unchanged against base and equal to `en.json`, nothing added, removed or moved. Repeated keys scanned over the raw text rather than through `JSON.parse`, which keeps only the last occurrence: none. No placeholder invented or dropped anywhere, which is the independent confirmation that mk's three `{shortcut}` tooltips are back. No stray leading, trailing or doubled whitespace left in any value.

five changed values carry markup, all `root.powered-by` in be, fi, fr, it and ua, each preserving `<a>Remark42</a>` character for character. Nothing in the diff adds or removes an attribute, an href, a URI scheme, a markdown link or a bidi control. 44 `fa` entries carry ZWNJ `U+200C`, all present before this PR and none touched, which is Persian joining. Those null results are worth something only because the same patterns were first shown to catch known positives, including the `<a href='diosfera.codeberg.page'>` string that used to sit in `mk` `root.powered-by`.

`errors.7` equalled `errors.8` in exactly the 13 locales you name and in none after. The Czech shift is as described. Quotation codepoints match your mapping for all 15 named locales, and every item under "Other defects" reproduces against base and head.

**mk labels the replies feed as comments**

`mk` `subscribeByRSS.replies` carries `Коментари`, the same value as `user.comments`, in a catalogue where `comment.reply` and `commentForm.reply` are both `Одговори`. That option subscribes to `/rss/reply?user=`, and `UserReplies` at `backend/app/store/service/service.go:567` is documented as comments replied to that user, so the feed is replies and the label says comments. The catalogue collapses two distinct concepts, in the same class as `errors.7`/`errors.8` and in the same audit's blind spot. Strong internal evidence rather than a Macedonian ruling, so worth a look rather than settled.

re-running that audit takes it from 97 cross-key duplicates to 84, the 13 removed being exactly these fixes. Most of the remainder is real ambiguity: German `Antworten` covers both reply and replies, `Abmelden` covers both sign out and unsubscribe, likewise Bulgarian `Отговори` and Vietnamese `Trả lời`.

**the Japanese and Chinese paragraphs still render with spaces in them**

the grammar fix works. `ja` goes from `テレグラムを開く リンクで または QR コードをスキャン そこで…` to `テレグラムを … 開き、そこで…`, which is correct SOV, and `zh` and `zh-tw` read correctly end to end now.

`telegram-link.tsx:23` and `:27` hardcode `{' '}` between the fragments, so the assembled sentence keeps those spaces:

```
ja  テレグラムを リンクから または QR コードで 開き、そこで「スタート」（Start）を…
zh  通过 此链接 或扫描二维码 打开 Telegram，然后点击“开始”。
```

neither language uses inter-word spaces. In the `zh` line one space separates a preposition from its object and another separates an adverbial phrase from its verb. Korean is unaffected, since Korean does use spaces. The remaining half of this fix lives in the component rather than in any catalogue, which puts it outside this PR, so I mention it because you write that both layouts were rendered and checked.

**what this review does not cover**

the languages I can read came out clean, and `Sitoo` really is the third person of `sitoa`, so that rewrite is right. I cannot vouch for Thai, Persian or Finnish register beyond the specific items above. Two claims sit outside anything checkable here: whether the quotation conventions suit each language is a judgement rather than a codepoint comparison, and Telegram shipping no interface translation for Japanese, Thai or Vietnamese is a claim about their product.

PR #2178, opened 2026-08-21
Consolidate the frontend toolchain onto babel, and ship one bundle

paskal:
Four toolchain upgrades that were finished but never merged, the compiler collapse they enable, and the dependency sweep that follows. Direct devDependencies go from 77 to 59 and dependencies from 10 to 9.

## One compiler

Three were doing the same job: `ts-loader` stripped types in webpack, `babel-loader` did everything else, and `@swc/jest` repeated both for the tests with its own copy of the JSX settings. Babel is the one that survives, because the `data-testid` stripper has no equivalent elsewhere.

`ts-loader` ran `transpileOnly: true`, so it only stripped types, which `@babel/preset-typescript` does; `fork-ts-checker-webpack-plugin` was already what type-checks. Jest runs `babel-jest` against the same `.babelrc.js` the bundle uses, passed as `configFile` because a file-relative babel config does not reach the `node_modules` packages in `transformIgnorePatterns`. `jest.config.mjs` is plain ESM rather than TypeScript, because a `.ts` config is compiled against `tsconfig.json`, whose `verbatimModuleSyntax` rejects ESM syntax in a file the package has not declared as a module, and which loader jest picks for it varies by environment.

That removes `ts-loader`, `@swc/jest` and `@swc/core`. The last was pinned to 1.2.205 from 2022 with no way forward, because newer builds emit non-configurable exports and break `jest.spyOn` across 13 suites.

Babel compiles a file at a time with no type information, so it cannot tell a type-only import from a real one and keeps the module. `verbatimModuleSyntax` and `@typescript-eslint/consistent-type-imports` mark them properly across roughly 176 imports; the statement has to be a separate `import type`, since verbatim semantics keep an inline `import { type X }` and load the module anyway.

## One bundle

The legacy and modern compilations produced the same bytes. Both read the same browserslist query, `defaults, not IE 11, not samsung 12` resolves to chrome 109 and up, and nothing in the source needs transforming for that set, so 28 of the 29 output pairs were byte-identical.

That made the module/nomodule switch worse than redundant: it served the `.js` file to browsers with no ES module support, and those files carried `??`, `?.` and class fields, so the fallback handed its own audience a syntax error. There is now one bundle, always loaded as a module, in the five templates and in every `site/` document integrators copy from. A production build emits 29 files rather than 58, in about 3 seconds rather than 17.

Two of those documents did not work at all beforehand: the SPA snippet could not parse, because `var host = // comment` swallowed the following statement and the wrapper was never invoked, and the subdomain example had an unterminated string.

## Node 24

`@babel/core` 8 declares `^22.18 || >=24.11` and `size-limit` 13 declares `^22.18 || ^24 || >=26`, so 20 was below the floor of two things installed here; pnpm only warns, which is why every build passed. The branch goes to 24 rather than the floor, because 24 is the active LTS and 22 has dropped to maintenance. All seven places the frontend pins it move together, including the contributing guide's prerequisite. Verified on `node:24-alpine`: install, type-check, 392 tests, lint and a production build.

## Rule coverage

`eslint --print-config` before and after gives 173 active rules on an application file against 172, and 172 on a spec file and a plain JS file against 171. What is gone is three `flowtype` rules with no Flow here, `no-new-object` and `no-new-symbol` whose upstream replacements are on, `react/forbid-foreign-prop-types` with no propTypes anywhere, and, on TS only, `no-useless-constructor`, whose typescript-eslint version is on at error.

`@babel/core` is pinned to 8 across the workspace because `@jest/transform` and `istanbul-lib-instrument` depend on 7 outright. A second scoped override holds `eslint-config-preact` on 7, since its `@babel/eslint-parser` loads babel 7 syntax plugins that babel 8 rejects.

## Also removed

`fast-async` rewrote every async function into nodent promise chains, calls babel's `transform` synchronously, which babel 8 removed, and every browser in the target list runs async natively. `prefresh` blew its stack on `createContext` under babel 8 with no newer release to move to, which compiled `intl.tsx` and `store/context.tsx` into throwing stubs, so `pnpm dev:app` could not run the widget at all; without the babel plugin the other two packages do nothing, so all three go and fast refresh with them. `core-js` is not injected now that `useBuiltIns` is gone, `postcss-custom-properties` was reached directly although nothing declared it and resolved only through pnpm's private hoist directory, and `cssnano` ran in both postcss chains although `CssMinimizerPlugin` already uses it. `url-loader` ran with `limit: false`, so it delegated every asset to `file-loader` and did nothing else; the rule now names `file-loader` directly and `url-loader` goes. Also gone: `@babel/plugin-transform-react-jsx`, `ts-node`, and a dead workspace `lint-staged` script.

`CleanWebpackPlugin` now runs for production too. It was development-only, so a production build over an existing `public/` left the previous dual build's output in place, and the Dockerfile copies that directory into the Go embed.

`pnpm lint`, `pnpm test` and `pnpm build` work from `frontend/` as `CLAUDE.md` and the contributing guide have always said they do; the workspace root defined none of them.

## Verified

`pnpm install --frozen-lockfile` from a clean `git archive` export, then eslint, stylelint, `tsc --noEmit`, 392 tests, coverage, all seven size budgets, the translation check, and a production build with no warnings. `grep -c data-testid public/*.mjs` is 0. Size budgets are pinned back to gzip, which `size-limit` 13 had switched to brotli, leaving every limit measured against a smaller number than it was written for.

The widget was driven in Chrome against the built bundle in a container: rendering in Russian off the dynamically loaded locale chunk, anonymous sign-in, posting with markdown, editing, deleting through its confirm dialog, sort and draft persistence across reload, and the lazily-loaded emoji chunk. No console output at any level.

`codecov/project` is red and expected to stay red, though not for the reason a coverage drop usually has. Codecov reports files 138 to 119, lines 3284 to 2836, hits 2202 to 1730 and misses 1076 to 1100: nineteen files left the report rather than joined it. Those are the barrels and type-only modules babel compiles away to nothing, which the previous setup counted as covered. Patch coverage is clean and nothing among the vanished paths carries behaviour that wants a test.


paskal:
Copilot's real findings from both rounds are fixed: the `deleteme.html` bundle reference, the two published integration snippets, the `subdomain/index.md` string, the contributing guide's Node version, and `CleanWebpackPlugin` not running in production.

The remaining fourteen do not hold, and I am resolving them. Evidence for each class below.

**Twelve `rgb(var(--x), <alpha>)` comments.** Comma-form `rgb()` with an alpha argument is valid CSS Color 4, not a mistake for `rgba()`. The two forms are the same declaration to a browser. Computed in Chrome against this branch's own custom properties:

| declaration | `getComputedStyle` |
| --- | --- |
| `background-color: rgb(var(--white-color), 0.5)` | `rgba(255, 255, 255, 0.5)` |
| `background-color: rgba(var(--white-color), 0.5)` | `rgba(255, 255, 255, 0.5)` |
| `border-left: 1px solid rgb(var(--primary-color), 0.4)` | `rgba(66, 133, 244, 0.4)` |
| `box-shadow: 0 0 0 2px rgb(var(--primary-color), 0.4)` | `rgba(66, 133, 244, 0.4)` |

Borders and shadows apply, and nothing is dropped. The browserslist query resolves to Chrome 109+, Safari 26.3+ and Firefox 140+, all of which have supported the syntax for years. The rewrite came from stylelint 17's `color-function-notation` rule, which is what the upgrade brings.

**`import.meta.dirname` in `jest.config.mjs`.** Standard since Node 20.11 and 21.2. This branch requires Node 22 and builds on `node:22-alpine`, where `node --input-type=module -e "console.log(import.meta.dirname)"` prints a path rather than `undefined`.

**Two `import type` comments.** `typeof getPreview` and `Partial<typeof boundActions>` at `comment.tsx:54-57`, and `keyof typeof messages` at `auth.utils.ts:22`, are all in type positions, where a type-only import is correct. The value references at `comment.tsx:292` and `:526` read `this.props.uploadImage`, which is the prop rather than the import. `tsc --noEmit` passes from a clean `git archive` export.

umputun (dismissed):
approving. Two sentences in the description do not match what the tree does, and one small thing about the stripper.

**the coverage paragraph has it backwards**

it says 18 entry files the previous setup omitted now appear, so the percentage falls with identical tests. Codecov's own numbers for this PR go the other way: files 138 to 119, lines 3284 to 2836, hits 2202 to 1730, misses 1076 to 1100. Nineteen files left the report rather than joining it, and they are the barrels and type-only modules babel compiles away to nothing, which were previously counted as covered. That is what drags the average down, along with 24 more misses than before.

the conclusion still stands, since patch coverage is clean and nothing among the vanished paths carries behaviour that wants a test. But the stated reason is the opposite of what happened, and it is the sentence someone will read later when deciding whether a red project check is acceptable.

**`file-loader` is still installed and still runs**

it is under "also removed", and it is gone as a direct dependency. The rule at `webpack.config.js:165` is still `url-loader` with `limit: false`, which takes the fallback branch for every asset, and url-loader's default fallback is `require('file-loader')`. The lock still has `file-loader@6.2.0` as its optional peer. Nothing is broken under the frozen graph, so this is a wording question, unless the intent was actual removal, in which case webpack 5's `asset/resource` replaces that rule outright.

**the vendored stripper**

`tasks/babel-plugin-remove-test-id.js` matches a literal `JSXAttribute` named `data-testid`. Every use in the tree is that shape, so it covers everything today, and `grep -c data-testid public/*.mjs` catches it if that stops being true. The package it replaces also handled a spread, so `{...{'data-testid': x}}` would now ship. Worth a line in `frontend/CLAUDE.md` next to the grep, rather than a change.

**the bundle collapse**

checks out, and the fallback was worse than redundant. `defaults, not IE 11, not samsung 12` resolves to chrome 109 at the oldest, which has `??`, `?.` and class fields, so preset-env never downlevelled them and the `.js` file carried all three. The only clients that would take the `nomodule` path are `op_mini all`, `kaios 2.5`, `and_uc` and `and_qq`, none of which parse any of it. They now ignore a module script instead of choking on one, which is not a loss.

umputun:
#2181 merged ahead of this, so it wants a rebase. Two files conflict: `frontend/apps/remark42/package.json` and `frontend/pnpm-lock.yaml`. `webpack.config.js` auto-merges even though both touched it, and so do `.stylelintrc.js`, the two `CLAUDE.md` files and the contributing page.

the lock is the one worth regenerating rather than resolving by hand.

paskal:
Addressed the feedback + rebased on top of master to resolve conflicts.

umputun (approved):

PR #2190, opened 2026-08-22
Make backend tests wait on conditions instead of durations

paskal:
The backend workflow has a long tail of runs that fail once and pass on a rerun. Going through every retained CI log of that workflow, plus the history of the tests we have complained about before, the failures are all the same shape: the test assumed an operation would finish within some duration instead of waiting for the state it actually needed.

Three were reproducible, and each was reproduced against the old code before being changed.

`TestServerAuthHooks` minted a token with a one-second lifetime and never tested expiry, so a slow runner turned the first POST into a 401. Stalling 1.1 seconds between minting the token and using it produces the exact CI output.

`TestServerApp_AnonMode` failed with `connection refused` on a ping, because `waitForHTTPServerStart` returned silently after a three-second budget and left a later assertion to fail with something unrelated.

`TestFsStore_Cleanup` slept 200ms against a 300ms TTL, which `Cleanup` widens to 400ms with its commit grace. The second staged image was already 300ms old by then, so roughly 100ms of stall anywhere in the setup aged it past the line and it was collected. Injecting a 250ms stall reproduces it; the version here survives 2 seconds.

`TestRest_CreateWithPictures` and `TestServerApp_MainSignal` are both on the list of tests we were chasing back in 2019 and 2020, and `TestRest_CreateWithPictures` has failed in both directions over the years, once for moving a picture too early and once for not moving it in time. Each round of fixes raised a number: the staging TTL went from 50ms to 100ms, the sleep from 800ms to 1500ms. That is the approach this change replaces.

## What changed

Fixed sleeps before asserting on asynchronous work are replaced with polls on the condition itself, using `require.Eventually` and `require.EventuallyWithT`, and `require.Never` where the assertion is that something did not happen. Polling closures assert on the `*assert.CollectT` they are handed rather than on `t`, since testify runs them on a separate goroutine, and the interval for polls that issue HTTP requests stays under the rate limit on the routes they poll through.

Where a test genuinely needs time to have passed, the clock input is pinned rather than waited for. Staging ages are set with `os.Chtimes` immediately before each `Cleanup` call, and the RSS tests set `store.Comment.Timestamp` explicitly instead of racing the wall clock into the first 100ms of a second so that `pubDate` matches.

`chooseUnusedPort` takes a port from the kernel's ephemeral range. It previously picked at random out of a fixed 10000-port window, so two package test binaries, which `go test ./...` runs concurrently, could pick the same number between the probe closing and the server binding. The start helpers fail naming the port they waited on rather than returning silently, and the SSL tests wait on the redirect port as well as the TLS port before using it. Because a failing `require` exits the test goroutine, anything that starts a server in a goroutine now cancels it on the way out, so a failed readiness wait is reported as itself rather than as a goroutine leak.

Arbitrary wall-clock budgets that nothing tests are gone: ten HTTP clients with a one-second timeout against Bolt-backed import and export, the "should take about 100msec" assertions, and a one-second bound on an app noticing an already-cancelled context. Shutdown is still bounded, at ten seconds, so a hang is caught without depending on how loaded the runner is.

Two assertions got stronger rather than weaker. `TestServerAuthHooks` accepted either 403 or 401 from a blocked user, an alternative that only existed because the short-lived token could expire mid-test; it is now deterministically 403. `TestAdmin_BlockedList` asserted two users blocked while one of them carried the same 150ms TTL the next step waits to lapse, so the two halves of the test raced each other.

## goleak

`goleak` no longer reports the `regexp2` clock goroutine, which chroma pulls in for syntax highlighting. regexp2 keeps one shared clock alive for up to a second after the last match with a timeout, waking every 100ms; it ends on its own, but a binary that finishes inside that window is reported as leaking, and this suite now finishes sooner. All three packages running `VerifyTestMain` ignore it.

The other ignores were checked rather than assumed. The one for `net/http.(*Server).Shutdown` stays: the goroutine at `server.go:854` calls `restSrv.Shutdown()` on context cancellation, `Rest.Shutdown` gives `httpServer.Shutdown` a one-second context, and `serverApp.Wait()` waits only on `terminated`, so it is genuinely unjoined. golang-lru's expirable cache goroutine is documented in the vendored source as never exiting and trips all three packages immediately, and the SIGQUIT dump handler started by `app`'s `init` is deliberately permanent.

## Example module

`_example/memory_store` had a flake of its own, unrelated to the rest of this but in the same CI step. Its teardown calls `s.Shutdown()`, which gives `http.Server.Shutdown` a five-second context; every test there builds `jrpc.Client{Client: http.Client{Timeout: x}}`, so they all share `http.DefaultTransport` and nobody closes idle connections, leaving `Shutdown` waiting on the keep-alive pool until its deadline expires. It surfaced as a different test name each run with the same 5.01s signature. Measured at one failure in ten runs on master and none in ten with `CloseIdleConnections()` before the shutdown.

## Coverage the change would otherwise have lost

Two behaviours were only ever covered as a side effect of the flakiness, so they are now covered on purpose. The one-second token was the only thing exercising the authenticator's `ClaimsUpd` hook on refresh; `TestServerApp_ClaimsUpd` calls the hook directly and checks admin, blocked, email and restricted-name impersonation, along with the two pass-through cases. Lifting the open-route limit in `ConsistentCount` removed the last incidental exercise of `rateLimiter`, which no test asserted on at all; `TestRateLimiter` drives a burst past the allowance and checks both the refusals and that the limit is per client rather than global.

Neither uses a wall clock, and both were confirmed to fail when the behaviour they cover is removed.

## Second commit: go fix under Go 1.27

Go 1.27 extends `go fix` with the modernizers, so it now rewrites patterns the language has since replaced. It is kept separate from the work above. Legacy `sync/atomic` calls on plain integers become the atomic types, reverse index loops become `slices.Backward`, a `Split`-then-index becomes `strings.Cut`, counted loops become range over an int, and `interface{}` becomes `any` in the e2e suite. The example module needed no changes; the e2e module is behind a build tag and only matches with `go fix -tags e2e ./...`.

## Numbers

Sleeps in backend Go code drop from 67 to 23. Of those 23, seventeen sit inside `testing/synctest` bubbles where the clock is fake, two are poll intervals inside condition waits, two sleep until a deadline they computed, and two are in production code and untouched: the 429 backoff in `app/cmd/cleanup.go` and the submit poll in `app/store/image/image.go`. No CI failure implicates either, and changing production concurrency without evidence is risk without payoff.

The suite is also a little quicker, around 30 seconds against 36 on master, measured three times each locally under `-race`.

The 300s per-package test budget now matches across `ci-backend.yml`, `release.yml`, the `race_test` Makefile target and the documented command, and `CLAUDE.md` gains a Backend Test Determinism section so the convention is written down rather than inferred from whichever test a contributor opens first.

## Verification

The full suite is green under `-race`, `-shuffle=on`, `TZ=America/Chicago`, at `GOMAXPROCS` of 2 and 4, and under CPU oversubscription, including with `-covermode=atomic` as CI runs it; ten consecutive runs of the exact CI command under load passed. `golangci-lint` is clean on both the main module and the example, and the example module builds and tests with `-race`.

Everything here is on Go 1.25, which is what CI pins. Worth flagging separately: five tests fail on Go 1.27 (`TestRest_QR`, both `TestBackup_*`, both `TestService_Resize*`), all of them comparing byte-exact compressed output, and they fail identically on an unmodified master. That is a toolchain change rather than anything in this branch, but it will need addressing before the workflow can move to 1.27.


paskal:
Related: #2191 carries the `TestFsStore_Cleanup` fix on its own, since that one has been failing on unrelated branches in the same window. The same change is included here, so whichever merges first the other will need a rebase.

umputun (approved):
approving. One thing to fix, and it is two words.

**two of the four `chooseUnusedPort` copies promise more than the code gives**

`rpc_test.go:22` and `main_test.go:131` say concurrently running package test binaries "never land on the same number". They can. All four copies listen on `:0`, read the assigned port, close the listener and bind later, so nothing holds the number across that gap and another binary can take it. `server_test.go:1071` and `rest_test.go:850` describe the same function as making a collision "very unlikely", which is accurate.

the kernel's range is a real improvement over a random pick from a fixed 10000-port window, so the change is right; only the two comments overclaim. Worth matching them to the other two, because this PR's whole subject is removing reasons a backend test fails once and passes on a rerun, and a comment saying a port collision cannot happen is what would send the next person chasing one somewhere else. Closing the window properly means the server binding `:0` itself and reporting its address, which is a larger change than this.

**the rest**

nothing got weaker. 38 assertion lines out, 56 in, and the one disjunction in the suite became exact: `assert.True(code == Forbidden || code == Unauthorized)` is now `assert.Equal(Forbidden, code)`. Every `EventuallyWithT` closure asserts on the `*assert.CollectT` it is handed rather than on `t`, so none of them can silently never fail.

the new goleak ignore holds up. `fastclock.go:20` in the vendored regexp2 says the goroutine stops on its own once `clockEnd` is reached, so it is a self-terminating worker that outlives a fast binary rather than a leak.

the `sync/atomic` rewrites go the safe way. A field declared `atomic.Int32` cannot be read without `Load`, where the plain `int32` it replaces could be, so the modernizer removes a class of mistake rather than risking one.

Issue #1990, opened 2026-01-13
Render placeholder in remark42 div until comments are loaded?


geerlingguy:
I would like to add a placeholder on my site until the comments are loaded in—right now, it looks like this if the comment server is down or taking a long time to respond:

<img width="545" height="274" alt="Image" src="https://github.com/user-attachments/assets/bc0ba69b-6b35-421e-a6f4-e020fc34c01f" />

I'd like to have a short message inside the `<div id="remark42"></div>` div prior to loading, and was trying to javascript my way towards doing it... but it seems like the comments just load in _after_ the message I put there, instead of replacing the message.

I'd like to have a placeholder in the comments section until the comments load, if possible. Could there be a setting in the remark_config for, maybe, forcibly replacing the contents of the "remark42" div?

PR #2192, opened 2026-08-22
Restore the legacy /web/*.js URLs and fix iframe reuse

umputun:
Two defects found after deploying master to radio-t.com, where every comment counter went blank.

**The .js URLs are gone.** #2178 collapsed the module/nomodule pair into one `.mjs` bundle. Dropping the second compilation was right, since it was shipping `??`, `?.` and class fields to browsers with no module support, so the fallback handed its own audience a syntax error. But it also removed URLs remark42 published itself: v1.16.4's SPA documentation named `/web/embed.js` directly and its loader snippet requested `.js`. Any page that hard-coded one now 404s, with no deprecation.

The break is wider than the pages that hard-code those names. Requesting every name v1.16.4 served under `/web` against a master build returns 404 for 34 of them, and this fixes 29: the six entry bundles (`embed`, `counter`, `remark`, `last-comments`, `deleteme`, `node-emoji`) and the 23 locale chunks. The locale chunks are requested by the bundle runtime rather than held by any page, so they were live 404s in every deployment rather than something an integrator could have noticed. Credit to @paskal for measuring this over HTTP against both images. A documentation-based audit finds only the handful integrators copied.

The remaining five are webpack licence banner sidecars for bundles that no longer exist. They stay unfixed: a `*.js.LICENSE.txt` request is not a `.js` request, nothing in the current output references them, and the one banner this build does emit, `remark.mjs.LICENSE.txt`, resolves. Restoring them would mean emitting files nothing reads.

`webFiles.Open` now retries a missing `.js` against the `.mjs` sibling. The bundles contain no `import` or `export`, so the same bytes serve both names, and Go returns `text/javascript` for both extensions. The retry runs only once both sources report the name missing, so a real `.js` an operator drops into `--web-root` still wins, and an unreadable sibling reports its own error rather than being flattened into a 404. Docs stay module-only: this is a compatibility bridge, not a second supported path. It lives in the Go file server, so it does not cover a deployment serving `/web` from nginx or a CDN directly, or the webpack dev server.

**Embed adopted whatever was in its root.** `createInstance` took `root.firstElementChild` as its iframe. The pirates.radio-t.com theme puts a `<noscript>` inside `#remark42`, so embed adopted that, never called `createIframe`, and posted height messages to an element that cannot show comments. The line dates to `5825a55b`, January 2021.

It also defeats the placeholder support added for #1990. The documented example is a text node (`<div id="remark42">Comments loading...</div>`) and that case works. What fails is the wider promise the same pages make, "any placeholder content" and "any content you place inside the div": an element placeholder is mistaken for the iframe, so `inited` never arrives and the cleanup never runs.

The iframe now carries `data-remark42-iframe` and the lookup is scoped to a direct child, so a second `createInstance` still reuses it, which real integrations rely on, while nothing else in the root can be adopted.

An e2e case loads `/web/counter.js` as a classic script into `privacy.html`, which carries no remark scripts of its own, so the alias cannot pass vacuously and a future real-ESM bundle cannot silently invalidate it.

**A third commit is unrelated to the above.** `TestBackup_MakeBackup` and `TestBackup_Do` pinned the gzip output at 52 bytes. The same input encodes to 57 on go 1.27, so both fail there. They now read the backup back and compare it against what the exporter wrote.

**Left out deliberately.** Reuse still installs a second set of listeners around one iframe, so a double `createInstance` leaks the first instance's handlers. Containing that needs a per-root instance registry. It predates this change, since the old lookup adopted the same iframe and installed the same second set of handlers.

`config.node` is the other one, and it is a dropped feature rather than documentation for something never built. `createInstance` honoured it from #723 (`config.node instanceof HTMLElement ? config.node : document.getElementById(config.node || NODE_ID)`) until `5825a55b` removed the read in the same hunk that introduced the `firstElementChild` line above. The SPA docs have gone on promising it for five years, so a SPA mounting into a differently identified node has been broken that whole time. The resolution is to restore the read, not to delete the docs: that means putting `node` back on `RemarkConfig`, which no longer carries it, along with the HTMLElement and string semantics and tests for both, so it belongs in its own change rather than here.

### DIFF
--- a/backend/app/migrator/backup_test.go
+++ b/backend/app/migrator/backup_test.go
@@ -1,6 +1,7 @@
 package migrator
 
 import (
+	"compress/gzip"
 	"context"
 	"fmt"
 	"io"
@@ -50,9 +51,7 @@ func TestBackup_MakeBackup(t *testing.T) {
 	expFile := fmt.Sprintf("/tmp/remark-backups.test/backup-site1-%s.gz", time.Now().Format("20060102"))
 	assert.Equal(t, expFile, fname)
 
-	fi, err := os.Lstat(expFile)
-	assert.NoError(t, err)
-	assert.Equal(t, int64(52), fi.Size())
+	assert.Equal(t, exportedPayload, gzContent(t, expFile))
 }
 
 func TestBackup_Do(t *testing.T) {
@@ -71,15 +70,31 @@ func TestBackup_Do(t *testing.T) {
 		bk.Do(ctx)
 
 		expFile := fmt.Sprintf("/tmp/remark-backups.test/backup-site1-%s.gz", time.Now().Format("20060102"))
-		fi, err := os.Lstat(expFile)
-		assert.NoError(t, err)
-		assert.Equal(t, int64(52), fi.Size())
+		assert.Equal(t, exportedPayload, gzContent(t, expFile))
 	})
+}
+
+const exportedPayload = "some export blah blah 1234567890"
+
+// the compressed size is not assertable: it moves with the compress/flate version
+func gzContent(t *testing.T, name string) string {
+	t.Helper()
+	fh, err := os.Open(name) //nolint:gosec // path is built by the test
+	require.NoError(t, err)
+	defer func() { assert.NoError(t, fh.Close()) }()
+
+	gz, err := gzip.NewReader(fh)
+	require.NoError(t, err)
+	defer func() { assert.NoError(t, gz.Close()) }()
+
+	b, err := io.ReadAll(gz)
+	require.NoError(t, err)
+	return string(b)
 }
 
 type mockExporter struct{}
 
 func (mock *mockExporter) Export(w io.Writer, _ string) (int, error) {
-	_, err := w.Write([]byte("some export blah blah 1234567890"))
+	_, err := w.Write([]byte(exportedPayload))
 	return 1000, err
 }
--- a/backend/app/rest/api/webfiles.go
+++ b/backend/app/rest/api/webfiles.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"io/fs"
 	"path/filepath"
+	"strings"
 )
 
 // webFiles serves /web from two sources: a name present in the frontend build is served from there,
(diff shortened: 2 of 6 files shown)


paskal (commented):
The fix is sound and I would merge it. I verified the alias, the traversal guards and the precedence claims by running them rather than by reading; below is only what I think should change.

## The `node:` option was implemented, and the commit this PR fixes is what removed it

The description says the SPA docs document a `node:` option `createInstance` has never read. It read `config.node` from the day it existed, in #723 (`git show 5d6729dd:frontend/app/embed.ts`, line 79), until `5825a55b` removed it on 2021-01-06, in the same hunk that introduced the `firstElementChild` lookup this PR is fixing.

So it is a feature silently dropped five years ago while the documentation kept promising it, not documentation for something never built. As written, whoever picks the item up will read it as a docs cleanup and delete the docs, rather than restore one line in a function this PR already edits.

## The break was 34 URLs, not three

I ran `umputun/remark42:v1.16.4` and an image built from master side by side and requested all 90 names v1.16.4 served under `/web`. 34 now 404, and this PR fixes 29: the 6 entry bundles and 23 locale chunks. The remaining five are `*.LICENSE.txt` names that terser emitted by default, nothing references them, and they need no action.

The locale chunks are the part worth noting: no page holds those URLs, so they were 23 live 404s in every deployment and only an HTTP comparison surfaces them. A documentation-based audit finds the three already known.

Which is the argument for pinning the surface in the e2e suite. This list fails on master and passes on this branch:

```go
// the URLs v1.16.4 served under /web. written out rather than derived from the build output:
// a derived list passes by construction and cannot notice a name going away.
entries := []string{"embed", "counter", "remark", "last-comments", "deleteme", "node-emoji"}
locales := []string{"ar", "be", "bg", "bp", "cs", "de", "es", "fa", "fi", "fr", "it", "ja",
	"ko", "mk", "pl", "ro", "ru", "th", "tr", "ua", "vi", "zh", "zh-tw"}
// plus, unsuffixed: iframe.html, counter.html, last-comments.html, deleteme.html,
// privacy.html, markdown-help.html, remark.css, last-comments.css, 400x400.jpeg
```

Each name is requested as both `.js` and `.mjs` and asserted 200 with a non-empty body. `index.html` is left out because it answers 301 and needs its own expectation.

## Nothing pins the output format

`webpack.config.js` sets no `experiments.outputModule` and no `output.module`, so the bundles are non-module because that is the default rather than because anything decided it. If that ever flips, every aliased `.js` URL begins serving a syntax error to precisely the oldest integrations this bridge exists to rescue, and no configuration would object. A comment in the webpack config, or an assertion in the e2e case, would catch it.

## Two corrections to the description

The placeholder sentence inverts which case is documented. The documented example is a text placeholder, at `configuration/frontend/_index.md:74` and `getting-started/installation/index.md:120`, and that case works. What breaks is the wider prose promise, "You can place any placeholder content".

"The same input encodes to 57 bytes on go 1.27, so both fail on any toolchain newer than the one CI pins" is one step past the evidence. CI pins go 1.25 and 1.27 does produce 57, but nothing establishes that 1.26 differs. "on go 1.27" would be exact.

## Two smaller notes

The bridge lives only in the Go file server, so `/web/counter.js` still 404s on the webpack dev server and on any deployment serving `/web` from nginx or a CDN rather than proxying to remark42.

Neither `CLAUDE.md` records it. Both state the `/web` lookup rule as frontend build first, then embedded assets; this adds a third, and those files are where a contributor would look before deleting the retry as dead code.

umputun:
All three factual corrections are in the description, and you were right about the one that mattered.

The `node` paragraph said the docs promise something `createInstance` never read. It now says it was honoured from #723 until `5825a55b` removed the read in the same hunk that introduced the `firstElementChild` line, and that the resolution is restoring it rather than deleting the docs. Your point about what the old wording would cause is why it needed changing, more than the wording itself.

One amendment to your estimate: restoring it is more than one line. `RemarkConfig` in `typings/global.d.ts` no longer carries `node` at all, so it needs the field back along with the HTMLElement and string semantics and tests for both. That is why it stays a separate change rather than riding along here.

The 34 against 29 is now in the description with credit to you. The 23 locale chunks are the part worth having: no page holds those URLs, so they were live 404s in every deployment and nothing short of an HTTP comparison finds them. Confirmed against the live master instance, and separately that all 29 current `.mjs` bundles parse as classic scripts, so the premise the alias rests on is measured.

The licence sidecar exclusion is in as well, with your reasoning: a `*.js.LICENSE.txt` request is not a `.js` request, nothing in the current output references the five, and the one banner this build emits resolves.

The placeholder sentence and the go 1.27 overreach are both corrected.

**On the enumerated e2e case, I am declining it**, and the arithmetic is why. The list is 29 names times two suffixes plus nine static entries, so 67 rather than the 90 v1.16.4 served. It is already a curated selection of what should stay resolvable, which is the judgement call the hand-written form was meant to avoid making. Past that it freezes generated locale chunks as permanent API, goes stale the moment a locale is added, and needs a policy decision to remove one.

The backend cases already pin the suffix behaviour for arbitrary names, and the `counter.js` browser case pins the cross-layer fact that an alias is served over HTTP and executes as a classic script, which is the part neither layer can check alone.

Your point about nothing pinning the output format is fair, and the derived form is the better test: enumerate the emitted `.mjs` bundles and assert each parses as a classic script and its `.js` route serves the same bytes. Only the input set is derived there, so it can still fail. It belongs beside a change that touches the webpack output config, where someone is about to break the premise, rather than in a compatibility repair.

The commit message wording goes into the squash body rather than a force-push, since the squash replaces it and a force-push would drop your review state.

Nothing outstanding from my side. Could you convert the review when you get a chance, it is the only thing blocking the merge.

paskal (approved):

Issue #2139, opened 2026-08-06
Email-auth Sign In dropdown opens and self-closes instantly (~100-150ms)


wpguyuk:
### Describe the bug

The email-auth "Sign In" dropdown in the comment widget opens and self-closes almost instantly (~100-150ms, confirmed via frame-by-frame screen recording analysis), before the user can interact with the username/email fields. It reproduces on every post, via both mouse click and keyboard (Tab + Enter/Space) activation, on a full-width desktop browser window.

### Environment

- Remark42 version: `v1.16.4-6e7820d-20260709T22:34:18` (Docker, `ghcr.io/umputun/remark42:latest`)
- Auth: email-only (`AUTH_EMAIL_ENABLE=true`), no social/anonymous providers
- Embed: standard `<div id="remark42">` + `remark_config` snippet, no customisation of the widget itself
- Browser: Chrome (Fedora Linux desktop), reproduces identically in the affected user's normal profile
- Does **not** reproduce: in an Incognito window (same Chrome, same machine), in our own automated testing (Chrome via browser automation, logged in and logged out), or on remark42.com's own `/demo/` page tested at the same time

### Steps to reproduce

1. Load a page with the Remark42 widget (email-only auth)
2. Click "Sign In" (or Tab to it and press Enter/Space)
3. The username/email dropdown form flashes briefly on screen, then closes on its own — no click occurs outside the dropdown

### What we ruled out on our end

Before concluding this is likely an upstream issue, we checked (with network/console tooling) and ruled out:
- Browser extensions (ad-blockers explicitly disabled, still reproduces)
- Cached JS/HTML (hard-refreshed repeatedly)
- Stale/invalid cookies on the comments subdomain (none present)
- Any site-side script/plugin conflict (diffed the full list of enqueued scripts between the failing session and a clean one; removed the one extra script present only in the failing session — no change)
- Custom CSS touching the widget, iframe, or its container (none exists)
- Duplicate widget embeds on the page (only one `#remark42` mount point)
- Narrow/mobile-width viewport (window was full-width desktop)
- The backend auth flow itself — `GET /auth/email/login` works correctly and delivers the confirmation email when triggered directly (via curl), so the backend side is healthy; the browser Network tab shows **no request at all** fires to the comments domain during the failing click, meaning the click isn't reaching the point where Remark42's own auth-request code would run

### Additional notes

[...]

PR #2196, opened 2026-08-22
Broaden the e2e suite from 21 cases to 63, and harden its harness

paskal:
The e2e suite goes from 21 cases to 63, and the harness gains three changes so a failure names itself. Every addition is tied to something that broke before, or to a contract a rewrite would have to keep.

## What the suite could not tell you about itself

An uncaught exception in the page now fails the test that caused it. A widget that throws while rendering still satisfies most of these assertions, since they are about elements the browser lays out either way, and #1979 and #851 were both exactly that. A rate-limit response fails a test too: a refused `/auth/status` renders as a signed-out reader, and the sign-in assertion then fails somewhere that does not name the cause. Console errors are collected and logged when a test fails but never fail one, since the browser writes one for every failed request.

The stack is checked against the sources under test. Every checkout builds the image tag the compose file names, so a stack from another worktree, or from this one before an edit, answers on these ports and passes every readiness probe while serving code nobody is looking at. `e2e/stamp.sh` digests the content of the paths that go into the image, compose puts it in the container environment, and the suite reads it back. It digests content and not `HEAD`, so a commit touching only the suite does not invalidate a stack.

`assertSignedIn` no longer waits out the whole timeout on a refused status read. `/auth/` is capped at two requests a second for the entire suite, and a case signing in on two pages spends that twice; the short first wait now ends in a focus handoff, which the widget answers by re-probing. A sign-in that genuinely failed still fails.

## Iframe geometry

The widget measures its own document and posts the number for the parent to apply, and every way that has gone wrong is invisible to assertions about elements, which read the same whether the frame is right, twice too tall or collapsed to a strip.

- the first height the parent is given describes rendered content, not the preloader (#2126)
- the frame matches the document it holds and does not stand 24px taller (#1487, #2151)
- `no_footer` leaves the last comment inside the frame instead of clipping it (#2073, #2076)
- the frame follows the sign-in dropdown and the growing textarea, and comes back down (#1495)
- collapsing a thread shrinks it, which the four above would all miss

Each of the first three was verified by reintroducing the defect and watching the case fail: a 63px report ahead of the real one, six pixels of body padding, and a frame sized under the content.

## The embed and config surface

An element placeholder gives way to exactly one iframe carrying the embed's own marker, and a second `createInstance` reuses it instead of adding another (#1990, #2002, #2192). `destroy` takes the iframe away with the handles it documents dropping. A theme change after load reaches both the element and the document. A host page posting a message of its own, which is all `embed.ts`'s title observer does, no longer empties an open login form (#2139, #1761).

From `remark_config`, none of which was covered: `__colors__`, the one setting that travels through `window.name` instead of the query string; the `url` override, which is how a canonical address keeps one conversation across pages that differ; the subscription controls an integrator turns off; `simple_view`; and an unrecognized locale falling back to English, which is `loadLocale`'s only observable guarantee. Plus timestamps rendering in the reader's own timezone, the one part of a comment that cannot move to the server.

## Cross-origin, which nothing reached

Every host page in the suite was served by remark42 itself, so the separate-domain setup in the manuals was never exercised. An nginx on its own name and port now serves one, and the widget it embeds addresses remark42 by `remark_config.host`: the frame has to be revealed, meaning its document loaded and reported itself inited across the boundary, and the thread it renders has to be the one the page's address names. The other half is `ALLOWED_HOSTS`: the instance that names only itself is refused everywhere else, its document never runs, and the reveal comes from the widget's own fallback.

Signing in there is deliberately not asserted. An embedded cookie needs `SameSite=None`, which browsers take only as `Secure`, and this stack speaks http. That is #1139.

## Configurations that cannot share an instance

Four services, each because the setting changes the widget for every reader: an admin's unlimited edit window (#1986), a session carried in a header instead of a cookie (#1877), an instance with no auth provider (#1456), and anonymous voting, which `rest_private.go` turns down unless `ANON_VOTE` is set. The main instance gains the notify module, without which `email_notifications` is false and the subscribe control never renders, making the whole subscription round trip unreachable.

The admin instance takes its admin from an email address, not a name: remark42 hashes an anonymous id from the name and the client address together, so an id written into `ADMIN_SHARED_ID` belongs to nobody on another machine. That is why the first CI run had no admin at all.

## Votes, auth and the rest

Downvoting and its survival of a reload. A vote on your own comment, refused by the widget and by the backend. A vote with the `X-XSRF-TOKEN` header stripped, which has to be refused: that check is why a document navigation is always anonymous and why the widget hydrates its user over XHR, so anything designed around it wants it pinned. A transient `/auth/status` failure that must not end the session (#1763). Sign-out that actually ends it. The edit form handing back the source that was posted (#2040), drafts, a refused comment, image upload and its failure path, pinning and verifying, a blocked author, a read-only thread, a hidden user, every locale against the catalogue on disk, the delete-me page, and the published `/web` surface with its legacy `.js` names.

## Two findings and one defect, none fixed here

#2204: **a page whose URL has two query parameters cannot be commented on.** `Comment.Sanitize` round-trips the locator through bluemonday, so `&` is stored as `&amp;`; the bucket is created under the escaped key, the read-back uses the real one and answers 500, and every later `find`, `counts` and feed asks for the real URL and is told the thread is empty. Reduced to a two-line curl reproduction in the issue.

An admin's own freshly posted comment renders an edit countdown that should not exist under `ADMIN_EDIT`, correcting itself on the next load (#2205). After subscribing by email and reloading, the widget offers to subscribe again, reading state from `user.email_subscription` which is absent from what it hydrates the user with (#2206).

#812 reproduces as reported: the setting reaches the widget, appears in the iframe's query string and changes nothing, four comments rendering with it set to two. No case for it here, since it would be red.

#2208 records why Telegram has no case: the base URL is formatted inline inside `go-pkgz/auth`, so nothing here can point it elsewhere. The fix belongs upstream, in v1 and v2 both. What the audit turned up and this PR did not cover is collected in #2210.

## Note

The suite runs about five minutes, so the workflow's budget goes to 20m to match the Makefile, and the workflow stamps the stack it starts the way the Makefile does.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -88,8 +88,12 @@ jobs:
           key: playwright-${{ hashFiles('e2e/go.sum') }}
           restore-keys: playwright-
 
+      # E2E_STAMP is what the suite compares the running stack against, so a stack started here
+      # has to carry the same value `make e2e-up` and the suite itself would give it
       - name: Build & start the stack
-        run: COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker compose -f compose-e2e-test.yml up -d --build --quiet-pull --wait
+        run: |
+          COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 E2E_STAMP=$(./e2e/stamp.sh) \
+            docker compose -f compose-e2e-test.yml up -d --build --quiet-pull --wait
 
       # no retry: a failure here is evidence about a suite too young to have a flake rate,
       # and a rerun is how an intermittent regression becomes invisible. revisit when there
@@ -99,7 +103,10 @@ jobs:
         # log names the run it came from
         env:
           E2E_RUN_ID: ${{ github.run_id }}-${{ github.run_attempt }}
-        run: cd e2e && go test -tags=e2e -count 1 -timeout 8m -v ./...
+        # 20m, matching the Makefile. the suite runs about four minutes on a laptop and a runner
+        # is slower, so a tighter budget turns a loaded runner into a timeout panic instead of a
+        # readable failure. the job's own timeout above is what bounds a wedged run
+        run: cd e2e && go test -tags=e2e -count 1 -timeout 20m -v ./...
 
       - name: Server logs on failure
         if: failure()
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ rundev:
 	docker compose -f compose-private.yml up
 
 # stamped the same way the suite stamps a stack it starts itself, so one brought up here is
-# accepted rather than rejected as belonging to another checkout
+# accepted instead of rejected as belonging to another checkout
 e2e-up:
 	E2E_STAMP=$$(./e2e/stamp.sh) docker compose -f compose-e2e-test.yml up -d --build --quiet-pull --wait
 
--- a/Makefile
+++ b/Makefile
@@ -39,8 +39,10 @@ rundev:
 		docker compose -f compose-private.yml build
 	docker compose -f compose-private.yml up
 
+# stamped the same way the suite stamps a stack it starts itself, so one brought up here is
+# accepted rather than rejected as belonging to another checkout
 e2e-up:
-	docker compose -f compose-e2e-test.yml up -d --build --quiet-pull --wait
+	E2E_STAMP=$$(./e2e/stamp.sh) docker compose -f compose-e2e-test.yml up -d --build --quiet-pull --wait
 
 e2e-down:
 	docker compose -f compose-e2e-test.yml down -v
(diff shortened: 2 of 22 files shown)


umputun:
merging this. One gap worth a follow-up, not blocking:

`TestVote_DownvoteAndReversal` (`e2e/vote_test.go:180`) never reverses. The godoc says the score has to end where the second vote put it, but the body clicks Vote down once, checks `-1`, reloads and checks `-1` again. The correction from `-1` to `+1` can break with this green.

either add the up-vote after the reload assertion and check the persisted score, or rename to match what it covers.

paskal:
Fixed, and you were right about more than the naming: the case never cast the second vote, so nothing there covered #728 at all.

Writing the correction turned up what the behaviour actually is. The opposite vote **takes the first one back instead of flipping it**, so the score returns to 0 and never reaches +1. My first attempt asserted +1 and failed, which is the assertion the old godoc implied and nobody had ever run.

The case now downvotes to -1, reloads and checks it persisted, casts the opposite vote, checks the score returns to 0, and reloads again to check the server recorded the correction rather than only the screen. Renamed to `TestVote_DownvoteAndCorrection` to match.

umputun (approved):

PR #2198, opened 2026-08-22
Fill the instance URL into the embedded frontend at serve time, and stop pinning compressor output in tests

paskal:
The widget falls back to a compiled-in URL whenever a page omits `remark_config.host`. The bundler cannot know that URL, so it emits `{% REMARK_URL %}` and each distribution substitutes it. The docker image rewrites the files under its web root at container start; the release binary, which serves the build embedded in itself, had nothing doing it. `scripts/prepare-release-assets.sh` filled the marker with `http://127.0.0.1:8080` before the embed instead, so every copy of the binary shipped pointing at the visitor's own loopback, and on an https site that request is blocked as mixed content besides.

Every published binary carries it. Checking the release tarballs, the literal appears from v1.11.0, the first release to embed the frontend, through v1.16.4, and the releases before that embedded no frontend at all, so the binary distribution has never served a correctly addressed widget:

| release | embeds the frontend | fallback in the bundles |
| --- | --- | --- |
| v1.4.0 to v1.10.0 | no | none present |
| v1.11.0 to v1.16.4 | yes | `http://127.0.0.1:8080` |

Docker is unaffected either way: `docker-init.sh` substitutes the real `REMARK_URL` into `/srv/web`, and that disk copy is served ahead of the embedded one.

The placeholder now survives into the embedded copy, and the file server fills it with the configured `REMARK_URL` as it serves, in html, js and mjs, the same set `docker-init.sh` rewrites. The served size is the substituted one, so a response is neither truncated nor left hanging, and range requests still work. The Dockerfile no longer bakes the loopback address into the image's embedded copy either, so the fallback it keeps for a missing web root is correct rather than misleading.

The release script no longer needs perl, since substituting was the only thing it used it for, so `CLAUDE.md` drops it from what a local release build requires.

Checked against a real artifact rather than a `go run`: `goreleaser build --snapshot`, which runs the release hook, then the binary started from an empty directory with `REMARK_URL=https://comments.example.com`.

```
embed.mjs      remark_config.host??"https://comments.example.com"
embed.js       remark_config.host??"https://comments.example.com"
counter.mjs    remark_config.host??"https://comments.example.com"
remark.mjs     remark_config.host??"https://comments.example.com"
```

No placeholder and no loopback address left anywhere in the served output, Content-Length matches the body on every type including the untouched jpeg, and a range request still answers 206.

## Also here: #2200, three tests that assert the compressor's output

Folded in rather than opened as a fourth PR. `TestRest_QR` compared a response to a golden PNG byte for byte, and the two resize cases pinned the exact encoded length, so all three fail on go 1.27 while CI's pinned 1.25 passes and the images themselves are perfectly valid.

The QR case now decodes both sides and compares pixels, which is the same assertion about the QR code and none about the encoder. The resize cases assert the image fits the box `resize` was given and touches one of its sides, which is what fitting to a box means.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,7 +60,7 @@ git push origin backend/vX.Y.Z
 
 GoReleaser must ignore `backend/*` tags in `.goreleaser.yml` so release notes and current-tag detection use only product tags. Docker image publishing stays separate and is handled by the existing Docker workflow.
 
-For local artifact runs, install GoReleaser, Go 1.25, Node 24+, PNPM 10, and Perl, then use `make release`. The target runs a snapshot/no-publish GoReleaser build, leaves local artifacts and metadata in `dist/`, and cleans generated frontend embed files after GoReleaser exits. Do not run raw `goreleaser release` for local artifacts unless you also run `./scripts/cleanup-release-assets.sh` afterward.
+For local artifact runs, install GoReleaser, Go 1.25, Node 24+ and PNPM 10, then use `make release`. The target runs a snapshot/no-publish GoReleaser build, leaves local artifacts and metadata in `dist/`, and cleans generated frontend embed files after GoReleaser exits. Do not run raw `goreleaser release` for local artifacts unless you also run `./scripts/cleanup-release-assets.sh` afterward.
 
 ## Milestones and Issue Labels
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,6 @@ RUN apk --no-cache add gcc libc-dev
 ADD backend /build/backend
 # to embed the frontend files statically into Remark42 binary
 COPY --from=build-frontend /srv/frontend/apps/remark42/public/ /build/backend/app/cmd/web/
-RUN find /build/backend/app/cmd/web/ -regex '.*\.\(html\|js\|mjs\)$' -print -exec sed -i "s|{% REMARK_URL %}|http://127.0.0.1:8080|g" {} \;
 WORKDIR /build/backend
 
 RUN echo go version: `go version`
--- a/backend/app/rest/api/rest.go
+++ b/backend/app/rest/api/rest.go
@@ -394,7 +394,7 @@ func (s *Rest) routes() http.Handler {
 		log.Printf("[WARN] no embedded frontend, serving built-in assets only: %v", err)
 		embeddedFrontend = emptyFS{}
 	}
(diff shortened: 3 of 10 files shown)


umputun:
one thing before merge.

**the ETag does not vary with `REMARK_URL`.** `cacheControl` (`backend/app/rest/api/middleware.go:210`) builds the validator from `version + ":" + r.URL.String()`, and `addFileServer` passes only `version` (`rest.go:534`). The served body now depends on `remarkURL`, so an operator who corrects a wrong `REMARK_URL` and restarts the same binary gets a 304 on revalidation and the client keeps the bundle addressed to the old host. `Cache-Control: no-cache` means it revalidates every time and keeps getting the same wrong 304, so it does not age out after the hour either.

that is the exact case this PR exists for, a config-only fix to a misaddressed widget. Putting `remarkURL` into the key handed to `cacheControl` covers it.

not raising the unescaped substitution into the JS string literal: `REMARK_URL` is operator config, `server.go:596` checks only the scheme, and `docker-init.sh` does the same raw `sed`. A value carrying a quote breaks every substituted bundle, but that is config robustness across both distribution paths rather than something for this PR.

paskal:
Good catch, and it is the case this PR exists for rather than an edge of it.

Fixed: `addFileServer` now hands `version+":"+remarkURL` to `cacheControl`, so the validator moves when the substituted body does. Covered by `TestRest_FileServerEtagVariesWithTheInstanceURL`, which builds the same file server twice with different instance URLs and asserts the etags differ; reverting the key fails it.

On the unescaped substitution into the JS string literal, agreed it is out of scope here and I have not touched it. Worth recording somewhere though, because the two distribution paths now differ in one respect: `docker-init.sh` substitutes once at image start and a broken value shows up immediately, while serve-time substitution re-applies per response, so a `REMARK_URL` carrying a quote would break every bundle on every request with nothing at startup to hint at it. `server.go` validating the value rather than only its scheme would cover both paths.

paskal:
Rebased on master and squashed to two commits: the image assertions, and the instance-url substitution with the etag fix folded in. Tree is byte-identical to before the squash, verified by diffing it.

Your feedback is in: `addFileServer` hands `version+":"+remarkURL` to `cacheControl`, so the validator moves when the substituted body does, covered by `TestRest_FileServerEtagVariesWithTheInstanceURL` which fails if the key is reverted.

Backend tests, lint and build green.

umputun (approved):

PR #2199, opened 2026-08-22
Cover the locale chunks, the delete page and the last-comments stylesheet

paskal:
Three things the e2e suite could lose without going red. Each was found the same way: remove the asset from the running instance, run the whole suite, and watch every test stay green.

**The locale bundles.** Nothing asked for one, so every run was English. `loadLocale` imports the chunk dynamically and falls back to the English defaults on any error, so a missing or broken chunk renders a perfectly good widget and says nothing. The test opens the widget document with `locale=ru` and asserts two strings that only come from the chunk. It has to be the widget document rather than the demo page, which has no way to pass a locale, and on the instance's own origin, since the bundle addresses the host it was substituted with and its CSP is `self`.

**The delete page.** `deleteme.html` and its bundle had no coverage of any kind. It needs an admin token to do its work, so the test drives the branch it takes without one: reaching that message proves the page and its bundle were both served and executed.

**The last-comments stylesheet.** It is appended at runtime and nothing waits for it, so the comments render whether or not it arrives. The test now waits for that response and asserts its status.

That last one is worth spelling out, because the first version of it was useless. It collected failed responses and asserted the list was empty after the comments appeared, which passed with the file deleted: the 404 had not landed yet when the assertion ran. Waiting for the specific response is what makes it fail.

Each test was verified by deleting the asset it covers from the running instance and confirming it goes red.

### DIFF
--- a/e2e/thread_test.go
+++ b/e2e/thread_test.go
@@ -4,6 +4,7 @@ package e2e
 
 import (
 	"fmt"
+	neturl "net/url"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


paskal:
Folded into #2196 as c27e6f44, which is where this coverage belongs: it touches the same two files and would have conflicted on merge, as the cherry-pick did.

The overlap with what #2196 already had is worth stating, since the locale ones look alike. `TestWeb_EveryBundleServesUnderBothSuffixes` asserts every locale chunk is served, counting the catalogues on disk as a floor. `TestThread_LocaleLoadsItsTranslationChunk` asserts one is fetched, parsed and rendered: `loadLocale` swallows a broken chunk and falls back to the English defaults, so a file that serves perfectly well but fails to parse passes the first and fails the second.

PR #2211, opened 2026-08-22
fix(dropdown): decide outside clicks in the capture phase

### LINKED ISSUE #2209: Unsubscribing by email closes the panel before it can confirm
Clicking Unsubscribe in the email subscription panel closes the dropdown, so the reader never sees the message confirming it. The request is made and the server answers, but the panel that would say so is removed at the same moment.

`comment-form__subscribe-by-email.tsx` describes the mechanism in a comment of its own: the click changes the step, the dropdown looks for the event target in the rerendered view, does not find it and treats the click as being outside itself. The comment invites a better solution, so this is a known awkwardness rather than a surprise, but from the reader's side an action with no acknowledgement looks like an action that did not happen.

The same shape affects the subscribed step, which is why the end-to-end case added in #2196 asserts the request and its response instead of the panel.

yzxcj797:
## Problem

Clicking **Unsubscribe** in the email subscription dropdown closes the dropdown at the same moment, so the reader never sees the "You have been unsubscribed" confirmation (#2209). The mechanism is documented in the comment inside `comment-form__subscribe-by-email.tsx`: the click changes the step, and by the time the dropdown's outside-click check runs in the bubble phase, the clicked button has been detached by the rerender, so `contains(target)` reports the click as outside and the dropdown closes itself.

The same shape affects any click inside dropdown content whose handler rerenders and replaces the clicked node.

## Fix

Evaluate outside clicks in the **capture phase** (`addEventListener('click', ..., { capture: true })`), which runs before the click reaches inner controls and before any handler can rerender and detach the target. The verdict — inside or outside — is then unaffected by what the inner handlers do afterwards.

No change to the closing behavior itself: a genuine outside click still closes the dropdown, a click on the title still toggles it.

## Testing

- New `dropdown.test.tsx`:
  - "closes on an outside click" — unchanged behavior;
  - "stays open when an inner click rerenders and detaches the clicked node" — simulates the mid-dispatch detach (a capture listener removes the clicked node during propagation, exactly what a step change does in the browser). Fails on the current code (dropdown closes), passes with this fix.
- New end-to-end case in `comment-form__subscribe-by-email.test.tsx`: open the dropdown, unsubscribe, and assert the confirmation plus the Close button are still visible.

Note: jsdom dispatches all listeners on one JS stack, so the natural browser timing (microtask rerender between listener phases) cannot be reproduced directly; the mid-dispatch removal reproduces the same DOM state the bubble-phase check sees.

Fixes #2209

### DIFF
--- a/e2e/subscribe_test.go
+++ b/e2e/subscribe_test.go
@@ -77,14 +77,25 @@ func TestSubscribe_EmailRoundTrip(t *testing.T) {
 	unsubscribe := frame.Locator(`button:text-is("Unsubscribe")`)
 	waitVisible(t, unsubscribe)
 
-	// the panel confirming an unsubscribe is not observable: the click changes the step, and the
-	// dropdown closes on an element no longer in the rerendered view, which the component notes
-	// as a known awkwardness of its own. So assert what the server was asked and what it
-	// answered, which is what decides whether the reader still gets mail
 	resp, err = page.ExpectResponse("**/api/v1/email**", func() error {
 		return unsubscribe.Click()
 	}, playwright.PageExpectResponseOptions{Timeout: playwright.Float(float64(waitTimeout.Milliseconds()))})
 	require.NoError(t, err, "clicking Unsubscribe asked the server nothing")
 	assert.Equal(t, "DELETE", resp.Request().Method())
 	assert.Equal(t, http.StatusOK, resp.Status(), "the server refused the unsubscribe")
+
+	// the confirmation the panel shows afterwards, which nothing asserted before: the comment that
+	// used to stand here said it was unobservable because the dropdown closed over it. Note what
+	// this does not pin. It passes with the capture listener reverted, because the unsubscribe is
+	// awaited and the step changes a whole task after the click has finished propagating, so the
+	// button is still in the tree when a bubble-phase listener asks. The phase only decides a
+	// detach that lands inside the click, which is what dropdown.test.tsx drives directly.
+	waitVisible(t, frame.Locator(`text=You have been unsubscribed by email to updates`))
+
+	// the panel closes when asked to, which is the half a null render used to fake: the old trick
+	// emptied the content and left the listbox itself open on the page
+	closeButton := frame.Locator(`button:text-is("Close")`)
+	waitVisible(t, closeButton)
+	require.NoError(t, closeButton.Click())
+	waitHidden(t, frame.Locator(`div[role="listbox"]`))
 }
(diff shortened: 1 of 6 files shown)


paskal:
Three things to change, one of them behavioural.

**The Close button stops closing the dropdown.** `SubscribeByEmail` renders `null` for `Step.Close`, and the comments above that branch and above the `Step.Unsubscribed` branch say why: the dropdown closes because the clicked node is gone by the time the document listener runs. In the capture phase that click is recognised as inside, so nothing closes the panel. After Unsubscribe then Close it is still in the DOM and empty, `document.querySelectorAll('div[role="listbox"]').length === 1` with `textContent === ""` (jsdom, this head).

The platform behaviour behind both #2209 and that trick, measured on a real click in Chromium, Firefox and WebKit (playwright 1.62.1), with the clicked node detached synchronously and from a microtask, which is what a preact rerender does:

| detach | capture: `panel.contains(target)` | bubble |
|---|---|---|
| in the handler | true | false |
| in a microtask | true | false |

The capture phase sees the same inside click in a browser as in jsdom, so the empty panel is what a reader gets there too.

Suggestion: keep the capture move, give `SubscribeByEmail` a real way to close the dropdown (a callback `Dropdown` hands its children, or lifting `isActive` into the parent), and delete `Step.Close` together with both comments that document the quirk.

**The test carrying the issue number does not cover the issue.** `<SubscribeByEmail/> dropdown > keeps the dropdown open after unsubscribing so the confirmation is visible (#2209)` passes with `{ capture: true }` reverted in `dropdown.tsx`; of the 13 tests in those two files only the new `dropdown.test.tsx` case fails. jsdom cannot show the difference at all: its dispatch keeps `contains(target)` true in both phases, which is the entire content of the table above, so no jest test can pin #2209. That regression test needs a real browser and belongs in `e2e/`. The synthetic case that does fail detaches the node from a native capture listener it installs itself, so it pins the general shape and not the flow from the issue.

**Lint.** `eslint --max-warnings=0` on the new subscribe test, with the toolchain on current master:

```
207:13  error  'container' is assigned a value but never used   @typescript-eslint/no-unused-vars
207:57  error  Replace `·...initialStore,·user:·{·...user,·email_subscription:·true·}` with a multi-line form   prettier/prettier
```

`dropdown.test.tsx` also appends a button to `document.body` and never removes it. Clicking `document.body` itself is an outside click and leaves nothing behind.

umputun (approved):

Issue #2200, opened 2026-08-22
Byte-exact image assertions fail on go 1.27


paskal:
Three tests assert the exact output of Go's compressor, so they fail on any toolchain whose deflate or PNG encoder emits different bytes. CI pins go 1.25 and passes; go 1.27.0 fails.

Reproduced on `a82dc8d3` with go1.27.0 darwin/arm64, nothing else changed:

| test | assertion | on go 1.27 |
| --- | --- | --- |
| `TestRest_QR` (`app/rest/api/rest_public_test.go:960`) | compares the whole PNG byte for byte | IDAT stream differs |
| `TestService_Resize` (`app/store/image/image_test.go:44`) | `assert.Equal(t, 1135, len(img))` | 1166 |
| `TestService_ResizeJpeg` (`app/store/image/image_test.go:53`) | `assert.InDelta(t, 10913, len(img), 100)` | 11320, 407 outside the delta |

Nothing is wrong with the code: the images are valid, only shorter or longer than they were. The assertions are pinned to a compressor implementation rather than to anything the code promises.

The fix is to assert the property instead: that the response decodes as a PNG of the expected dimensions, and that a resized image has the bounds resize was asked for, with a generous size band if a size check is wanted at all.

This bites anyone developing on a current Go release, and it will bite CI the moment the pinned version moves.

PR #2212, opened 2026-08-22
Release the response before tearing the test server down

umputun:
`rest/api` cannot finish under go 1.27. `TestRest_securityHeaders` and `TestRest_frameAncestors` both hang, the package dies on the timeout, and nothing names a cause.

Both tests start a server, read one response, then call `teardown()` partway through so they can start a second server with different options. The first response body is only closed by a `defer`, which does not run until the test returns. `httptest.Server.Close` waits on connections still in use, so it waits on a body that will not be closed until after it returns.

CI pins go 1.25 and passes, which is why this has been invisible. The fix is to close the body and the client's idle connections before `teardown()` in both, which is what `CLAUDE.md` already says to do under "Close idle connections before shutting a test server down".

Measured on go 1.27.0: `rest/api` goes from a 300s timeout panic to 17s. Reverting either half brings the hang back.

One other test in the package fails on 1.27, `TestRest_QR` at `rest_public_test.go:995`, which compares a golden PNG byte for byte. That is #2200 and #2198 fixes it, so it is deliberately untouched here. With #2198 and this branch together the package is green on 1.27.

Related to #2200

### DIFF
--- a/backend/app/rest/api/middleware_test.go
+++ b/backend/app/rest/api/middleware_test.go
@@ -271,11 +271,14 @@ func TestRest_securityHeaders(t *testing.T) {
 	client := http.Client{}
 	resp, err := client.Get(ts.URL + "/web/index.html")
 	require.NoError(t, err)
-	defer resp.Body.Close()
 	assert.Equal(t, http.StatusOK, resp.StatusCode)
 	assert.Contains(t, resp.Header.Get("Content-Security-Policy"), "img-src *;")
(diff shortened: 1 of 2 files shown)


paskal:
Nothing to change. I swept `backend/app` for the same shape, a deferred `resp.Body.Close()` still pending when `teardown()` runs inside the same test, and found no third occurrence, so the two fixed here are the complete set.

paskal (approved):

Issue #2208, opened 2026-08-22
Telegram cannot be covered end to end: the API base URL is not configurable


paskal:
Telegram authentication, the Telegram subscription flow and `telegram-link` are covered by jest and by nothing else. No browser test of that area can be written while the test stack has no way to answer as Telegram, so that coverage cannot be replaced.

### Why a flag here would not be enough

`server.go` builds `provider.NewTelegramAPI(s.Telegram.Token, ...)`, and the host is not a parameter of it: `go-pkgz/auth/v2` formats `https://api.telegram.org/bot%s/%s` inline in `tgAPI.request`, and the avatar path does the same. Nothing in this repository can point those requests elsewhere.

`NotifyGroup.Telegram.API` already exists, marked deprecated and unused, and warns when set. It is not the hook either, and reviving it needs a decision about what it now means.

### Two ways in, complementary

1. An upstream constructor taking the base URL. This is the shape the library wants anyway and would need to land in `go-pkgz/auth` v1 and v2 both, since a change in one only is how the two drift. The change here then becomes a flag feeding a constructor argument.
2. As an interim, an `http.RoundTripper` on the `*http.Client` this repository already constructs, rewriting requests for `api.telegram.org` to a local host. It needs nothing released upstream and keeps the whole production path, library code included. It should be gated on an environment variable unset outside the test stack, and refuse any target that is not loopback, so a misconfiguration cannot send a live token elsewhere.

Supplying our own `TelegramAPI` implementation, which the interface allows, is not worth doing: a test that brings its own client no longer covers the path that ships.

### Then the fake

A small service answering `getUpdates`, `getMe`, `sendMessage` and `getFile` well enough for the flow: the widget asks for a link, the user "messages" the bot, and the poller picks the update up on its next tick. `apiPollInterval` is five seconds and unexported, so a test has to wait a tick out; it cannot force one.

paskal:
This was implemented in #2228 and released in [v1.17.0](https://github.com/umputun/remark42/releases/tag/v1.17.0). `--telegram.api-url` configures the Bot API base URL for both authentication and notifications, and browser tests cover sign-in and subscriptions through a local Telegram stub.

PR #2214, opened 2026-08-22
Reach what http cannot: the widget over TLS, embedded cross-origin

paskal:
Every service in the suite spoke http, and the browser gates a whole class of behaviour on the page protocol: `Secure` cookies, `SameSite=None`, `Partitioned`, and any code reading `location.protocol`. None of it was executed, which is how `setAuthCookie` came to decorate its cookies with `__Host-` on https pages and survive for years.

A TLS pair joins the stack: remark42 with `SSL_TYPE=static` on 8443, and an nginx serving a host page on its own name on 8444, both on a self-signed certificate that `e2e/tls/generate.sh` makes and `.gitignore` keeps out. Every context accepts it, and so does the readiness client, since those are the only servers either talks to. The instance also runs with `AUTH_SEND_JWT_HEADER`, which is what makes the widget write cookies of its own: without it the client-side writer never runs on any https page here and every assertion about the attributes it chooses is vacuous.

Three cases:

Each of the first and third cases runs both flows the criterion can reach, anonymous and email. The widget's writer keys off the `X-JWT` header rather than off the provider, so the two are expected to behave alike, and expected is not measured; email is also the flow an operator running off-domain actually reaches for. OAuth is out of scope by construction and Telegram is unreachable until #2208.

- **Signing in across origins and then reloading**, which is the one the http cross-origin case cannot make. The widget holds its token in memory for the life of a page, so signing in and posting says nothing about persistence; only the reload asks whether the cookie was delivered, stored under a name the backend reads, and sent back from a third-party frame.
- **The cookies themselves**, read out of the browser store while the widget is embedded elsewhere, since a cookie the browser refused is absent from that list entirely and one it kept but will not send is worse than useless. Every copy of both names has to be `Secure` and `SameSite=None`, at least one has to be partitioned, and none may carry a `__Host-` prefix nothing on either side reads.
- **The same reload under a browser that blocks third-party cookies**, which the widget's own partitioned pair is the only reason to survive.

That last one needs a browser Playwright does not offer: its default `--disable-features` argument carries `ThirdPartyStoragePartitioning`, and it beats both `--test-third-party-cookie-phaseout` and `--block-third-party-cookies` passed through `Args`. A run configured that way keeps every third-party cookie and the case would pass while asserting nothing. `IgnoreDefaultArgs` drops that list and re-supplies it without the one feature, and a control cookie set from inside the frame has to be refused before anything else is read, so a Playwright release that changes the list fails as itself instead of going quietly vacuous.

| | ordinary third-party cookie | `Partitioned` cookie |
|---|---|---|
| Playwright defaults | kept | kept |
| partitioning left enabled | dropped | stored, with its partition key |

All three pass against master. What TLS still cannot reach, the OAuth popup above all, is written down in `e2e/README.md`: the partition key is the top-level site at the moment the cookie is set, and the popup is its own top-level context, so a case asserting OAuth works cross-domain would be asserting something untrue.

#2213 has merged, so this is now a single commit on current master.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -94,6 +94,7 @@ jobs:
       # has to carry the same value `make e2e-up` and the suite itself would give it
       - name: Build & start the stack
         run: |
+          ./e2e/tls/generate.sh
           COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 E2E_STAMP=$(./e2e/stamp.sh) \
             docker compose -f compose-e2e-test.yml up -d --build --quiet-pull --wait
 
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ http-client.env.json
 
 # traces from failed e2e runs
 /e2e/traces/
+
+# self-signed certificate for the e2e https services, made by e2e/tls/generate.sh
+/e2e/tls/*.pem
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ rundev:
 # stamped the same way the suite stamps a stack it starts itself, so one brought up here is
 # accepted instead of rejected as belonging to another checkout
 e2e-up:
+	./e2e/tls/generate.sh
 	E2E_STAMP=$$(./e2e/stamp.sh) docker compose -f compose-e2e-test.yml up -d --build --quiet-pull --wait
 
 e2e-down:
--- a/compose-e2e-test.yml
+++ b/compose-e2e-test.yml
@@ -245,6 +245,71 @@ services:
       timeout: 3s
       retries: 30
 
+  # the same widget over TLS, which is the only way to reach anything the browser gates on the
+  # page protocol: Secure cookies, SameSite=None, Partitioned, and any code reading
+  # location.protocol. The certificate is self-signed and the suite passes IgnoreHTTPSErrors
+  remark42-https:
+    image: ghcr.io/umputun/remark42:dev
+    container_name: "remark42-e2e-https"
+    pull_policy: never
+    depends_on:
+      remark42:
+        condition: service_started
+      mailpit:
+        condition: service_started
+
+    ports:
+      - "127.0.0.1:8443:8443"
+
+    environment:
+      - REMARK_URL=https://remark42-https:8443
+      - SECRET=12345
+      - AUTH_ANON=true
+      - SSL_TYPE=static
+      - SSL_PORT=8443
+      - SSL_CERT=/srv/tls/cert.pem
+      - SSL_KEY=/srv/tls/key.pem
+      # the widget is embedded from another origin here, so its cookies have to survive a
+      # third-party frame; this is the setting that decides whether they do
+      - AUTH_SAME_SITE=none
+      # and the token arrives in a header, so the widget writes the cookies itself through
+      # setAuthCookie. Without this the client-side writer never runs on any https page in the
+      # stack, and every assertion about the attributes it chooses is vacuous
+      - AUTH_SEND_JWT_HEADER=true
+      # email as well as anonymous, so the flow most operators run off-domain is measured here
+      # and not inferred from the anonymous one. the writer keys off the X-JWT header rather than
+      # off the provider, which is exactly the assumption a case signing in by email checks
+      - AUTH_EMAIL_ENABLE=true
+      - AUTH_EMAIL_FROM=remark42@example.com
+      - SMTP_HOST=mailpit
+      - SMTP_PORT=1025
+      - UPDATE_LIMIT=100
+    volumes:
+      - remark42-e2e-https-var:/srv/var
+      - ./e2e/tls:/srv/tls:ro
+    healthcheck:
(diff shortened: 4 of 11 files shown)


umputun:
#2213 is merged, so this rebases to its own commit.

One blocker, in the guard rather than in what it guards.

**The negative control can pass while proving nothing.** `TestHTTPS_SessionSurvivesThirdPartyCookieBlocking` writes `e2econtrol` with `Evaluate`, then reads it back through `page.Context().Cookies()` and passes when the name is absent. Those are two channels with no ordering between them, so absence means either "the browser refused it" or "the write has not committed yet". That is the exact outcome the control exists to rule out, and it would be silent.

The mechanism, pinned rather than general: playwright 1.62.1's `browsers.json` gives chromium `151.0.7922.34`. In that tag Blink's `CookieJar::SetCookie` increments `required_committed_writes_`, calls `SetCookieFromString` and returns with no completion callback. The barrier is the renderer's own `document.cookie` getter, where `IPCNeeded` sees committed lagging required and forces the synchronous read. `Context().Cookies()` does not go through it: playwright issues browser-session `Storage.getCookies` from `crBrowser.ts`, which never touches that frame's `CookieJar`. So `Evaluate` returning proves the setter was queued, not that the refuse-or-store decision is visible to the next CDP read.

This is worth keeping tied to that build rather than stated as a rule about asynchronous cookie stores, since an implementation is free to serialise both calls in the browser process. It is what to re-check on a playwright bump.

The repair is one frame `Evaluate` that writes the ordinary control and a valid `Secure; SameSite=None; Partitioned` sentinel, then reads `document.cookie` and returns it. The getter is the synchronisation point, so both writes are settled by the time the value comes back. Assert the sentinel's exact name is present before the control's exact-name absence is allowed to mean refusal: the sentinel proves the write path and the partitioned path are live, and only then does the missing control mean the browser turned it down. Returning an object rather than a bool keeps the two diagnoses apart, since "the sentinel never arrived" and "the control survived" are different failures. Order of the two writes does not matter, as both precede the getter and its counter covers everything before it. The later `Context().Cookies()` reads stay as they are, since attributes and partition keys are what they are for.

Two that are not blocking.

`playwrightDisabledFeatures` and `withPartitioningEnabled` are 300-character strings differing by one entry, and a playwright change means editing both. Deriving the second from the first with the one feature removed makes them impossible to drift apart. The control catches the case where partitioning ends up unenforced, but only after a run has been spent.

The blocking browser passes its own `--host-resolver-rules` with two names where `TestMain` maps eight. Correct for every current https page, and its failure would be a loud navigation error rather than a quiet green, so this is duplication rather than a defect.

One thing that is wrong rather than improvable: `generate.sh` says the certificate is "usable from a browser or curl without disabling verification". Correct SANs remove the name mismatch, not the unknown issuer, so curl still needs `-k` and a browser still needs an exception. The following sentence about regeneration evicting what the browser stored also does not apply, since every context here is fresh and passes `IgnoreHTTPSErrors`. Generating once is still right; those are not the reasons.

The rest holds. `IgnoreHttpsErrors` in `newPageInContext` reaches the separately launched blocking browser through `newPageOn`, the probe client's `InsecureSkipVerify` is scoped to a client that only ever talks to this stack, no key material is committed, and the resolver rules gained exactly the names the new services use.

paskal:
All four addressed, pushed as `ee27625e`. The branch was already level with master, so it is still one commit.

**The blocker.** The control now writes both cookies and reads them back through `document.cookie` inside one frame evaluate, and returns an object so the two failures stay apart. The sentinel is asserted first: a valid `Secure; SameSite=None; Partitioned` cookie has to be present before the ordinary control's absence is allowed to mean refusal. Your mechanism is recorded in the helper's comment, pinned to chromium `151.0.7922.34` as you framed it rather than as a rule about cookie stores, with a note that it is what to re-check on a playwright bump.

Mutation-checked in three directions, since a guard that cannot fail is the defect it was written to prevent:

| mutation | result |
|---|---|
| derivation left `ThirdPartyStoragePartitioning` in place, as playwright's defaults do | control survives: fails in 0.43s naming the argument list, `cookies in the frame: [e2econtrol e2esentinel]` |
| sentinel written as `SameSite=None` without `Secure`, so the browser refuses it | fails in 0.17s with "nothing was written at all", `cookies in the frame: []` |
| `partitioningFeature` renamed to something absent from the list | fails before launching a browser: "no longer in playwright's default argument list, so removing it enables nothing" |

**The two non-blocking ones.** `withPartitioningEnabled` is now derived from `playwrightDisabledFeatures` with the one entry named once and removed, and it requires the removal to have changed the string, so a renamed feature upstream fails with that message instead of quietly producing a browser that blocks nothing. The blocking browser's resolver rules are left as they are, for the reason you gave.

**The certificate comment.** Corrected: the SANs remove the name mismatch and not the unknown issuer, so curl still needs `-k` and a browser still needs an exception. The regeneration sentence was wrong for the reason you say and now gives the real one, that the running containers read the certificate once at startup.

The README's trap section carries the ordering point as well, since the next person to write a blocking case will reach for `Context().Cookies()`.

umputun (approved):
lgtm

Issue #2203, opened 2026-08-22
Provider icons 404 on a path-prefixed installation


paskal:
`webpack.config.js:15` sets `const PUBLIC_PATH = '/web/'`, an origin-absolute path, so every asset the bundle requests at runtime is fetched from the root of whatever host serves the page.

That is correct only when remark42 owns the domain root. `manuals/subdomain` documents the other arrangement, `REMARK_URL=https://example.com/remark42`, and there the bundle asks for `https://example.com/web/google.svg` when the file lives at `https://example.com/remark42/web/google.svg`.

`remark.mjs` carries 15 such strings, one per auth provider icon, and `last-comments.mjs` has `/web/patreon.svg`. Verified by grepping the served bundles of both a v1.16.4 release binary and a build of current master: identical in both, so this is long-standing rather than new.

The value dates from `5825a55b` ("Update frontend", 2021-01-06), first released in **v1.7.0**, the same rewrite that dropped the `comments.html` build entry.

Locale chunks are unaffected: webpack rewrites its public path for them at runtime.

`output.publicPath: 'auto'` derives the base from the URL the bundle itself was loaded from, which is right for both arrangements and needs no configuration.

paskal:
Fixed in #2197, merged as 23be25d8. The webpack `output.publicPath` is now `'auto'`, so a bundle resolves its assets against its own script URL instead of a build-time `/web/` prefix, and a path-prefixed instance serves its provider icons.

Two corrections to what I wrote here earlier, having gone back and checked the built output.

The blast radius is three bundles, not the two this issue names: `remark.mjs`, `last-comments.mjs` and `deleteme.mjs` each carry a runtime public path, while the other 26 emitted bundles carry none.

And `'auto'` is not the whole mechanism, because all three overwrite `__webpack_public_path__` at init. `remark.tsx` and `last-comments.tsx` set it from `BASE_URL`, which comes from `remark_config.host` and carries the operator's prefix, so they land on the right answer by a different route than the one I described.

`deleteme.ts` sets it to `${window.location.origin}/web/`, which is origin-anchored and would drop a prefix. I said earlier that this was the defect surviving; it is not, and I was wrong to reopen on it. That bundle references no asset and loads no chunk, so the value is assigned twice and never read, and nothing 404s. It is a footgun for whoever first adds an image to that page, not a live fault.

Closing again. The reported defect is fixed, and the `deleteme.ts` line is hygiene I will handle separately rather than leave this issue open on something that does not reproduce.

PR #2215, opened 2026-08-22
Judge the iframe reveal by the mark, and pin the public path

paskal:
Two independent corrections to the e2e suite, both in cases that could pass or fail on machine speed instead of on behaviour.

### The iframe reveal

Three cases separate a reveal that came from the `inited` message from one that came from the five second fallback, and two of them judged it against a clock this process holds. That is a bet on how fast an engine is, and both bets lose on a loaded machine:

- `TestIframe_StaysHiddenUntilTheDocumentReportsInited` polled the DOM after a loop bounded 500ms below the fallback. An evaluate round trip outlasts that margin, so a fallback firing exactly on time reads as an early reveal. It failed on Firefox here on every run.
- `TestIframe_IsRevealedByTheInitedMessage` capped the message path at 3s. WebKit has reported `inited` 2.6s after creation on this machine, and over 3s under load. It failed there.

All three now judge the recorded reveal against one cutoff, half a second under the fallback. A timer cannot fire early, so below the cutoff the reveal can only be the message and above it only the fallback, and the mark carries the moment itself, so a slow read cannot move it. The hidden case drops its polled visibility assertion for the same mark, which is what the observer records in the first place. `messageRevealBudget` is gone and `revealCutoff` replaces it.

Three runs of the three cases across Chromium, Firefox and WebKit pass.

### The public path

`TestWeb_NoBundleHardcodesTheWebRoot` pins the fix for #2203, which shipped in #2197 with no test at any level: `publicPath` appears only in `webpack.config.js`, and neither suite touched asset paths. An instance mounted under a prefix, which `manuals/subdomain` documents, asked the domain root for its provider icons and got nothing.

The check is on the assignment webpack emits for its runtime public path, a string literal when the path is baked in and an expression when it is derived. Asset filenames are bare in both builds, so the obvious assertion, looking for a rooted `/web/name.svg` string, finds nothing either way and proves nothing.

Verified against a local build with `publicPath` set back to `PUBLIC_PATH`: `remark.mjs`, `last-comments.mjs` and `deleteme.mjs` each carry `.p="/web/"` there and the case fails on all three, while the current build emits ``.p=`${G}/web/` `` and passes. Every emitted bundle is checked instead of one entry, since the original defect put fifteen icons in `remark.mjs` and one in `last-comments.mjs`.

### DIFF
--- a/e2e/iframe_test.go
+++ b/e2e/iframe_test.go
@@ -100,9 +100,12 @@ const (
 	// be satisfied by the fallback alone and says nothing about the message path. bound the
 	// message-path assertions well under it
 	revealTimeout = 5 * time.Second
-	// generous enough for a cold navigation on a loaded runner, and still well under the
-	// fallback, which is the point of the assertion
-	messageRevealBudget = 3 * time.Second
+	// the line the three reveal cases are judged against, measured from the element's creation
+	// and not from anything this process can time. a timer cannot fire early, so below this the
+	// reveal can only have come from the message, and above it only from the fallback. a budget
+	// chosen for how long an engine takes instead would be a bet on the slowest one: webkit has
+	// reported inited 2.6s after creation on a loaded machine
+	revealCutoff = revealTimeout - 500*time.Millisecond
 )
 
 // openWithBlockedIframeDoc loads the demo page with the widget document aborted, so the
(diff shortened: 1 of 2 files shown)


umputun:
2214, 2219 and 2220 are merged, so this needs a rebase. The reveal commit is clear on both readings: judging the mark rather than what is visible when the read arrives is a better oracle and not a weaker one, since the mark records when the reveal happened, so a reveal followed by a re-hide is still caught and no evaluate round trip can move it.

One finding, on the second commit, and it is the shape you have just fixed in 2214.

**`TestWeb_NoBundleHardcodesTheWebRoot` is a negative assertion with nothing proving it can fire.** It requires the body non-empty and then asserts a regex for a baked-in path is absent. `require.NotEmpty` proves bytes arrived; it says nothing about whether the pattern still recognises a public-path assignment. The regex is shaped to minified webpack output, so a webpack or terser change to the property, the quoting, or the emitted helper makes it stop matching, and the case then passes forever while the defect it pins can walk back in.

Reverting `publicPath` and watching all three fail is the right evidence that it works today. It is no evidence that it keeps working, and that is the half a test has to carry itself.

The repair is the one you built for the cookie control: assert a live assignment is there before its absence in the baked form is allowed to mean anything. Three things about the shape, because the obvious version of this is wrong.

Restrict the positive half to `remark.mjs` and `last-comments.mjs`, the two the test already requires in the listing. Not every name `emittedBundles` returns: the locale chunks carry no assignment at all and never should. From the bytes the pre-2203 build still serves:

```
remark.mjs         .p="/web/"   .p=`${X}/web/`
last-comments.mjs  .p="/web/"   .p=`${k}/web/`
ru.mjs             (none)
ar.mjs             (none)
```

Have it share the `\.p\s*=\s*` prefix with the negative matcher and require only that the right-hand side is not the baked literal. Do not pin the template-literal form: after 2219 `deleteme.mjs` has no app-level override left, so webpack's `auto` derivation is its only assignment and need not contain `/web/` at all. A matcher written against the template shape is wrong for that bundle as of now.

The negative scan can stay over every bundle. Only the positive witness needs narrowing.

Worth knowing what the witness does and does not prove: the two assignments visible above are webpack's own from the config and then the app-level `__webpack_public_path__` override from `remark.tsx`, and the second is present in the fixed and unfixed builds alike. So the witness proves the matcher still has a live assignment shape to inspect. It is not a second check on the fix, and nobody reading it later should take it for one.

umputun:
not merging this yet. The two things I asked for on the 23rd are both still open, and there is a third to sort out first.

**1. the witness is still missing.** `TestWeb_NoBundleHardcodesTheWebRoot` still asserts only that a baked-in path is absent, so nothing proves the matcher can fire. My comment above has the shape.

Worth having, because it sharpens that scoping. On a build of current master the assignment is missing from more bundles than I expected:

```
remark.mjs         .p=`${Y}/web/`
last-comments.mjs  .p=`${M}/web/`
deleteme.mjs       (none)
counter.mjs        (none)
every locale       (none)
```

`deleteme.mjs` and `counter.mjs` now sit with the locale chunks. Requiring a live assignment across every name `emittedBundles` returns fails on a correct build, so restricting the witness to `remark.mjs` and `last-comments.mjs` is what makes it work at all rather than what keeps it tidy.

**2. the rebase is still outstanding**, and it matters more now: the only CI run here is against base `3286f028`, three minutes before #2224 merged, so nothing has tested this against a master carrying the asset fix. A rerun will not move that, since it reuses the original `GITHUB_SHA` and retests the old merge ref. It needs a push after the rebase.

**3. a third commit**, `343d6c28`, taking the iterator form of `strings.Split` in the bundle listing. Not in the description and belongs to neither purpose here. Drop it or split it out.

the reveal commit is settled and I have nothing further on it. Witness in, rebase, green CI on the new head, and this goes in.

umputun (approved):

PR #2217, opened 2026-08-23
Scope the frontend plan to the static build

paskal:
Draft. The plan document tried to cost two directions and carry an authentication requirement at the same time, and did none of the three well. This scopes it to one subject: shipping the widget as static HTML, CSS and JavaScript with no npm in the build.

**Taken out of the document.** The server-rendered direction of #825 is a separate proposal and is not costed here; removing it does not decide against it. Multi-domain hosting is an authentication question that touches this work at exactly one point, the fallback page built from `comments.ejs`, and is now tracked outside the document so that a backend decision cannot block a build decision.

**Corrected against master.** The document claimed that dropping npm drops the entire unit suite, so the only route was closing every e2e gap first. #2226 makes that false: 236 of 470 cases now run under `node --test` with nothing installed, so the open question is 234 DOM-rendering cases rather than the whole suite. The same change settles by precedent whether dependency-free node scripts may stay, which the document had listed as open.

Figures were recounted at the current head: 167 TypeScript sources, 29 CSS modules over 2,219 lines, 61 devDependencies of which eight exist only to run jest.

**What the document now says is unresolved**, narrowed to one hard item. `type-check` is the only npm gate with no route: 167 files are typed today, and there is no non-npm checker and no e2e substitute that keeps that guarantee. `size-limit` has a designed answer, `translations-check` is npm-bound only in its extract step, and `lint` can be an accepted loss.

One item in it is cheap and worth doing separately: `documentedWebPaths` in `e2e/webfiles_test.go` still has no entry for `/web/comments.html`. That page returned 404 for four years in exactly the configuration that needs it, and nothing pins it today.

### DIFF
--- a/docs/plans/2026-08-19-frontend-simplification.md
+++ b/docs/plans/2026-08-19-frontend-simplification.md
@@ -1,561 +1,172 @@
-# Frontend direction: two viable paths, and what has to be true for either
+# Frontend simplification: shipping the widget as static HTML, CSS and JavaScript
 
-Written 2026-08-19, revised 2026-08-22. Every file reference below was re-checked against master
-`a82dc8d3` plus #2196, #2197 and #2198, which are treated here as landed: they change the e2e net,
-the manifest layout, the fallback page, the asset path and the instance URL, and costing either
-direction against the state before them would be costing a world that no longer exists.
+The goal is a widget that ships without npm: no installed dependency tree, no bundler, no node in
+the build, and static HTML, CSS and JavaScript as the output. This document plans that and nothing
+else.
 
-## Overview
+Two subjects that earlier revisions carried have been taken out. The server-rendered direction of
+#825 is a different proposal and is not costed here; removing it from this document does not decide
+against it. Multi-domain hosting is an authentication question that constrains this work at exactly
+one point, noted in its place below, and is tracked separately so that a backend decision does not
+block a build decision.
 
-- The widget's measured cost was the **toolchain and the React compatibility alias**, not Preact
-  itself. The alias is now gone, so what remains is the toolchain
-- Two paths are genuinely viable: **incremental simplification** (keep Preact, finish collapsing the
-  toolchain) or the **server-rendered rewrite** of #825. This document sets out to cost both rather
-  than to recommend against one, and it half succeeds: Path A is costed, Path B is not, because two
-  of its inputs are still open. What language the retained client code is written in, and whether a
-  scoped XSRF exemption can make the document render authenticated, both change the figure
-  materially. The "months" estimate below is therefore a placeholder rather than a finding, and
-  should not be quoted as though the comparison had been made
-- Server rendering is not architecturally blocked. Authenticated *fragments* work; the initial
-  *document* render is anonymous in every configuration, and that is forced rather than chosen
-- Whichever path is taken, two standing requirements constrain it, and three pieces of work are
-  worth doing first regardless. One of those three is now done
+## Where this stands
 
-## Standing requirements
+The direction is settled. What was open was whether the verification could survive the move, and
+that is now half answered.
 
-Not tasks. Tests any proposal has to pass.
+The test suite has split in two. Pure logic, meaning anything that is a function over strings,
+objects and dates, runs under `node --test` directly from source with nothing installed, and its CI
+job installs nothing so that a test reaching for a dependency fails. What remains on jest is the
+half that renders components into jsdom.
 
-### R1: Remark42 must be hostable on a domain other than the site embedding it
+That retires the earlier framing, which held that dropping npm drops the whole unit suite and that
+the only route was to close every gap in the e2e suite first. The question is now specifically about
+DOM-rendering cases, and each has three possible answers: move the subject's logic out and test it
+under node, cover the behaviour in the e2e suite, or accept the loss.
 
-This is a product requirement rather than a constraint to respect in passing, and it is the one
-users report against most. The target arrangement is remark42 serving from its own name, say
-`remark42.example.com`, with comments appearing on a different site entirely, say `food.com`. It is
-documented in `site/content/docs/manuals/separate-domain/index.md`, so operators follow it and then
-find that authentication behaves differently from the same-domain case.
+It also settles, by precedent, a question earlier revisions listed as open. Dependency-free node
+scripts may stay, so the translation tooling written against node built-ins does not have to be
+ported to Go to satisfy the goal.
 
-**Acceptance criteria.** A reader on `food.com` can sign in with any configured provider, post,
-reload the page, and still be signed in, on a browser that blocks unpartitioned third-party cookies.
-Nothing short of the reload proves it: the widget holds a token in memory for the life of a page, so
-a test that signs in and posts without reloading passes while the persistence is entirely broken.
+### The extraction pattern
 
-**Where that stands.** The e2e suite covers the rendering half through
-`TestCrossOrigin_WidgetRendersOnAnotherOrigin`, which proves the document loads on another origin
-and reports itself through postMessage across the boundary, and `ALLOWED_HOSTS` refusal through
-`TestCrossOrigin_DisallowedHostNeverReportsInited`. The authentication half is not covered and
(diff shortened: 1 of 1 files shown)


umputun:
The R1 correction is right and the mechanism paragraph is a real fix: the token does travel as a third-party cookie, and `Partitioned` is what carries it rather than the write happening in the frame. `activeJwtToken` being module-level and `getCookie` appearing once for `XSRF-TOKEN` are both as described.

One correction, on the Storage Access route at lines 44-49.

**Its cost is understated, and as the flow stands today the route is not available at all.** The section prices it as a permission prompt plus a revocable grant. The prerequisite it leaves out is that the browser denies the request outright when the embedded origin has no recent first-party interaction. MDN puts it as "access requests are automatically denied if the browser detects that the user hasn't interacted with the embedded content in a first-party context recently".

Remark42 never gets that interaction. The reader interacts on the provider's origin, and the callback returns to the remark42 origin at `iframe.ejs:10-12`, whose first statement is `window.close()`. Nothing happens on that origin that a browser would count, so the later `requestStorageAccess()` from the frame has no interaction history to be granted against.

So the real cost is not a prompt on top of an existing flow. Either the callback stops self-closing and obtains a click first, or something else establishes first-party interaction on the remark42 origin before the frame ever asks. That is a change to the first-party experience, which is a different order of cost from what the paragraph currently implies, and it belongs in the comparison against the other two routes.

One precision in the same sentence: it says the grant would let the frame "read the cookie the callback already set". The grant permits the frame to send requests carrying that cookie; it does not make it script-readable, and the cookie is `HttpOnly`, which the OAuth paragraph above already says. Worth saying it the same way in both places.

Everything else in the section checks out.

umputun:
Source for the part of that finding you would most reasonably push back on. The remark42 origin does become top-level during the callback redirect, so the fair question is whether that navigation itself counts as the first-party interaction.

WebKit's 2021 [Updates to the Storage Access API](https://webkit.org/blog/11545/updates-to-the-storage-access-api/#how-to-1-meet-and-greet-the-user-as-first-party), How-To #1 step 3: "Have the user interact (tap, click, or use the keyboard) with your website as first party", and then "Navigating to and from your website in a redirect without user interaction does not count." The requirement it states is interaction as first party within the last 30 days of browser use.

That is WebKit's own guidance rather than a cross-browser rule, so it prices the route on Safari specifically, which is the browser the section is written for.

paskal:
Fixed and rebased onto master.

You are right that the route is not reachable as the flow stands, and that is a bigger correction than the wording. I had priced it as a permission prompt bolted onto the existing flow, when the prerequisite is first-party interaction the instance origin never acquires: the reader interacts on the provider's origin, and the callback lands on a document whose first statement is `window.close()` under `?selfClose`. So the cost is changing the first-party experience, either by having the callback collect a click before closing or by establishing interaction some other way, and the section now says that and compares it against the other two routes on those terms.

The precision is taken too. The grant lets the frame's requests carry the cookie; it does not make it script-readable, and the JWT cookie is `HttpOnly` regardless. It now says that in the same terms as the OAuth paragraph above rather than in weaker ones.

Worth recording what this cost me twice over. The first version of the section claimed no frontend change could bring OAuth inside the criterion, which was too strong. Correcting it, I reached for the Storage Access API and overcorrected into presenting it as nearly free. Both errors came from pricing a mechanism I had not traced to the point where it touches this codebase, which is the same failure in opposite directions.

umputun:
The Storage Access repair is right and it is a bigger correction than the wording, as you say. Four things left, and the first one is not visible from inside this PR.

**The SAA section and #2218 will contradict each other the moment both merge.** Chrome's current Storage Access documentation says a cross-site cookie reached through SAA must carry `SameSite=None; Secure` ([privacy sandbox guide](https://privacysandbox.google.com/cookies/storage-access-api#samesite-and-secure), a product page rather than a platform contract, so worth re-checking rather than treating as settled). The OAuth callback sets its cookie top-level, so under #2218's new recommendation, which drops `AUTH_SAME_SITE=none`, that cookie carries the default and is treated as `Lax`. Chrome will not expose it cross-site through SAA even after the instance has first-party interaction and the grant is given.

Lines 47-49 dismiss `SameSite` on the basis that the manual supplied `AUTH_SAME_SITE=none`, and lines 95-97 describe adding the header flag while leaving that recipe standing. #2218 removes it. So the route's real price is browser-dependent: Safari can use the unpartitioned cookie without that restriction, Chrome additionally needs `AUTH_SAME_SITE=none` kept, with exactly the unpartitioned-cookie exposure #2218 documents and recommends against. Either say that, or scope the route to Safari.

**Lines 134-136 keep the claim the rest of the PR corrects.** They say `AUTH_SEND_JWT_HEADER` returns the token in a response header "so the client can present it without relying on an ambient cookie". Lines 86-92 say the opposite and are right: after a reload `activeJwtToken` is gone, the first request sends no header, and the token arrives as an ambient cookie that survives only because it is partitioned. Both sentences are in one document. Qualify the older one to the life of a single page, or rewrite it around seeding the partitioned pair.

Two stale facts, from reading the whole file rather than the diff:

- line 108 calls #2214 open. It merged as `2640aaee`, which is in this branch's base.
- line 421 says #2196 took the suite to 48. Its own title and inventory say 21 cases to 63.

Neither changes the design, but removing claims this document no longer supports is what the PR is for.

paskal:
Ran the whole document against current master with codex, since the question was whether it still makes sense rather than whether this diff is right. It found two places where this branch had overstepped, both now fixed here, plus enough staleness that the plan was starting to mislead.

**Two things I had presented as corrections are decisions, and they are yours.** The criteria narrowed R1 from "any configured provider" to three flows. That is product scope, not fact, and it sits badly beside `fixing #1139 must not get harder` in the Path B constraints, which is the same subject. Either the requirement excludes OAuth and #1139 is a separate goal with its own timeline, or the requirement keeps OAuth and is not met today. Separately, the criterion is satisfied for anonymous and email by measurement and for Telegram by inference, which sits badly with a standing requirement being a test a proposal passes. Both are now written as open decisions with the alternatives stated, instead of being settled quietly in a diff.

**Two costs were wrong, one in each direction.** The R1 line said a server-rendered fragment has no client-side writer and therefore needs the upstream `Partitioned` work. That contradicts Path B's own task list, which retains the existing auth through `htmx:configRequest`; `request` in `fetcher.ts` already turns `X-JWT` into the partitioned pair. So the upstream work is mandatory only for a server-set, `HttpOnly`, default-on solution, which is a different choice from the one Path B makes, and Path B is cheaper than the document claimed.

The other way: the XSRF caution leaned on `GET /deleteme` as though a blanket GET exemption handed it to an attacker. It sits under `radmin` behind `Auth`, `AdminOnly` and `matchSiteID`, and `deleteMeRequestCtrl` then requires a separately signed token carrying `delete_me`. Exempting GET removes one gate of three. The caution stands, but it now asks for an audit of every authenticated GET instead of resting on an example that does not carry it.

**Task 2's backlog was sending someone to write tests that already exist.** The cross-origin host page, the `comments.html` fallback and its injection case, and the whole `remark_config` surface, timezone and unknown-locale included, are covered. What is genuinely left is hash deep links, `max_shown_comments`, Telegram, the storage-denied trigger that needs WebKit, and a path-prefixed deployment which #2219 makes concrete.

Also refreshed the figures against `7de51ad2` and added two items master made necessary: the published locale list names 17 languages against 24 shipped, and Path B has to keep the explicit height report #2213 added on panel close, since neither ResizeObserver sees an absolutely positioned panel disappear.

**On the decision itself, nothing has changed.** Path A is still the direction, and most of it has landed. Path B got cheaper in two places, non-OAuth cross-domain persistence and the asset-serving work, and its large costs are untouched: the retained client, ~2,400 lines of stateful auth and composer, authenticated rerendering, cache integration, i18n, and replacement coverage for the unit suite. The document still cannot cost Path B well enough to select it, which is itself worth knowing.

One thing I did not act on: codex reported Path A claiming 59 devDependencies against the verified section's 61. There is no such line; the only figure in the document is 61 and it matches `package.json`.

paskal:
Should be good now, cut everything off except for css and static html + css + js.

umputun (approved):

PR #2218, opened 2026-08-23
Document what actually keeps a cross-domain reader signed in

paskal:
Two documentation defects about the same mechanism, found while checking the claims in the frontend plan against the code.

**The separate-domain manual leaves out the setting that makes it work.** It tells operators to set `ALLOWED_HOSTS` and `AUTH_SAME_SITE=none`, and states that authorisation then works on every listed domain. That was true when browsers still permitted third-party cookies. It is not true on Safari, and increasingly not on Chrome: the auth cookies `Service.Set` writes carry no `Partitioned` attribute, so a browser enforcing the block drops them whatever their `SameSite` value is, and the reader is signed out by the next page load.

What survives the block is `AUTH_SEND_JWT_HEADER`. The token comes back in an `X-JWT` response header, `fetcher.ts` writes the cookie itself from inside the embedded frame, and `authCookieOptions` marks it `SameSite=None; Secure; Partitioned`, which is what the browser keeps for that embedding site and sends back after a reload. The manual never mentioned the flag, so the arrangement worked and nobody was told how. It now names it, points at its security considerations, and says plainly that this rescues Email, Telegram and anonymous but not oAuth, whose callback completes in a popup that is a top-level page of its own.

**The parameter page's mitigation list was left wrong by #2197.** It still promises `SameSite=Strict` cookies and a `__Host-` prefix on HTTPS. `authCookieOptions` dropped the prefix outright, and uses `SameSite=None; Secure; Partitioned` whenever the widget is embedded on another domain, which is the only case the flag is for. So the page described protections the code no longer applies, on the page an operator reads specifically to weigh the risk. The list now describes both branches, including what `Partitioned` buys.

Almost entirely documentation. The one exception is `backend/app/cmd/server.go`, where the `--auth.send-jwt-header` description said the header is sent instead of the server-set cookie; `Service.Set` writes the header and then falls through to set both cookies, so the flag's own help misdescribed what it changes about an operator's exposure. A description string rather than behaviour, but it is a Go file.

### DIFF
--- a/backend/app/cmd/server.go
+++ b/backend/app/cmd/server.go
@@ -102,7 +102,7 @@ type ServerCommand struct {
 			Cookie time.Duration `long:"cookie" env:"COOKIE" default:"200h" description:"auth cookie TTL"`
 		} `group:"ttl" namespace:"ttl" env-namespace:"TTL"`
 
-		SendJWTHeader bool   `long:"send-jwt-header" env:"SEND_JWT_HEADER" description:"send JWT as a header instead of server-set cookie; with this enabled, frontend stores the JWT in a client-side cookie (note: increases vulnerability to XSS attacks)"`
+		SendJWTHeader bool   `long:"send-jwt-header" env:"SEND_JWT_HEADER" description:"also send JWT as a header, so the frontend can store it in a client-side cookie that survives third-party cookie blocking; server-set cookies are still sent (note: increases vulnerability to XSS attacks)"`
 		SameSite      string `long:"same-site" env:"SAME_SITE" description:"set same site policy for cookies" choice:"default" choice:"none" choice:"lax" choice:"strict" default:"default"` // nolint
 
 		Apple     AppleGroup         `group:"apple" namespace:"apple" env-namespace:"APPLE" description:"Apple OAuth"`
--- a/site/content/docs/configuration/parameters/index.md
+++ b/site/content/docs/configuration/parameters/index.md
@@ -197,8 +197,10 @@ When `auth.send-jwt-header=true` is enabled:
   - `SameSite=Strict` when the widget and the page share an origin, which is what prevents the
     cookie being sent from another site
   - `SameSite=None; Secure; Partitioned` when the widget is embedded on another domain, where
-    `Strict` would never be sent at all. `Partitioned` confines the cookie to that embedding site,
-    so it is not readable from any other page the browser visits
+    `Strict` would never be sent at all. `Partitioned` keys the cookie to the embedding top-level
+    site, so a different site gets a separate cookie and cannot reach this one. Pages and
+    subdomains under that same site do share it, since the partition key is the site rather than
+    the page
   - Secure flag automatically added on HTTPS connections
   - Double Submit Cookie pattern with XSRF token matching the JWT ID
 
--- a/site/content/docs/configuration/parameters/index.md
+++ b/site/content/docs/configuration/parameters/index.md
@@ -194,9 +194,12 @@ When `auth.send-jwt-header=true` is enabled:
 - **Security Impact**: JWT tokens are stored in client-accessible cookies that can be accessed by JavaScript
 - **Vulnerability**: This increases vulnerability to XSS attacks compared to server-set HttpOnly cookies
 - **Implementation Mitigations**:
-  - SameSite=Strict cookies to prevent CSRF attacks
+  - `SameSite=Strict` when the widget and the page share an origin, which is what prevents the
+    cookie being sent from another site
+  - `SameSite=None; Secure; Partitioned` when the widget is embedded on another domain, where
+    `Strict` would never be sent at all. `Partitioned` confines the cookie to that embedding site,
+    so it is not readable from any other page the browser visits
   - Secure flag automatically added on HTTPS connections
-  - __Host- prefix added on HTTPS to prevent subdomain attacks
   - Double Submit Cookie pattern with XSRF token matching the JWT ID
 
 This configuration should only be used when:
--- a/site/content/docs/configuration/parameters/index.md
+++ b/site/content/docs/configuration/parameters/index.md
@@ -79,7 +79,7 @@ services:
 | image.resize-height            | IMAGE_RESIZE_HEIGHT            | `900`                   | height of a resized image                                |
 | auth.ttl.jwt                   | AUTH_TTL_JWT                   | `5m`                    | JWT TTL                                                  |
 | auth.ttl.cookie                | AUTH_TTL_COOKIE                | `200h`                  | cookie TTL                                               |
-| auth.send-jwt-header           | AUTH_SEND_JWT_HEADER           | `false`                 | send JWT as a header instead of a server-set cookie; with this enabled, frontend stores the JWT in a client-side cookie. [See security considerations](#security-considerations-for-authsend-jwt-header). |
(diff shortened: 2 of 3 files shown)


paskal:
Added a third item, which is a one-line description string rather than prose, so this is no longer strictly documentation-only.

Both the `--auth.send-jwt-header` help text and the parameter table said the header is sent `instead of` the server-set cookie. `Service.Set` does neither of those things: it writes the header and then falls through to set both cookies, and the vendored source carries a comment explaining why, which is that headers do not survive the OAuth redirect. So an operator reading either description would expect the server to stop setting cookies once the flag is on, and would misjudge exactly the exposure the security section asks them to weigh.

Both now say the header is sent in addition to the cookies.

umputun:
The core of this is right and worth having: the separate-domain manual really did omit the setting that makes the arrangement work, and the parameter page really was left describing protections #2197 removed. Three corrections and one question.

**The link is broken.** `@/docs/configuration/parameters/index.md#...` is Zola syntax. This site is Hugo since #2179 and has no render-link hook, so Hugo emits that literally as a relative href. There are no other `](@/` links anywhere under `site/content/`, and the other manuals use the relative form, for example `reproxy/index.md:84` and `nginx/index.md:56` both link as `../../configuration/parameters/#anchor`. Use that. The site build stays green either way, which is why nothing catches it.

**The CHIPS sentence overstates the isolation**, on the page an operator reads to weigh risk. It says `Partitioned` "confines the cookie to that embedding site, so it is not readable from any other page the browser visits". The partition key is the top-level site including the scheme, so pages and subdomains under that same site share the cookie. MDN's own example has one partitioned cookie reachable from `shoppy.example`, `support.shoppy.example` and `checkout.shoppy.example` alike. It is isolated from a different top-level site, not from any other page.

**"Chrome increasingly" no longer holds.** Google's April 2025 position keeps ordinary Chrome on user choice and names Incognito as the mode blocking by default. "Safari, Chrome Incognito, and browsers configured to block third-party cookies" says the same thing and stays true.

**The question: is `AUTH_SAME_SITE=none` still part of the recommended arrangement?**

Reading the code, it looks redundant in the setup this manual now recommends, and not free. `authCookieOptions` in `common/cookies.ts:67` picks `SameSite=None; Secure; Partitioned` from `IS_THIRD_PARTY && isSecure` alone and reads nothing from the backend, so the client-written pair is unaffected by the flag. `AUTH_SAME_SITE` reaches exactly one place, `SameSiteCookie` at `cmd/server.go:1405`, so it governs only the server-set pair. Under a blocking browser that pair is dropped regardless, which this PR says. Under a permissive one, setting it to `none` broadens delivery of the `HttpOnly` JWT, and `corsMiddleware` reflects any origin with credentials, with its own comment noting that any site a signed-in reader visits can read authenticated responses.

That reasoning is a code trace, not a demonstration, and it would reverse guidance the project has carried for years, so it is a question rather than a correction. The experiment that settles it: take only the `remark42-https` service back to the default `AUTH_SAME_SITE` and run `TestHTTPS_CrossOriginSignInSurvivesAReload` and `TestHTTPS_SessionSurvivesThirdPartyCookieBlocking`.

Not the whole file, and this part matters: `TestHTTPS_AuthCookiesCarryTheThirdPartyForm` asserts every stored copy of both names is `SameSite=None`, so it should fail on the default server-set pair even while the reader stays signed in through the partitioned client pair. Using the file as the criterion would report the setting as necessary when it is not.

If both reload cases pass, the recipe can drop it or say it governs only the additional server-set cookies. If either fails, it names a dependency the code trace missed.

One small thing: the body says documentation only, and it edits `backend/app/cmd/server.go`. A flag description rather than behaviour, but it is a Go file.

paskal:
Three corrections applied and rebased onto master. The question is delegated and I will come back with a measurement rather than a guess.

**The link.** You are right, and it was the only `](@/` in the whole of `site/content`, so nothing else was following that pattern. Now the relative form the other manuals use. It also explains why nothing caught it: the build stays green because Hugo emits it as a literal href instead of failing.

**The CHIPS sentence.** Corrected, and this was the one worth catching, since overstating isolation on the page an operator reads to weigh risk is the wrong direction to be wrong in. It now says the partition key is the embedding top-level site, so a different site gets a separate cookie, while pages and subdomains under that same site share it.

**Chrome.** Corrected to Safari, Chrome Incognito, and browsers configured to block third-party cookies. I had carried "Chrome increasingly" over from the plan without checking whether it was still true, and it is not.

**The `AUTH_SAME_SITE` question.** I am not going to answer it from the same code trace you already did, because that would just be agreeing with you for no additional evidence. The e2e session owns that harness and is running exactly the experiment you specified: only `remark42-https` back to the default, then `TestHTTPS_CrossOriginSignInSurvivesAReload` and `TestHTTPS_SessionSurvivesThirdPartyCookieBlocking`, with `TestHTTPS_AuthCookiesCarryTheThirdPartyForm` explicitly excluded from the criterion, since it asserts every stored copy is `SameSite=None` and would report the flag as necessary when it is not. If both reload cases pass, I will change the recipe to drop the flag or scope it to the server-set pair; if either fails, it names a dependency the trace missed and I will document it as required.

**On the PR body.** Fair. It edits `backend/app/cmd/server.go`, and I noted that in a follow-up comment but never corrected the body itself, which is what a reader sees first. The body now says so.

paskal:
Ran your experiment, and the answer is that the flag is not required. The reason your control did not behave as predicted turned out to be worth more than the result, so both are below.

**Both reload cases pass without `AUTH_SAME_SITE`**, all four subtests: `TestHTTPS_CrossOriginSignInSurvivesAReload/{anonymous,email}` and `TestHTTPS_SessionSurvivesThirdPartyCookieBlocking/{anonymous,email}`. Only the `remark42-https` service was changed.

**`TestHTTPS_AuthCookiesCarryTheThirdPartyForm` also passes without it**, which you expected to fail. It passes because the server-set pair is not stored at all: a cross-site `Set-Cookie` lacking `SameSite=None` is refused outright, so there is nothing in the jar carrying the wrong attribute for the assertion to catch. The trace assumed that pair would be stored with a stricter attribute; it is simply absent. Your warning about not using the whole file as the criterion was right, for a different reason than the one given, and that reason is now a sentence in the manual.

The jar was dumped after an anonymous sign-in in each configuration instead of inferred. With the flag, four cookies:

| name | httpOnly | partition key | written by |
| --- | --- | --- | --- |
| `JWT` | true | none | server |
| `XSRF-TOKEN` | false | none | server |
| `JWT` | false | `https://host-site-https` | widget |
| `XSRF-TOKEN` | false | `https://host-site-https` | widget |

Without it, two: the client-written partitioned pair alone. Every copy is `SameSite=None; Secure` in both configurations, which is why the attribute case is green either way.

So the setting is observably doing something, which is what makes the green above meaningful instead of a run where nothing moved, and what it does is exactly the cost you suspected: an unpartitioned `HttpOnly` JWT stored and delivered as a third-party cookie to every listed domain wherever the browser still permits them. Persistence does not depend on it in either kind of browser.

The recipe now drops it, with that table in the manual so the reversal carries its evidence rather than asking anyone to take my word for it. It stays documented for the arrangement that does need it, which is one without `AUTH_SEND_JWT_HEADER`, where the server's cookies are the only ones there are.

Credit where due: the experiment and the jar dump are the e2e session's work on its harness, not mine.

umputun (approved):
lgtm

PR #2222, opened 2026-08-23
Document what each browser actually does with cross-domain auth

paskal:
The separate-domain manual now says which settings to use, but not what a reader in a given browser actually gets. This adds that, measured instead of reasoned.

**How it was measured.** Remark42 on three subdomains of one registrable domain, the host page on a second registrable domain, real wildcard certificate, real DNS, no resolver flags. Three server configurations (header flag only, `AUTH_SAME_SITE=none` only, both) across Chromium, Firefox and WebKit, each under its default cookie policy and under blocking, signing in and then reloading. Every blocked column carries a control: an ordinary third-party cookie written from inside the widget frame has to be dropped, or the run is not blocking anything and its passes mean nothing. That control caught two runs that would otherwise have been reported as evidence.

Three results the manual did not carry:

**Safari blocks third-party cookies with no configuration**, so `AUTH_SAME_SITE=none` on its own has already stopped working there. The old recipe is broken today for those readers, not deprecated for later.

**Firefox reaches a working session by a weaker route.** Chromium and WebKit refuse the server's cookie when it carries no `SameSite` attribute, which leaves the field clear for the widget to write its own `Partitioned` pair. Firefox accepts that cookie, and since the server's `JWT` is `HttpOnly`, the browser then forbids the widget's script from replacing it, because a cookie set by JavaScript may not overwrite an `HttpOnly` one of the same name. So on Firefox the session rides on an ordinary unpartitioned third-party cookie even with the header flag on, and it goes the moment the reader blocks those. The cookie jars show it directly: Chromium and WebKit end with two `httpOnly=false` cookies carrying a partition key, Firefox with the server's `httpOnly=true` pair and no partitioned copy at all.

**Firefox's "block all third-party cookies" discards partitioned cookies too**, so the CHIPS escape hatch does not apply and no configuration survives it.

Two parameter descriptions were wrong in ways that matter here, and are corrected in the same commit. `AUTH_SAME_SITE=default` emits no `SameSite` attribute at all and leaves the choice to the browser; it is not `Lax`, and the manual said it was. That absence is exactly what makes the recommended configuration work on Chromium and Safari, by getting the server's cookie refused so the widget's partitioned one can land. And `AUTH_TTL_COOKIE` does not govern the cookie that carries the session once the header flag is on, because the frontend hardcodes 200h to mirror the default, so an operator who shortens it changes nothing.

Documentation only.

### DIFF
--- a/site/content/docs/configuration/parameters/index.md
+++ b/site/content/docs/configuration/parameters/index.md
@@ -78,9 +78,9 @@ services:
 | image.resize-width             | IMAGE_RESIZE_WIDTH             | `2400`                  | width of a resized image                                 |
 | image.resize-height            | IMAGE_RESIZE_HEIGHT            | `900`                   | height of a resized image                                |
 | auth.ttl.jwt                   | AUTH_TTL_JWT                   | `5m`                    | JWT TTL                                                  |
-| auth.ttl.cookie                | AUTH_TTL_COOKIE                | `200h`                  | cookie TTL                                               |
+| auth.ttl.cookie                | AUTH_TTL_COOKIE                | `200h`                  | TTL of the server-set auth cookie. Note it does not govern the cookie the frontend writes under `auth.send-jwt-header`, which is fixed at 200h |
 | auth.send-jwt-header           | AUTH_SEND_JWT_HEADER           | `false`                 | also send JWT as a header, so the frontend can store it in a client-side cookie that survives third-party cookie blocking; the server-set cookies are still sent. [See security considerations](#security-considerations-for-authsend-jwt-header). |
-| auth.same-site                 | AUTH_SAME_SITE                 | `default`               | set same site policy for cookies (`default`, `none`, `lax` or `strict`) |
+| auth.same-site                 | AUTH_SAME_SITE                 | `default`               | SameSite attribute for the server-set auth cookies (`default`, `none`, `lax` or `strict`). `default` emits no attribute at all and leaves the choice to the browser, which is not the same as `lax` |
 | auth.apple.cid                 | AUTH_APPLE_CID                 |                         | Apple client ID (App ID or Services ID)                  |
 | auth.apple.tid                 | AUTH_APPLE_TID                 |                         | Apple service ID                                         |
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm

paskal:
The Safari column is now Safari, not WebKit.

It was WebKit through Playwright, which is the engine but not the browser: ITP sits above WebKit and could plausibly have been stricter. Driven through Safari 27's own WebDriver against the same rig, all three configurations match the WebKit result exactly. So the column says Safari and means it, and as a side effect the e2e suite's existing WebKit coverage is a faithful proxy for this behaviour rather than an assumption that it is one.

The evidence for the recommended configuration, read from inside the widget frame on real Safari: `document.cookie` returns the partitioned sentinel and a JS-readable `JWT`, which is the widget's own cookie rather than the server's `HttpOnly` one, while the ordinary third-party control cookie written beside them is gone. Sign-in succeeds and the reload keeps the reader signed in. On `AUTH_SAME_SITE=none` alone the reload fails, with no configuration touched on the browser side.

Worth noting what it took to get there, since it will bite the next person: `sudo safaridriver --enable` and Develop > Allow remote automation are both necessary and neither is sufficient while a normal Safari instance is running. With one open, session creation times out after 30 seconds with no diagnostic, which reads like a permissions problem and is not one. safaridriver also serves one session at a time, so a session left undeleted blocks the next.

umputun:
Merged. The Firefox row makes two older sentences in the same file false, and they are on master now.

Line 15 says `AUTH_SEND_JWT_HEADER` keeps a reader signed in "in any browser configured to block third-party cookies". Line 83 says the recommended arrangement survives a reload "whatever the browser's third-party cookie policy". Line 73, added here, says no configuration survives Firefox's "block all third-party cookies" setting.

Both predate this PR and neither needs new evidence, so they are mine to fix rather than yours unless you are already in that file.

paskal:
Both folded into #2223, which was already open against the same file.

You were right that they needed no new evidence, and the Firefox row is what made them false. They are now scoped to Firefox's block-all setting specifically, so the table, the prose and the Firefox paragraph agree on the one case where no configuration survives.

PR #2223, opened 2026-08-23
Measure the Safari column on Safari, and fix two sentences it makes false

paskal:
Two things against `manuals/separate-domain`, both following from #2222.

**The Safari column is now Safari, not WebKit.** That table measured WebKit through Playwright, which is Safari's engine but not Safari: ITP sits above WebKit and could plausibly have been stricter, and the table did not say which had been measured. Driven through Safari 27's own WebDriver against the same rig, all three configurations match the WebKit result exactly. A useful side effect is that the e2e suite's existing WebKit coverage is now a measured proxy for Safari on this behaviour instead of an assumed one.

Evidence for the recommended configuration, read from inside the widget frame on real Safari with blocking confirmed by a control cookie: `document.cookie` returns the partitioned sentinel and a JS-readable `JWT`, which is the widget's own cookie rather than the server's `HttpOnly` one, while the ordinary third-party control written beside them is gone. Sign-in succeeds and the reload keeps the reader signed in. On `AUTH_SAME_SITE=none` alone the reload fails, with nothing configured on the browser side.

**Two sentences that #2222 made false.** Both predate it and both are on master, so they currently contradict the table three lines away from them. One said the header flag keeps a reader signed in in any browser configured to block third-party cookies; the other said the recommended arrangement survives a reload whatever the browser's third-party cookie policy. Firefox's "block all third-party cookies" is the exception to both, because it discards partitioned cookies as well, and no configuration survives it. Both are now scoped to that case, so the table, the prose and the Firefox paragraph agree.

umputun offered to take these himself; folding them in here since this branch was already editing the file.

Documentation only, one file.

### DIFF
--- a/site/content/docs/manuals/separate-domain/index.md
+++ b/site/content/docs/manuals/separate-domain/index.md
@@ -12,9 +12,9 @@ Unless discussion [#1139](https://github.com/umputun/remark42/discussions/1139) 
 
 Set `ALLOWED_HOSTS="'self',https://example1.org,https://example2.org"` with your domain names, and `AUTH_SEND_JWT_HEADER=true`.
 
-`AUTH_SEND_JWT_HEADER` is what keeps a reader signed in across a page reload on Safari, in Chrome Incognito, and in any browser configured to block third-party cookies. Read [its security considerations](../../configuration/parameters/#security-considerations-for-authsend-jwt-header) before enabling it: it puts the token in a cookie JavaScript can read, which costs XSS exposure that a server-set `HttpOnly` cookie does not.
+`AUTH_SEND_JWT_HEADER` is what keeps a reader signed in across a page reload on Safari, in Chrome Incognito, and in Chrome configured to block third-party cookies. The one setting it does not survive is Firefox's "block all third-party cookies", which discards partitioned cookies as well, and which no configuration survives. Read [its security considerations](../../configuration/parameters/#security-considerations-for-authsend-jwt-header) before enabling it: it puts the token in a cookie JavaScript can read, which costs XSS exposure that a server-set `HttpOnly` cookie does not.
 
-**`AUTH_SAME_SITE=none` is not needed alongside it**, which reverses what this page recommended for years, so it is worth showing the measurement instead of asserting it. Signing in anonymously from an embedded widget over https and dumping the browser's cookie jar gives, with the setting:
+**`AUTH_SAME_SITE=none` is not needed alongside it**, which reverses what this page recommended for years, so it is worth showing the measurement instead of asserting it. Signing in anonymously from an embedded widget over https on Chromium with its default third-party cookie policy, and dumping the browser's cookie jar, gives with the setting:
 
 | name | httpOnly | partition key | written by |
 | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


paskal:
Folded your two #2222 follow-ups in here, since this branch was already editing the file.

Both are now scoped to Firefox's block-all setting, which is the single exception in the table and sits three lines from each of them. The header sentence says Chrome configured to block rather than any browser configured to block, and names the exception explicitly; the combinations bullet says every third-party cookie policy except that one, which no configuration survives.

That leaves the file internally consistent: the table, the two prose sentences and the Firefox paragraph now all say the same thing about the one case where nothing works.

umputun:
the two sentence fixes are right, and measuring the Safari column through Safari itself is worth having. #2222 left two more passages inconsistent with its own browser results, plus one measurement that needs its browser named.

**the `Partitioned` paragraph**, the one opening "`SameSite=None` is not sufficient on its own". It says a browser blocking third-party cookies drops the cookie unless it is explicitly `Partitioned`, and that `AUTH_SEND_JWT_HEADER=true` is what closes that. Firefox's "block all third-party cookies" discards partitioned cookies too, as the page says three paragraphs later, so the flag closes nothing in that mode. Same exception as the two sentences you already fixed.

**the first two combination bullets.** Both say authorisation works only on the install domain and its subdomains. `token.Service.Set` always writes the server `JWT` and `XSRF-TOKEN` pair, and `SendJWTHeader` only adds the `X-JWT` header on top (`token/jwt.go:269-286`, which falls through to the cookies deliberately). Your Firefox paragraph has default Firefox accepting the attribute-less server cookie. Together those predict that Email, Telegram and anonymous auth do work cross-domain on Firefox default with neither variable set. The matrix has no row for that configuration, so pls run it on the rig before touching the prose. If it passes, both bullets need a Firefox-default exception.

**the cookie-jar dump** doesn't name its browser. The four rows match permissive Chromium. It cannot be Firefox default, because the `HttpOnly` overwrite explanation below it means the widget's own `JWT` write is refused there, and Safari and blocked Chromium both discard the server pair. A clause naming Chromium is enough.

the `Partitioned` paragraph is a correctness fix. The Firefox-default case is a measurement request first, and a correctness fix only if it passes. Naming Chromium is clarity. Happy to take any of them myself if you'd rather not reopen the branch.

paskal:
You were right, so both bullets needed the exception. Measured on the rig: with neither variable set, anonymous and email both survive a reload cross-domain on stock Firefox. The jar holds only the server's own `JWT` and `XSRF-TOKEN`, no widget-written copy, so it rides on the attribute-less cookie that Total Cookie Protection partitions instead of refusing.

Changed:

- both combination bullets now carry the Firefox-default exception, with the mechanism stated so it does not read as a special case
- the `Partitioned` paragraph carries the block-all exception, the same one as the two sentences last round
- the cookie-jar dump names Chromium on its default policy
- new table row for the default configuration, all five cells measured: `neither set | fails | fails | works | fails | fails`
- prose under the table now says that row holds on one engine only, so it does not read as a recommendation

Both new blocked cells carry the control the table promises: an ordinary third-party cookie written from inside the frame is dropped in each. On Safari the backend log confirms the login arrived and the widget showed the reader signed in, so only the reload loses it. Safari accepts neither `--host-resolver-rules` nor `network.dns.localDomains`, so the whole row ran on `localtest.me` and `nip.io` names that resolve to 127.0.0.1 through public DNS, which every browser reaches unaided.

umputun (approved):

PR #2224, opened 2026-08-25
Stop prefixing the instance origin onto an already absolute asset url

paskal:
The oauth provider icons and the ghost avatar fail to load on every `master` build since #2197. Reported from a live instance; the screenshot is three empty circles where the provider icons belong.

**What happens.** Both components build their `src` as `${BASE_URL}${icon}`, where `icon` comes from a webpack `require` of an svg. That was correct while `output.publicPath` was the fixed `/web/`, because the require then resolved to a relative `/web/name.svg` and the concatenation produced exactly one origin. #2197 changed it to `'auto'`, which the runtime resolves to an absolute url at bootstrap, so the same expression now yields `https://hosthttps://host/web/google.svg`.

**Blast radius.** Every `master` build, embedded or same-domain, since `'auto'` resolves absolute either way. No release is affected: `git tag --contains` returns nothing for #2197, and the newest tag still carries the fixed public path. So it is `:master` users only.

**Why nothing caught it.** The jest stub at `app/__stubs__/svg.tsx` returns a relative `/image.svg`, so a component prefixing the origin onto an asset url looks correct under test while being wrong in every real build. The avatar spec asserted the doubled form as the contract. The stub now returns an absolute url like the real thing, which makes that spec fail the moment the prefix returns; verified by putting it back.

**The durable guard is e2e**, because only e2e sees real asset urls. `TestAssets_WidgetImagesActuallyLoad` opens the auth panel and checks every image in the widget document. It asserts `naturalWidth`, not the presence of the element or its `src`, since a broken `` keeps both and asserting either proves nothing. It separately asserts no url carries more than one `://`, so a recurrence names its cause rather than its symptom.

Mutation-checked both ways. With the prefix restored the e2e test fails with `image url carries more than one origin ... http://remark42:8080http://remark42:8080/web/dev.svg` and the unit spec fails alongside it; with the fix in place both pass.

### DIFF
--- a/e2e/assets_test.go
+++ b/e2e/assets_test.go
@@ -81,3 +81,60 @@ func TestAssets_WidgetRunsOnTheCompiledInURL(t *testing.T) {
 		"the widget addressed %q rather than this instance", configURL)
 	assert.NotContains(t, configURL, remarkURLPlaceholder)
 }
+
+// TestAssets_WidgetImagesActuallyLoad covers a class of defect the unit suite structurally cannot
+// see. Asset URLs come from webpack's public path, and the jest stub for `.svg` returns a relative
+// string, so a component that prefixes an already-absolute asset URL with the instance origin
+// looks right under test and ships a doubled URL. That is how the oauth provider icons and the
+// ghost avatar came to 404 on every master build: `${BASE_URL}${icon}` produced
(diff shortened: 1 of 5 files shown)


umputun (approved):
lgtm

PR #2226, opened 2026-08-26
Add a frontend test layer that runs from source with no toolchain, and fix six defects

paskal:
A second frontend test layer, for everything that does not need a browser. Groundwork for dropping jest.

`*.unit.test.ts` files import nothing but the module under test, its fixtures and `node:test`. Plain node runs them straight from the TypeScript, with no bundler, no jsdom and no `node_modules`:

```
pnpm test:unit
```

236 cases across 15 files, and the jest cases they replace are removed in the same change, taking jest from 437 to 358. The `unit` job in `ci-frontend.yml` runs them with no `pnpm install` step at all, which is the point: a test that starts needing `node_modules` to pass has acquired a dependency, and the job goes red instead of hiding it. The job takes about 10 seconds.

`node --test` exits 0 on a glob that matches nothing, so on its own it cannot tell a renamed suffix from a passing run; the CI step counts the files first and fails when none match. Node's ESM resolver is also why value imports inside this layer write the extension in full and use relative paths, while type-only imports erase before node sees them and keep the ordinary style.

## What moved

Each of these was a pure computation inside a module that reads a browser global at import scope, which makes the whole file unimportable outside a browser however pure the computation is. Public API is unchanged throughout.

| new module | held back by | what it decides |
|---|---|---|
| `common/intl-message.ts` | Preact | every string a reader sees, and what happens to a broken translation |
| `common/cookie-format.ts` | `document`, `window.parent` | whether an auth cookie is delivered at all in a third-party frame |
| `common/fetcher-core.ts` | settings, cookies, StaticStore | request URLs, bodies, auth headers, the server clock reading |
| `common/base-url.ts` | `window.remark_config`, `window.location` | which host the widget loads itself from, attacker-controlled |
| `common/json-store.ts` | `localStorage`, probed at import | what a caller gets back when storage holds something unusable, or refuses |
| `store/thread/collapse-store.ts` | `localStorage` | whether a reader's collapsed threads survive a reload |
| `store/comments/comment-tree.ts` | `localStorage`, through a neighbour in the same file | which comments a blocked user's are hidden from |
| `utils/hidden-users.ts` | `localStorage` | a record the comment list is indexed by, writable by anything on the page |
| `hooks/session-storage.ts` | `sessionStorage` | what a caller gets back when storage holds something unusable |

`utils/errorUtils.ts` and `store/user/reducers.ts` needed no new file: the first only imported `defineMessages` from `common/intl`, which pulls Preact, and the second was already pure. `formatjs` matches messages by identifier and never by import source, so extraction is unaffected.

## Behaviour changes

Three defects a reader could hit:

- A permanent user block sent `ttl=undefined`, since `URLSearchParams` renders an `undefined` value as that literal string. It reached the right outcome only because `admin.go` fails to parse it and falls back to unlimited. `requestURL` omits the parameter, reaching that branch deliberately.
- `readCookie` handed a malformed percent escape to `decodeURIComponent`, which raises `URIError`, and `fetcher.ts` reads the XSRF cookie while building a request, outside its own `try`. A single bad cookie rejected every API call for as long as it was there, and anything on the embedding page can write it.
- An empty `XSRF-TOKEN` cookie produced an empty header, which lighttpd answers `400` to. Nothing is lost by omitting it, since go-pkgz/auth reads the header with `Header.Get` and cannot distinguish an empty one from a missing one.

Three hardening changes, none of them reachable today:

- The host protocol check was `startsWith('http')`, which admits any scheme beginning with those four letters. It now compares against `http:` and `https:` exactly. `javascript:` and `data:` were always refused, so this was never a hole.
- An error code with no catalogue entry made the error handler itself throw, leaving the reader with nothing instead of the failure. All three `RequestError` construction sites pass a code the catalogue carries, so it needed a fourth to be reached; it now falls back to the generic message.
- `setCookie` rewrote the caller's options object in place while converting `expires`. No caller reuses one, so nothing observed it; `serializeCookie` copies first, which matters because clearing a cookie only works when every attribute matches the ones it was set with.

## What jest loses

Four suites go outright, each covered more thoroughly here: `utils/parse-query.test.ts`, `utils/jwt.test.ts`, `utils/isUserAnonymous.test.ts` and `utils/get-hidden-users.test.ts`. `jwt.test.ts` was worth removing on its own account: it defines `window.Data` and not `Date`, so its clock mock never applied, and the case named `should be not expired` asserted the opposite of what `jwt.unit.test.ts` pins.

Seven shrink to the half only a browser can reach. `common/intl.test.tsx` keeps `<FormattedMessage/>` rendering, `useIntl` outside a provider and the sweep over every shipped catalogue; three of its cases the layer did not yet carry were moved across instead of dropped. `common/fetcher.test.ts` keeps its fetch integration, `common/cookies.test.ts` keeps what it writes to `document`, and `hooks/useSessionState.spec.ts` keeps the two cases about the hook itself.

Three keep a single case each for a boundary the pure modules cannot reach: that the configured host and the protocol of the page both arrive at the host validator, and that the two storage wrappers reach the real `localStorage`. Without them, hard-coding the protocol or making either writer a no-op goes unnoticed.

Coverage comes from the same run: node emits lcov through `--experimental-test-coverage`, reporting 100% of lines and 100% of branches across the nineteen files it reaches.

The contributing guide, `frontend/CLAUDE.md` and `CLAUDE.md` document the layer, the two import rules and why they hold for value imports alone.


umputun:
the layer itself is good and I have no correctness objection. Each of the five behaviour changes is exercised by a unit test named for it. Every reachable value import from the 15 unit entries follows the relative-path and full `.ts` extension rules, with type-only imports keeping the ordinary style as intended. FormatJS extracts 180 keys at both the merge base and this head, matching `en.json`, so the silent-zero failure `frontend/CLAUDE.md` warns about is not in play. The 62% drop in `deleteme.mjs` is worth saying out loud: `errorUtils` no longer imports `common/intl`, so that entry stops carrying the whole i18n graph.

what I would change is the accounting, not the code. The title describes test infrastructure while the PR also changes five behaviours, which hides the fixes in the PR history, makes them easy to miss when the release notes are written, and leaves a bisect pointing at the same 48-file commit. #2197 carried seven defects and said so in its title. I would retitle to mention the fixes, add the `setCookie` change to the description, and call the unknown-code fallback hardening.

two details make "Public API is unchanged throughout" broader than the code:

- `serializeCookie` copies its options, so exported `setCookie` no longer rewrites a numeric or `Date` expiry in place. No in-repo caller depends on that mutation, and the new test pins the copy deliberately, but it is an observable behaviour change.
- `store/comments/utils.ts` no longer exports `filterTree`, `findPinnedComments` or `getPinnedComments`. Every in-repo consumer moved with them and there is no library entry point, so the widget API is unchanged; the internal module surface is not.

the unknown-code fallback is hardening rather than a fix for a reachable defect. Every production `RequestError` construction passes `0`, `fetch-error`, or a status `errorMessages` already carries, and the test says no current construction site passes an unknown code.

one small thing for later, since the new tests cover the extracted helpers rather than the wrappers around them: `cookies.test.ts` only round-trips a valid value and `fetcher.test.ts` mocks `getCookie` as undefined and never as empty. Reverting either helper fails a test today, but a wrapper bypass could bring the malformed-cookie or empty-header behaviour back with the unit layer still green.

paskal:
Commit message and text of the PR are fixed.

frontend/apps/remark42/app/common/json-store.unit.test.ts
```diff
@@ -0,0 +1,187 @@
...
+function captureErrors() {
+  const real = console.error;
+  const seen: unknown[][] = [];
+
+  console.error = (...args: unknown[]) => {
+    seen.push(args);
+  };
+
+  return { seen, restore: () => (console.error = real) };
+}
```
akellbl4: I think mocks/spies would work better than doing all this manually

akellbl4 (dismissed):

umputun (approved):
lgtm

PR #2228, opened 2026-08-27
Drive Telegram authentication and subscriptions through the browser

paskal:
Telegram is the one authentication provider whose flow leaves the browser: the reader messages a bot, and no page can reach that. This makes the round trip drivable end to end, for both authentication and notification subscriptions.

`backend/go.mod` carries the released `go-pkgz/notify` v1.5.0 and `go-pkgz/auth` v2.3.0, both of which ship the settable bot API base this needs.

## The fixture

`e2e/telegramstub` answers as the bot API for the four calls the flow makes, and takes the reader's side of the exchange through `/control/send`, which is the step inside Telegram that a browser cannot perform. It runs on its own instance, because adding an auth provider to the main one would change the auth panel every other case reads.

## The setting it needs

Reaching a stub means the bot API base URL has to be settable, which is what `--telegram.api-url` adds. It serves a real operator need beyond the tests: a proxy or a self-hosted Bot API server, in places where Telegram is blocked. Both consumers honour it, the auth provider and the notification service. The value is an origin, and the bot token travels in the request path, so it has to be a host the operator controls.

## Two defects the coverage found

An instance with Telegram authentication enabled and notifications failing registered the provider and then started nothing to listen with, so it accepted the configuration and stayed permanently deaf to it. And a nil `*notify.Telegram` was assigned into an interface, where it is not nil, so the guard downstream let it through.

## What it adds to backend coverage

Measured with the e2e coverage reporting from #2232: the full browser suite run against an instrumented stack, once on #2232 alone and once with this change on top, so the only difference is this branch and its two Telegram tests.

**37.9% to 40.4% of backend statements, +142 newly covered.**

| | covered | total | |
|---|---|---|---|
| #2232 alone | 1938 | 5099 | 37.9% |
| with this change | 2080 | 5131 | 40.4% |

The 32 extra statements in the denominator are this branch's own production code, so the gain is not an artefact of measuring less.

| package | before | after | newly covered |
|---|---|---|---|
| `app/providers` | 0.0% | 86.4% | +19 |
| `app/notify` | 34.0% | 46.4% | +33 |
| `app/rest/api` | 40.0% | 42.9% | +44 |
| `app/cmd` | 30.6% | 33.4% | +34 |
| `app/store/service` | 48.6% | 49.9% | +9 |
| `app/store/engine` | 54.6% | 55.1% | +3 |

`app/providers` is the one worth pointing at. It holds a single function, `DispatchTelegramUpdates`, and nothing else in the suite reaches it: it was the only package at exactly 0% across the whole e2e run. The two defects above are covered rather than only fixed, at 83.3% for `makeTelegramAuth` and 64.7% for `startTelegramAuthAndNotify`.


paskal:
## What this adds to backend coverage

Measured with the e2e coverage reporting from #2232, running the full browser suite twice against an instrumented stack: once on #2232 alone, once with this branch's commit on top of it. Same machinery, same suite, the only difference being this change and its two Telegram tests.

**37.9% to 40.4% of backend statements, +142 newly covered.**

| | covered | total | |
|---|---|---|---|
| #2232 alone | 1938 | 5099 | 37.9% |
| with this branch | 2080 | 5131 | 40.4% |

The 32 extra statements in the denominator are this branch's own production code, so the gain is not an artefact of measuring less.

Where it lands:

| package | before | after | newly covered |
|---|---|---|---|
| `app/providers` | 0.0% | 86.4% | +19 |
| `app/notify` | 34.0% | 46.4% | +33 |
| `app/rest/api` | 40.0% | 42.9% | +44 |
| `app/cmd` | 30.6% | 33.4% | +34 |
| `app/store/service` | 48.6% | 49.9% | +9 |
| `app/store/engine` | 54.6% | 55.1% | +3 |

`app/providers` is the one worth pointing at. It holds a single function, `DispatchTelegramUpdates`, and nothing else in the suite reaches it: it was the only package sitting at exactly 0% in the whole e2e run. The two wiring defects described above are covered rather than only fixed, at 83.3% for `makeTelegramAuth` and 64.7% for `startTelegramAuthAndNotify`.

The measurement branch is throwaway. This branch is unchanged and unrebased.

umputun:
tags are cut, so the pins can become released versions: `go-pkgz/auth` v2.3.0 and `go-pkgz/notify` v1.5.0. The auth tag is on master head `5f6d12c4`, one commit past `d9c7fd35`, which adds a test race fix and no behaviour.

one more thing while you are in there: #2232 is merged, and `coverage.sh` takes both its must-produce-coverage list and its merge inputs from `compose-e2e-coverage.yml` alone. `remark42-telegram` is only in the base compose file, so without a stanza there with its own `GOCOVERDIR` and mount, `make e2e-cover` skips the Telegram instance and still reports success.

rest of the PR reads fine to me.

paskal:
Both pins are the released tags now, and the branch is rebased onto master.

`compose-e2e-coverage.yml` gains the `remark42-telegram` stanza; it was being skipped exactly as you describe.

umputun (approved):

PR #2230, opened 2026-08-27
Say what the auth limiter is actually keyed by

paskal:
The comment on `pauseForAuthLimit` describes the `/auth/` token bucket as keyed by instance and client IP. Tollbooth builds the key from the client IP and the request path, joined, and sets the burst to the rate, so a refill interval covers repeat requests by one reader to one path and two paths never compete at all.

The same comment says a fresh context needs no delay, while all thirteen call sites sleep unconditionally. Both statements are true and they read as a contradiction; the comment now says which calls the sleep is there for, namely repeat requests inside one context and the `/auth/` probes this process makes itself.

Comment only, no behaviour change.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -142,9 +142,14 @@ var (
 	}
 )
 
-// Everything under /auth/ is a token bucket refilling twice a second per instance and client IP.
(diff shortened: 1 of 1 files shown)


umputun (approved):
lgtm

PR #2231, opened 2026-08-28
Claim a site before accepting its migration, and clean up a failed upload

paskal:
Two requests could start a migration for the same site at once, and a failed upload left a temporary file behind.

## The race

`Migrator` already tracks a busy site behind a mutex, but the handlers checked and claimed in two steps: `isBusy` first, then acceptance returned, and only the background `runImport` set the flag. Between the check and that goroutine running, a second request saw the site as free and was accepted too, so both imported into the same site.

`setBusyIfFree` makes that a single compare-and-set under the lock the type already has, and the handlers call it before returning acceptance. `remapCtrl` never claimed at all and now does. The slot is released by the background work as before, and by the handler itself if it fails before handing off, so a rejected upload does not strand the site.

**A behaviour change worth knowing about**: another migration for the same site is refused with `409` from the start of the upload until the upload and import finish, or until reading the upload fails. Previously the window began only when the import did. The endpoint is administrator-only and the multipart body is capped at 256 MiB, which is what makes holding the slot for the upload the right trade against overlapping writes.

## The temporary file

`saveTemp` returned early when `io.Copy` failed, leaving the partial file on disk with its descriptor open. It now closes and removes an incomplete file on every failure path.

## Tests

The regression test blocks a handler between its claim and the hand-off, fires a competing request, and asserts both the `409` and that the slot is released afterwards. Raw and multipart uploads are covered on the claim and on the release, `remapCtrl` has an assertion against an occupied slot, and the temporary directory is checked for leftovers.

Each was mutation-checked separately: restoring the check-then-set ordering, removing the failed-upload release, dropping the remap claim, and removing the `saveTemp` cleanup each fail the case that names them.

`go test -race`, `golangci-lint`, `go build` and `gofmt` pass.

### DIFF
--- a/backend/app/rest/api/migrator.go
+++ b/backend/app/rest/api/migrator.go
@@ -48,11 +48,17 @@ type KeyStore interface {
 func (m *Migrator) importCtrl(w http.ResponseWriter, r *http.Request) {
 	siteID := r.URL.Query().Get("site")
 
-	if m.isBusy(siteID) {
+	if !m.setBusyIfFree(siteID) {
 		rest.SendErrorJSON(w, r, http.StatusConflict, fmt.Errorf("already running"),
 			"import rejected", rest.ErrActionRejected)
 		return
 	}
+	handedOff := false
+	defer func() {
+		if !handedOff {
+			m.clearBusy(siteID)
+		}
+	}()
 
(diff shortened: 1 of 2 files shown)


umputun (approved):
lgtm

PR #2236, opened 2026-08-30
Fix: serve compiled bundles under /web/ in dev mode

Bluetegu:
## Title

Fix(dev-only): serve compiled bundles under /web/ in webpack-dev-server

## Body

`output.publicPath: 'auto'` (from #2197, needed for instances mounted under a URL prefix) also
changed where `webpack-dev-server` itself serves the compiled bundles in dev mode, which no longer
matches `devServer.static`'s entries or any template's hardcoded `/web/<name>.mjs` script tags. The
documented `pnpm dev` workflow (`http://127.0.0.1:9000/web/`) 404s as a result — everything is
served from the root instead. This only affects the local dev server; production is unaffected.

Fix: `publicPath: isDev ? PUBLIC_PATH : 'auto'`. Production keeps `'auto'` untouched (verified via
`pnpm build`, no hardcoded `/web/` in the emitted bundle), so this doesn't reintroduce the bugs
#2219 and #2224 fixed for that mode.

### Verification

- `pnpm dev`: `/web/`, `/web/embed.mjs`, `/web/counter.mjs`, `/web/remark.mjs` all resolve, and the
  demo page's widget, last-comments iframe, and counter all render correctly.
- `pnpm build` still compiles cleanly with `'auto'` in production.
- `pnpm lint` passes.

Screenshot 2026-08-30 at 13 02 44

### Notes

Developed with Github Copilot

### DIFF
--- a/frontend/apps/remark42/webpack.config.js
+++ b/frontend/apps/remark42/webpack.config.js
@@ -179,6 +179,10 @@ module.exports = (_, { mode, analyze }) => {
     port: PORT,
     devMiddleware: {
       stats: 'minimal',
+      // serve the in-memory build output under the same PUBLIC_PATH the devServer.static entries
+      // and every template's script tags use. output.publicPath stays 'auto' for the runtime
+      // bundle; this only affects where the dev server itself serves its own output from.
+      publicPath: PUBLIC_PATH,
     },
     headers: {
       'Access-Control-Allow-Origin': '*',


paskal (dismissed):

paskal (commented):
lgtm, but set it on the dev server instead of on `output`. Keep `output.publicPath: 'auto'` and add the path to `devMiddleware`:

```js
devMiddleware: {
  stats: 'minimal',
  publicPath: PUBLIC_PATH,
},
```

Same routing, but it moves only where the dev server serves its output, so dev keeps the same runtime resolution production uses. Splitting `output.publicPath` by mode is where #2203, #2219 and #2224 each came from: the broken arrangement was always the one nobody runs locally.

No CI has run on this branch yet, so the frontend checks have not covered it.

Bluetegu:
Done — moved to `devMiddleware.publicPath`, `output.publicPath` is back to unconditional `'auto'`. Verified `/web/`, all `.mjs` bundles, and HMR resolve correctly; `pnpm build` and `pnpm lint` still pass.

Screenshot 2026-08-30 at 14 37 30

paskal (approved):

umputun (approved):

PR #2239, opened 2026-08-30
Remove the jest cases the browser suite already asserts

paskal:
The browser suite drives the real widget and #2234 reports what it reaches, so the jest cases it duplicates are two tests of one behaviour: one fails when the widget breaks, the other tells nobody anything new.

Twenty-four cases go, across nine suites. Thirty-three suites give up nothing.

Stacked on #2228 and #2234; the reviewable diff is the last commit alone.

## The rule

A case may go only when a **named** e2e case asserts the same behaviour, and the unit of the decision is the **assertion**, not the case. A case asserting that a control renders *and* that its countdown reads `300s` needs both covered; the browser suite only waits for that element to appear, so that one was split and the countdown assertion kept.

Coverage rules candidates out, never in. A component renders during unrelated browser tests and marks its lines covered while nothing checks it behaves correctly.

The commonest claim is that a browser case clicking a control covers a jest case asserting it renders. Playwright's `Click` requires the element to exist, be visible, enabled and stable, and renaming the `Vote up` title makes `TestVote_UpvoteCountsOnce` fail at its click, at `vote_test.go:49`, rather than somewhere incidental downstream.

## What went, and from where

| suite | before | after |
|---|---|---|
| `comment.test.tsx` | 10 | 3 |
| `comment-actions.spec.tsx` | 16 | 12 |
| `comment-votes.spec.tsx` | 13 | 9 |
| `comment-form.spec.tsx` | 17 | 14 |
| `create-iframe.test.ts` | 9 | 7 |
| `auth.spec.tsx` | 18 | 17 |
| `comment-form__subscribe-by-telegram` | 14 | 13 |
| `profile.spec.tsx` | 12 | 11 |
| `sort-picker.spec.tsx` | 4 | 3 |

`comment.test.tsx` gave up the most because its cases were almost all "does this element render", which a click proves.

## Why thirty-three suites gave up nothing

They assert redux action objects, spy call counts, `postMessage` payloads, `execCommand` calls, CSS-module class names, refs, icon dimensions and translation-tag fallbacks. The browser suite asserts none of those, and mostly should not: asserting a request and its response is the better test of the same thing.

## What the deletions were checked against

Jest coverage was generated with the deletions reverted and again with them applied, on this same tree, and the covered-line sets diffed against the widget profile from `make e2e-cover`. **Jest now covers no line it covered before that the browser suite does not also cover.**

That check is necessary and not sufficient, and three deletions got past it:

- the auth dropdown ignoring a `clickOutside` whose source is not `window.parent`. `auth.hooks.ts:30` is entered by nothing else, because the only such message the browser suite sends comes from the real parent and takes another branch
- Hide rendering for an ordinary reader on another reader's comment. `TestThread_HideUserRemovesTheirCommentsOnly` clicks Hide as the dev user, which the stack gives `ADMIN_SHARED_ID`, so it would pass unchanged if the control became admin-only
- a duplicate pair deduped the wrong way round, leaving a case whose title says it asserts an empty form while its body asserts a restored draft

The first came from the coverage diff; the other two from review, and neither could have been caught by it, since equal lines can carry different assertions. All three are resolved.

`docs/backlog/jest-cases-the-browser-suite-could-take.md` records what a browser-side change would release, grouped by cause rather than by file, since one fix frees cases across several suites. Only `comments-counter` genuinely lacks a production selector; `spinner`, `preloader` and `comment-actions-additional` already carry static classes and simply have no browser case written yet.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,9 @@
+# The project gate fails on any drop by default. The frontend suite is shedding jest cases whose
+# behaviour the browser suite asserts instead, and codecov measures only the jest half, so every
+# such removal reads as a drop. Half a percent is enough to pass a deliberate removal and still
+# fail a change that stops covering something.
+coverage:
+  status:
+    project:
+      default:
+        threshold: 0.5%
--- a/docs/backlog/jest-cases-the-browser-suite-could-take.md
+++ b/docs/backlog/jest-cases-the-browser-suite-could-take.md
@@ -1,0 +1,81 @@
+---
+worth: later
+where: frontend/apps/remark42/app, e2e/
+added: 2026-08-30
+---
+# What keeps jest cases out of the browser suite, and what would let them go
+
+Removing a jest case is only safe when a named e2e case asserts the same behaviour, assertion for
+assertion. Working through the suites on that rule, the cases that stay divide into a few recurring
+causes, and most of them are a gap in the browser suite rather than something it cannot reach.
+
+This is the list to work from. It is deliberately not acted on in the same change as the deletions:
+each item is an e2e case to write, and the jest cases it releases can then go with it.
+
+## Selectors the browser cannot reach
+
+`tasks/babel-plugin-remove-test-id.js` removes `data-testid` from anything but a test build, so a
+browser case cannot select by it. Only one of the elements these jest cases assert through is
+actually stuck behind that:
+
+- `comments-counter` in `profile.spec.tsx`, asserted in four cases, is
+  `<div className={styles.container} data-testid="comments-counter">`. Its only class is hashed by
+  the CSS modules, so nothing outside the bundle can name it. Giving it a stable class, the way
+  `.auth-button`, `.auth-submit`, `.comment-actions` and `.sort-picker` are kept outside the
+  modules for this reason, is what makes the counter assertable in a browser
+
+Three others are already reachable and simply have no browser case written yet, which is a smaller
+job than it looked:
+
+- `spinner` renders `clsx('spinner', styles.root, …)` with `role="presentation"`
+- `preloader` renders `clsx('preloader', className)` with `aria-label="Loading..."`
+- `comment-actions-additional` renders `clsx('comment-actions-additional', …)`, so the order of the
+  admin actions can be asserted from a browser case today
+
+## Transient states nothing waits on
+
+- buttons disabled while a vote request is in flight (`comment-votes.spec.tsx`, three cases)
+- the loading indicator in the telegram subscription panel
+- the spinner between pages of the profile list
+- the preloader that must not reappear after a load-more click
+
+Each needs a browser case that holds the request open, which the suite already knows how to do:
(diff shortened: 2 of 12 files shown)


umputun:
the replacement holds everywhere except score visibility, so this is the one thing to settle before it goes in.

`comment-votes.spec.tsx` asserted `toBeVisible` on `Votes score` at 0, +1 and -1, and `comment.test.tsx` had its own visibility check. Nothing picks those up. The browser suite reads the score through `score()` in `e2e/vote_test.go:19`, which goes to `pollText` and so to `Locator.InnerText`, and `innerText` returns the descendant text content when the element is not rendered. The vote buttons are separate siblings, so clicking them proves nothing about the score. `display: none` on that div still satisfies these score assertions after this PR. What survives in jest is `toHaveTextContent` and one `not.toBeInTheDocument`, neither of which is a visibility check.

either keep those three cases, or add a visible-score assertion at zero, positive and negative in the browser cases. #2240 needs a rebase anyway now that master moved, so it can carry the fix if that is easier.

everything else checks out. I read each removed group against the browser case named for it: the vote buttons and the disabled upvote against `UpvoteCountsOnce` and `DownvoteAndCorrection`, verification and the admin toggle against `AdminPinsAndVerifies`, edit and delete against `EditWithinTheDeadline`, `EditExpiresAfterTheDeadline` and `DeleteRemovesTheText`, the restored draft against `DraftSurvivesReloadAndClearsAfterPost`, simple view against `SimpleViewHidesTheEditingFurniture`, anonymous profile removal against `AnonymousReaderIsNotOfferedRemoval`, the iframe pair against `StaysHiddenUntilTheDocumentReportsInited` and `IsRevealedByTheInitedMessage`, reply against `TestComment_ReplyNestsUnderItsParent`, pin and unpin against `TestComment_AdminPinsAndVerifies`, the enabled Telegram subscription against `TestTelegramSub_RoundTrip`, and the sort-picker class against the `.sort-picker select` helper `TestThread_SortChangeReordersComments` uses. The auth-closes-on-document-click case keeps its proof in the retained height-on-close jest case rather than in the browser, which is fine.

paskal:
Added the browser assertions instead of keeping the three cases. `TestVote_UpvoteCountsOnce` now asserts the counter visible at zero before the click and at one after it, and `TestVote_DownvoteAndCorrection` at minus one, beside the text each already read. Checked the way you described it: with `display: none` on the counter and the image rebuilt, both cases fail at the new lines, `vote_test.go:53` and `:213`, and the `-1` text assertion before the second of those still passes, so `innerText` was indeed reading the hidden element.

The branch is rebased onto master now that #2228 and #2234 are in, so the diff is the one commit plus this one, and #2240 is rebased on top of it. Nothing else in the removal changed.

paskal:
One more commit since your pass: a `codecov.yml` giving the project status half a percent of slack. This branch and #2240 remove jest cases whose behaviour the browser suite asserts, and codecov measures only the jest half, so each removal read as a drop of a few hits with the line count unchanged and painted the branch red while covering nothing less. The removal itself is unchanged; #2240 is restacked on top.

umputun (approved):

PR #2248, opened 2026-09-08
Correct stale configuration and contributor documentation

paskal:
The documentation lists 17 of the 26 shipped locales and contains stale widget defaults, configuration names and contributor commands.

Update the locale list, mobile comment limit, subscription options and widget behaviour. Correct the authentication requirements, formatting environment variable, email template options, frontend development and translation instructions, and outdated paths and archive extraction commands.

Validation: Hugo build, internal link and anchor checks across all 35 generated HTML files, and an exact comparison of the documented locale codes with all 26 supported catalogues. Configuration and command changes were checked against their source definitions.


paskal:
Corrected the restore/import grammar and separated authentication guidance from the two required startup parameters. The Hugo build, link checks and rendered-list check pass.

umputun (approved):

PR #2250, opened 2026-09-13
Stop capping cached responses at 64 KiB by default

### LINKED ISSUE #2249: Reduced database search speed
The new version has resulted in a significant (threefold) decrease in the speed of data extraction from the database (https://site.com/remark/api/v1/find)
On a page without comments, the speed is identical between the old and new versions, but if there are many messages on the page, the speed drops sharply.

**Version 1.16.4**
Image

**Version 1.17.0**
Image

paskal:
Previously, `cache.max.value` defaulted to 65536 bytes, and the limit had never been enforced for the byte slices remark42 caches. lcw 2.1.0 began measuring byte slices, which activated the cutoff in 1.17.0: a `/api/v1/find` response of 64 KiB or more is rebuilt on every request, while smaller pages stay cached.

After this change, the default is 0, unlimited per value, which is the behaviour every earlier release had in practice. `cache.max.size` still caps the whole cache at 50 MB, and an explicit `CACHE_MAX_VALUE` keeps its value.

On a local page with 300 comments, a 275 KB response, median server time was 4.6 ms on v1.16.4, 200 ms on v1.17.0, and 5.0 ms on v1.17.0 with `CACHE_MAX_VALUE=0`. The new test builds the cache from default options and asserts that a 300 KB value is loaded once across three reads; it fails with the previous default.

Resolves #2249.


paskal:
Addressed in the last commit: the test now clears the four `CACHE_*` variables for its parse and restores them on cleanup, so a value in the runner's environment cannot replace the built-in default.

umputun (approved):
lgtm, thx

